PR #7, opened 2019-04-17
Toggle: allow customized attributes

carlito-scality:


src/lib/components/toggle/Toggle.component.js
```diff
@@ -77,12 +77,12 @@ const StyledSwitchLabel = styled.span`
...
 `;
 
-function ToggleSwitch({ toggle, label, onChange }) {
+function ToggleSwitch({ toggle, label, onChange, ...rest }) {
   return (
-    <ToggleContainer>
-      <Switch className="switch">
-        <input type="checkbox" checked={toggle} onChange={onChange} />
-        <span className="slider" />
+    <ToggleContainer className="sc-toggle">
```
nicolas2bert: **idea**: why not using `cu` or `cui` prefix (i.e.: cui-toggle). We might have used/or going to use `sc` prefix in other css.

carlito-scality: yes but you can put whatever name and you can not be sure that it is unique in your app except it contains a hash as styled-components does. The reason I add it to allows consuming project to override it. So I would say it is the consuming project's accountability to make sure the name's uniqueness.

PR #8, opened 2019-04-18
Color: add to peerdependencies

carlito-scality:


package.json
```diff
@@ -52,6 +52,7 @@
   },
   "peerDependencies": {
     "@fortawesome/fontawesome-free": "5.7.2",
+    "color": "^3.1.0",
```
nicolas2bert: Need to be remove here: https://github.com/scality/core-ui/pull/8/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R38?

nicolas2bert: You might want to add "styled-components" as well inside your "peerDependencies"

PR #26, opened 2019-05-17
Component: Notifications

carlito-scality:
image

### DIFF
--- a/src/lib/components/notifications/Notification.component.js
+++ b/src/lib/components/notifications/Notification.component.js
@@ -1,9 +1,10 @@
 import React, { useState, useEffect } from "react";
-import styled from "styled-components";
+import styled, { css } from "styled-components";
 import Button from "../button/Button.component";
 import { Timer } from "../../utils";
 import * as defaultTheme from "../../style/theme";
 import { mergeTheme } from "../../utils";
+import type { Variant } from "../constants";
 
 type Item = { label: string, onClick: () => void };
 type Items = Array<Item>;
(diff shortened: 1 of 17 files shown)


src/lib/components/notifications/Notification.component.js
```diff
@@ -0,0 +1,106 @@
+import React, { useState, useEffect, useRef } from "react";
```
MonPote: Do you want to add flow ?

src/lib/components/notifications/Notification.component.js
```diff
@@ -0,0 +1,106 @@
...
+import styled, { css } from "styled-components";
+import Color from "color";
+import Button from "../button/Button.component";
```
MonPote: You can remove Button dependency I think.

src/lib/components/notifications/Notifications.component.js
```diff
@@ -0,0 +1,65 @@
+import React from "react";
+import styled, { css } from "styled-components";
+import Button from "../button/Button.component";
```
MonPote: Button and Timer are not used.

src/lib/components/notifications/Notifications.component.js
```diff
@@ -0,0 +1,65 @@
...
+import { Timer } from "../../utils";
+import * as defaultTheme from "../../style/theme";
+import { mergeTheme } from "../../utils";
```
MonPote: Same for mergeTheme

PR #28, opened 2019-05-23
Monpote/27 improve table

MonPote:
Add the possibility to change the width of columns.


carlito-scality:
You should update tests  when updating stories. Otherwise LGTM

PR #31, opened 2019-05-31
Loader: remove LoaderText if no children

carlito-scality:
-remove LoaderText if no children otherwise there will be always padding problem because of LoaderText


MonPote (approved):
LGTM

PR #72, opened 2019-09-06
SearchInput: add the possibility to disable the default toggle

ChengYanJin:
**Component**:
SearchInput

**Description**:
We would like to disable the default toggle so that we can see the whole search bar at the beginning.

**Design**:
image


carlito-scality (approved):
LGTM

ChengYanJin:
@bert-e approve

PR #73, opened 2019-09-09
add a condidtion to check if reset button should be displayed

ChengYanJin:
**Component**: selectInput

**Description**:
When we don't define our own method `onReset` in Metalk8s, we are not supposed to see the delete button. 

**Design**:
image

### DIFF
--- a/src/lib/components/searchinput/SearchInput.component.js
+++ b/src/lib/components/searchinput/SearchInput.component.js
@@ -10,7 +10,7 @@ type Props = {
   placeholder?: string,
   value: string,
   onChange: () => void,
-  onReset: () => void,
+  onReset?: () => void,
   disableToggle: boolean
 };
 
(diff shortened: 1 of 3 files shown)


src/lib/components/searchinput/SearchInput.component.js
```diff
@@ -104,7 +104,10 @@ function SearchInput({
       <SearchIcon onClick={toggle} disabled={!docked}>
         <i className="fas fa-search" />
       </SearchIcon>
-      <ResetIcon onClick={reset} visible={value && !docked}>
+      <ResetIcon
+        onClick={reset}
+        visible={value && !docked && typeof onReset === "function"}
```
carlito-scality: ```value && !docked && onReset``` is enough

carlito-scality: Please update the test

carlito-scality (approved):

ChengYanJin:
Wrong branch name... Create a new branch & PR

PR #77, opened 2019-09-10
integrate flow check in ci

ChengYanJin:
**Component**:
CI 

**Description**:
In order to have the error of flow pop up in CL, we would like to do `yarn run flow check` and also integrate into CI.

1. Integrated flow check in CI
2. Locally, when you do`npm run flow` it will also do the flow check.

**Breaking Changes**:
You will get the error of flow during build

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -35,5 +35,5 @@ stages:
           workdir: build/core-ui
           command: >
             npm install -g yarn &&
-            yarn && yarn test
+            yarn && yarn test && yarn run flow
           haltOnFailure: false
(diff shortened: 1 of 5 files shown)


eve/main.yml
```diff
@@ -35,5 +35,5 @@ stages:
           workdir: build/core-ui
           command: >
             npm install -g yarn &&
-            yarn && yarn test
+            yarn && yarn test && yarn run flow check
```
carlito-scality: use npm instead of yarn

carlito-scality (approved):

ChengYanJin:
@bert-e approve

PR #85, opened 2019-09-18
Remove CRA

MonPote:
Remove Create React App.


carlito-scality (approved):
LGTM! 
We need also manage the Eslint and get rid of yarn in the CI, but it can be done in another PR.
Good job!

MonPote:
@bert-e approve

PR #91, opened 2019-09-27
Feature/component tooltip

### LINKED ISSUE #83: Component: Tooltips
**Component**: Tooltips

**Why this is needed**:

**What should be done**:

**Test plan**:

ChengYanJin:
**Component**: Tooltip

**Description**: implement tooltip component

**Design**:
The props in Tooltip:
we are able to change the overlay style by `overlaystyle`.
Since the current case to trigger the tooltip is hover, we remove the `trigger`.
The default `placement` to display the tooltip is on the top.

We remove the package of `rc-tooltip`.
**Breaking Changes**:

- [] Breaking Changes

Closes: #83

### DIFF
--- a/src/lib/components/tooltip/Tooltip.component.js
+++ b/src/lib/components/tooltip/Tooltip.component.js
@@ -1,51 +1,95 @@
-// @flow
-
-import React from "react";
-import styled from "styled-components";
-import RcTooltip from "rc-tooltip";
+import React, { useState } from "react";
+import ReactDOM from "react-dom";
+import styled, { css } from "styled-components";
 import type { Node } from "react";
+import * as defaultTheme from "../../style/theme";
+import { mergeTheme } from "../../utils";
 
+export const TOP = "top";
+export const BUTTOM = "buttom";
 export const LEFT = "left";
 export const RIGHT = "right";
-export const BOTTOM = "bottom";
-export const TOP = "top";
-type Position = typeof LEFT | typeof RIGHT | typeof BOTTOM | typeof TOP;
-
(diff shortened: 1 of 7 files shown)


src/lib/components/tooltip/Tooltip.component.js
```diff
@@ -1,51 +1,102 @@
...
 
 function Tooltip({
-  placement = "right",
-  mouseEnterDelay = 0,
-  mouseLeaveDelay = 0,
-  destroyTooltipOnHide = true,
-  trigger = ["hover"],
-  overlay = null,
-  children
+  placement = "top",
```
carlito-scality: use contant TOP instead

src/lib/components/tooltip/Tooltip.component.js
```diff
@@ -1,51 +1,102 @@
...
+    (props && props.overlaystyle && props.overlaystyle.fontSize) ||
+    defaultTheme.fontSize.small};
+  text-align: center;
+  vertical-align: middle;
+  padding: ${defaultTheme.padding.smaller};
+    ${props => {
+      switch (props.placement) {
+        case LEFT:
+          return css`
+            right: 105%;
```
carlito-scality: I would prefer ```left: calc((-100% - 10px))``` instead of ```right: 105%``` because we will have always the same padding.

carlito-scality (changes requested):

carlito-scality (approved):
LGTM Good job :)

ChengYanJin:
@bert-e approve

PR #93, opened 2019-10-02
Workflows: Add "--" so that  npm will pass all the arguments

carlito-scality:
**Component**:Build

**Description**:
- The github action to deploy storybook does not executed successfully after replacing yarn by npm since we need to Add "--" so that  npm will pass all the arguments after the "--" directly to npm run script. Otherwise, the token is not passed to the script.

**Design**:

**Breaking Changes**:

- [] Breaking Changes

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/.github/workflows/storybookpublish.yml
+++ b/.github/workflows/storybookpublish.yml
@@ -13,6 +13,6 @@ jobs:
         with:
           node-version: 10.x
       - run: npm ci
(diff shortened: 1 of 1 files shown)


ChengYanJin (approved):

carlito-scality:
@bert-e approve

PR #95, opened 2019-10-04
Create Chips component

### LINKED ISSUE #82: Component: Chips
**Component**:Chips

**Why this is needed**:
- We need this component to display in the filters bar ou Select component (https://github.com/JedWatson/react-select/issues/1263)
image

**What should be done**:

**Test plan**:

StheffanyHadlich:
**Component**: Chips

**Description**: implement chips component

**Design**:
image

This component accepts an icon, text, an button icon and styles variations according to pre defined themes.

**Breaking Changes**:

- [] Breaking Changes

---
Closes: #82

### DIFF
--- a/src/lib/components/chips/Chips.component.js
+++ b/src/lib/components/chips/Chips.component.js
@@ -1,6 +1,7 @@
 //@flow
 import * as defaultTheme from "../../style/theme";
-
+import type { Variant } from "../constants";
+import type { Node } from "react";
 import styled, { css } from "styled-components";
 
 import React from "react";
@@ -8,34 +9,26 @@ import { lighten } from "polished";
 import { mergeTheme } from "../../utils";
 
 type Props = {
-  text: String,
-  color?: String,
+  text: string,
+  color?: string,
   variant?: Variant,
   icon?: Node,
   buttonIcon?: Node,
   onClick?: any => void,
-  onRemove?: any => void,
-};
-
-const defaultProps = {
-  icon: null,
-  buttonIcon:null,
-  text: "",
+  onRemove?: any => void
 };
 
 const ChipsContainer = styled.div`
   display: inline-flex;
-  background-color:${defaultTheme.grayLighter};
+  background-color: ${defaultTheme.grayLighter};
   border-radius: 15px;
-  cursor: pointer;
 
   ${props => {
     const brandingTheme = mergeTheme(props.theme, defaultTheme);
-    
+
     return css`
       background-color: ${brandingTheme[props.variant]};
       color: ${props.color};
-    
     `;
   }}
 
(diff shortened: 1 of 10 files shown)


src/lib/components/chips/Chips.component.js
```diff
@@ -0,0 +1,114 @@
...
+  display: inline-flex;
+  background-color:${defaultTheme.grayLighter};
+  border-radius: 15px;
+  cursor: pointer;
+
+  ${props => {
+    const brandingTheme = mergeTheme(props.theme, defaultTheme);
+
+    const brandLight = lighten(0.1, brandingTheme[props.variant]).toString();
+    console.log(brandingTheme);
```
carlito-scality: It should be removed.

src/lib/components/chips/Chips.component.js
```diff
@@ -0,0 +1,114 @@
...
+
+function Chips({
+  text,
+  variant,
+  icon,
+  buttonIcon,
+  onClick, 
+  color = defaultTheme.white,
+}: Props) {
+  return ( //tirar negação dos ícone
```
carlito-scality: Comments should be in english 😉

src/lib/components/chips/Chips.component.js
```diff
@@ -0,0 +1,114 @@
...
+import { lighten } from "polished";
+import { mergeTheme } from "../../utils";
+
+type Props = {
+  text: string,
+  color?: String,
+  variant?: Variant,
+  icon?: Node,
+  buttonIcon?: Node,
+  onClick?: any => void,
```
carlito-scality, line 17: It would be nicer to have a remove button as part of this component as you do in the storybook. 
We can add ‘onRemove?: any => void’ in the Props. Then if ‘onRemove’ is defined we display the remove Button, otherwise not.

src/lib/components/chips/Chips.component.js
```diff
@@ -0,0 +1,114 @@
...
+  display: inline-flex;
+  justify-content: center;
+  align-items: center;
+  padding:7px 14px;
+
+`;
+
+function Chips({
+  text,
+  variant,
```
carlito-scality: You can set the default value here instead of putting it in the defaulProps like for ´color’

src/lib/components/chips/Chips.component.js
```diff
@@ -0,0 +1,114 @@
+//@flow
+import * as defaultTheme from "../../style/theme";
+import styled, { css } from "styled-components";
+import React from "react";
+import { lighten } from "polished";
+import { mergeTheme } from "../../utils";
+
+type Props = {
+  text: string,
+  color?: String,
```
carlito-scality: string instead of String

carlito-scality (changes requested):
Hello! I have some small request changes but very good job. Thanks 😊

src/lib/components/chips/Chips.component.js
```diff
@@ -0,0 +1,114 @@
...
+    const brandingTheme = mergeTheme(props.theme, defaultTheme);
+
+    const brandLight = lighten(0.1, brandingTheme[props.variant]).toString();
+    console.log(brandingTheme);
+    
+    return css`
+      background-color: ${brandingTheme[props.variant]};
+      color: ${props.color};
+    
+      &:hover {
```
carlito-scality: hover and active should be needed only when onClick is defined. 
« pointer: cursor » should be added to the hover as well

src/lib/components/chips/Chips.component.js
MonPote: Yes, and `defaultProps` for function will eventually [be deprecated](https://github.com/reactjs/rfcs/pull/107)

StheffanyHadlich:
> Hello! I have some small request changes but very good job. Thanks

haha thanks, I'll fix it

src/lib/components/chips/Chips.component.js
StheffanyHadlich: oh sorry, my bad

src/lib/components/chips/Chips.component.js
StheffanyHadlich: Can I add those when onRemove is defined as well?

carlito-scality: I think the Remove button should have its own hover style so that in the case when we have onRemove and onClick both defined, we know what is clicked when clicking.

carlito-scality:
It would be nicer to have this component with different sizes but we could implement this feature later in another PR :)

carlito-scality:
@StheffanyHadlich I just pushed some small changes. If you agree, we can merge this PR :)

carlito-scality (approved):

StheffanyHadlich:
> @StheffanyHadlich I just pushed some small changes. If you agree, we can merge this PR :)

That is fine with me :)

PR #96, opened 2019-10-04
Feature/multiselect component

### LINKED ISSUE #12: Multi-Select List
https://github.com/scality/pensieve-front/blob/master/src/react/ui-elements/MultipleSelect.jsx

State 1
Screen Shot 2019-04-18 at 2 28 04 PM

State 2
Screen Shot 2019-04-18 at 2 27 33 PM

State 3
Screen Shot 2019-04-18 at 2 27 21 PM

------------------------------
New Design:

image

image

edumoreira1506:
**Component**:

**Description**:
Hello guys!

I tried to follow the pattern of project, but if have some change to do, I will fix it. So, the proposal was to made a new component, the multiselect component, and each option in multiselect component should had one field like "used" or no with one checkbox, each option too should have one "label" and one "value" like the example of the issue.

Detail: I created two buttons (add button and remove button) with pure css because of performance, both are very simple and I thought than don't needed use third libs.

**Design**:
Screenshot from 2019-10-04 15-38-08

---

Closes: #12 

-->

### DIFF
--- a/src/lib/components/multiselect/MultiSelect.component.js
+++ b/src/lib/components/multiselect/MultiSelect.component.js
@@ -1,4 +1,4 @@
-import React, { Component } from "react";
+import React, { useReducer } from "react";
 import styled, { css } from "styled-components";
 import * as defaultTheme from "../../style/theme";
 type MultiSelectProps = {
@@ -15,7 +15,7 @@ type MultiSelectProps = {
   },
   title: string,
   labelCheckBox: string,
-  width: string,
+  width: string
 };
 
 type MultiSelectOptionProps = {
@@ -48,6 +48,7 @@ const defaultProps = {
       placeholder: 'Label of option'
     },
   },
+  currentInput: 'title'
 };
 
 const MultiSelectContainer = styled.div`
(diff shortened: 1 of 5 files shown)


src/lib/components/multiselect/MultiSelect.component.js
```diff
@@ -0,0 +1,420 @@
+import React, { useReducer } from "react";
```
carlito-scality: You should enable Flow by adding ```//@flow```

carlito-scality:
@edumoreira1506 Thanks. I'm reviewing it :)

carlito-scality (changes requested):
> **Component**:
> 
> **Description**:
> Hello guys!
> 
> I tried to follow the pattern of project, but if have some change to do, I will fix it. So, the proposal was to made a new component, the multiselect component, and each option in multiselect component should had one field like "used" or no with one checkbox, each option too should have one "label" and one "value" like the example of the issue.
> 
> Detail: I created two buttons (add button and remove button) with pure css because of performance, both are very simple and I thought than don't needed use third libs.
> 
> **Design**:
> Screenshot from 2019-10-04 15-38-08
> 
> Closes: #12
> 
> -->

Hello,
It looks good your component. However, this library contains only stateless components (See https://github.com/scality/core-ui#what-are-stateless-components).I just updated the README :) It means that those components do NOT manage state (except only for css rendering). Indeed, your component is a smart component ;) because it contains reducer, state and dispatch which should be implemented in the consuming project.

Based on that, the ```props``` of your component can look something lick below:
```
export type Props = {
  title: string,
  items: Array<Item>,
  options: Options,
  checkboxLabel: string,
};

export type Item = {
  label: string,
  description?: string,
  onRemove: any => void,  //on remove button click
  onChange: any => void  //on checkbox click
};

export type Options = {
  icon?: Node,
  placeholder?: string,
  optionsList: Array<Option>  // "Content of option"
  onSelect: any => void  //on option click
};

export type Option = {
  label: string,
  value: string | number
};

//pseudo code
function MultiSelect(props: Props) {
  return (
    <MultiSelectContainer>
        <MultiSelectHeader>{props.title}</MultiSelectHeader>
        <SelectedItemsContainer>{props.items.map(...)}</SelectedItemsContainer>
        <SelectContainer><Select options={props.options.optionsList}.../></SelectContainer>
    </MultiSelectContainer>
  );
}
```
- By the way, we need to reuse existing components like Checkbox (https://scality.github.io/core-ui/?path=/story/checkbox--default) for "Preferred" checkbox and Select (https://scality.github.io/core-ui/?path=/story/select--default) for "MultiSelectForm". Actually, the existing Select component does not support the idea to have the icon ("+" icon) inside the input. I will create a PR to add this feature.

Hopefully, it is clear for you. Feel free to ask questions. Thanks :)

src/lib/components/multiselect/MultiSelect.component.js
```diff
@@ -0,0 +1,420 @@
...
+          placeholder={state.newOption[state.currentInput].placeholder}
+          onChange={e => dispatch({ type: 'change', newValue: e.target.value })}
+          onClick={() => dispatch({ type: 'add' })}
+        />
+      </BlockContainer>
+    </MultiSelectContainer>
+  );
+}
+
+MultiSelect.defaultProps = defaultProps;
```
carlito-scality: You should remove defaultProps since it will be deprecated soon. You can use `Default Function arguments` instead

edumoreira1506:
Thanks for the feedback! I believe then I fixed everything of you said. :thinking:

carlito-scality:
> Thanks for the feedback! I believe then I fixed everything of you said. 🤔

Hello!!! Thanks. I'm reviewing it and I will probably come up with some improvement commits :)

PR #104, opened 2019-10-22
Added different sizes to chip component

### LINKED ISSUE #99: Chips should support different sizes
**Component**: Chips (https://scality.github.io/core-ui/?path=/story/chips--default)

**Why this is needed**:
- Chips current has only one size, it should be great to have it in different sizes like Button, Dropdown...

**What should be done**:

**Test plan**:

edumoreira1506:
**Component**: Chips

**Description**: Before the chips component had just one size. Now have 4 options.

**Design**: 
Screenshot from 2019-10-22 15-00-11

Closes: #99

### DIFF
--- a/src/lib/components/chips/Chips.component.js
+++ b/src/lib/components/chips/Chips.component.js
@@ -1,6 +1,6 @@
 //@flow
 import * as defaultTheme from "../../style/theme";
-import type { Variant } from "../constants";
+import type { Variant, LoaderSize } from "../constants";
 import type { Node } from "react";
 import styled, { css } from "styled-components";
 
@@ -15,7 +15,8 @@ type Props = {
   icon?: Node,
   buttonIcon?: Node,
   onClick?: any => void,
-  onRemove?: any => void
+  onRemove?: any => void,
+  size?: LoaderSize
 };
 
 const ChipsContainer = styled.div`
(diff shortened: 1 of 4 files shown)


src/lib/components/chips/Chips.component.js
```diff
@@ -1,6 +1,6 @@
 //@flow
 import * as defaultTheme from "../../style/theme";
-import type { Variant } from "../constants";
+import type { Variant, LoaderSize } from "../constants";
```
carlito-scality: It is a good a idea to use `LoaderSize` but the we should rename it to `Size` in `constant.js`

src/lib/components/chips/Chips.component.js
```diff
@@ -25,11 +26,45 @@ const ChipsContainer = styled.div`
...
+    const defaultStyle = `
       background-color: ${brandingTheme[props.variant]};
       color: ${props.color};
-    `;
+    `
+
+    switch (props.size) {
+      case "smaller":
+        return css`
+          font-size: ${defaultTheme.fontSize.smaller};
```
carlito-scality: We should have to the same mapping between `props.size` and `defaultTheme.fontSize`to be consistent.

=> `font-size: ${defaultTheme.fontSize[props.size]};`

src/lib/components/chips/Chips.component.js
```diff
@@ -25,11 +26,45 @@ const ChipsContainer = styled.div`
```
carlito-scality, line 26: ` border-radius` should be based on the `props.size` to be neat.

carlito-scality: idem for `border-radius: 15px;` and `padding: 7px 8px;` of ChipsIcon

src/lib/components/chips/Chips.component.js
```diff
@@ -25,11 +26,45 @@ const ChipsContainer = styled.div`
...
+          font-size: ${defaultTheme.fontSize.larger};
+          ${defaultStyle}
+          height: 35px;
+          width: 200px;
+        `;
+      case "larger":
+        return css`
+          font-size: ${defaultTheme.fontSize.huge};
+          ${defaultStyle}
+          height: 40px;
```
carlito-scality, line 59: `height`and `width` should be removed since they are necessary here

carlito-scality (changes requested):

src/lib/components/chips/Chips.component.js
edumoreira1506: I added `Size` in `constant.js` I could not rename, because is used in other files too.

carlito-scality: Yes you could :) since we should have only one type for the  Size. Like that,  you rename it where it is used.

edumoreira1506:
I believe that now, everything is ok! Look:
Screenshot from 2019-10-24 11-54-17

carlito-scality:
Merged in https://github.com/scality/core-ui/pull/111

PR #110, opened 2019-10-28
Theme: support dark/light modes

### LINKED ISSUE #105: Dark Mode and Light Mode support
**Component**: Theme

**Why this is needed**:
- It would be nice to have all components supporting Dark Mode and Light Mode. We currently support Light Mode.
- Do not forget to theme Storybook to test components (https://storybook.js.org/docs/configurations/theming/)

**What should be done**:

**Test plan**:

carlito-scality:
**Component**:Theme

**Description**:
- Update all component color (background-color, color, border-color...) based on new color palette
- In the theme.js, we define a default light mode's colors palette, but It is up to user to define their own palette.
- This color palette is to be improved later by our UX team.

**Design**:

**Breaking Changes**:

- [] Breaking Changes

---

Closes: https://github.com/scality/core-ui/issues/105

### DIFF
--- a/src/lib/components/searchinput/SearchInput.component.js
+++ b/src/lib/components/searchinput/SearchInput.component.js
@@ -22,8 +22,7 @@ const SearchInputContainer = styled.div`
     input[type="text"] {
       width: ${props => (props.docked ? "40px" : "100%")};
       box-sizing: border-box;
-      padding-right: ${props => (props.docked ? "12px" : "40px")};
-      padding-left: ${props => (props.docked ? "12px" : "40px")};
+      padding-left: ${props => (props.docked ? "28px" : "40px")};
       transition: width 0.2s ease-in-out;
     }
   }
@@ -38,7 +37,7 @@ const IconButton = styled.button`
   border-radius: 5px;
   font-size: ${defaultTheme.fontSize.base};
   color: ${props => mergeTheme(props.theme, defaultTheme).primary};
-  background-color: ${defaultTheme.white};
+  background-color: transparent;
   ${props => {
     return (
       !props.disabled &&
--- a/stories/chips.js
+++ b/stories/chips.js
@@ -3,17 +3,18 @@ import Chips from "../src/lib/components/chips/Chips.component";
 import React from "react";
 import { action } from "@storybook/addon-actions";
 import { storiesOf } from "@storybook/react";
+import { Wrapper, Title } from "./common";
 
 storiesOf("Chips", module).add("Default", () => {
   return (
-    <div className="storybook-chips">
-      <h3>Basic Chip</h3>
+    <Wrapper className="storybook-chips">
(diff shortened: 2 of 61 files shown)


stories/tooltip.js
```diff
@@ -57,15 +58,15 @@ storiesOf("Tooltip", module).add("Default", () => {
...
-          <span>tooltip with icon</span>
+          <SubTitle>tooltip with icon</SubTitle>
         </Tooltip>
       </div>
       <div>
-        <h3>Tooltip doesn't trigger</h3>
+        <Title>Tooltip doesn't trigger</Title>
         <Tooltip placement="bottom">
-          <span>Hover here!</span>
+          <SubTitle>Hover here!</SubTitle>
```
ChengYanJin, line 67: I think we should create another wrapper `<Text>` instead of `<SubTitle>`. 
For the single text, we don't need to use `<h3>` or `<h4>`, so that the tooltip can have the same padding in four different positions (left, right, top, bottom)

stories/loader.js
```diff
@@ -17,5 +18,9 @@ storiesOf("Loader", module)
     const size = text("Size", "base");
     const content = text("Children", "Loading");
 
-    return <Loader size={size}>{content}</Loader>;
+    return (
+      <Wrapper>
+        <Loader size={size}>{content}</Loader>
+      </Wrapper>
```
ChengYanJin, line 24: We need to add a wrapper for text as well. As we can see from the storybook, the color of {content} is still black in the Dark Theme.

src/lib/components/searchinput/SearchInput.component.js
```diff
@@ -22,8 +22,7 @@ const SearchInputContainer = styled.div`
     input[type="text"] {
       width: ${props => (props.docked ? "40px" : "100%")};
       box-sizing: border-box;
-      padding-right: ${props => (props.docked ? "12px" : "40px")};
-      padding-left: ${props => (props.docked ? "12px" : "40px")};
+      padding-left: ${props => (props.docked ? "28px" : "40px")};
       transition: width 0.2s ease-in-out;
     }
```
ChengYanJin, line 27: we need to set the background color of `input:-internal-autofill-selected `
Otherwise, when we choose something from the autofill list, the input background-color will be changed to white.

stories/chips.js
```diff
@@ -9,7 +9,6 @@ storiesOf("Chips", module).add("Default", () => {
     <div className="storybook-chips">
```
ChengYanJin: We need to use <Wrapper> as well in order to change the theme :)

ChengYanJin (changes requested):
Some small improvement, please.

ChengYanJin (approved):
LGTM!

carlito-scality:
@bert-e approve

PR #119, opened 2019-11-04
CI: Add post-merge stage to deploy Storybook

carlito-scality:
**Component**:CI

**Description**:
- Add post-merge stage to deploy Storybook in the CI instead of using Github Actions

**Design**:

**Breaking Changes**:

- [] Breaking Changes

---

Closes: #ISSUE_NUMBER


eve/main.yml
```diff
@@ -39,3 +41,34 @@ stages:
             npm run flow &&
             npm run lint
           haltOnFailure: false
+  post-merge:
```
tcarmet, line 33: For now, there's no need to split your post-merge stage into 2, I would directly put the `storybook_deploy` stage like the following:
```
 development/*:
    stage: storybook_deploy
```

or rename the `storybook_deploy` as `post-merge`

eve/main.yml
```diff
@@ -39,3 +41,34 @@ stages:
...
+    worker:
+      type: local
+    steps:
+      - TriggerStages:
+          name: Trigger storybook_deploy
+          stage_names:
+            - storybook_deploy
+          haltOnFailure: true
+  storybook_deploy:
```
tcarmet: nit: I find it weird to use `-` and `_` as a seperator to name things, either use `-` everywhere or `_`

but this is just a personal thought, feel free to do whatever.

eve/main.yml
```diff
@@ -39,3 +41,34 @@ stages:
...
+    worker:
+      type: kube_pod
+      path: eve/workers/pod-unit-tests/pod.yaml
+      images:
+        docker-unit-tests:
+          context: '.'
+          dockerfile: eve/workers/pod-unit-tests/Dockerfile
+    steps:
+      - Git: &git_pull
```
tcarmet: you can replace this whole step as
`- Git: *git_pull` and that's it, remove all the other lines.

here you're creating two reference with the `&` character, which makes your `yaml` wrong

tcarmet (changes requested):

eve/main.yml
```diff
@@ -39,3 +41,34 @@ stages:
...
+          method: clobber
+          retryFetch: true
+          haltOnFailure: true
+      - ShellCommand:
+          name: Deploy storybook
+          workdir: build/core-ui
+          command: >
+            npm ci &&
+            npm run storybook:deploy -- --ci --host-token-env-variable=%(secret:core_ui_github_token)s
+          haltOnFailure: false
```
tcarmet, line 50: could you add here the following parameter
```
env:
  GIT_CONFIG_NOSYSTEM: "1"
```

eve/main.yml
```diff
@@ -39,3 +32,22 @@ stages:
...
             npm run lint
           haltOnFailure: false
+  post-merge:
+    worker:
+      type: kube_pod
+      path: eve/workers/pod-unit-tests/pod.yaml
+      images:
+        docker-unit-tests:
+          context: '.'
+          dockerfile: eve/workers/pod-unit-tests/Dockerfile
```
tcarmet: Could you explain to me why the context needs to be `.` because I took a look at the dockerfile, and it doesn't seem to be needed. You could even replace those lines to :
```yaml
docker-unit-tests: 'eve/workers/pod-unit-tests'
```

and same story for the one on the `pre-merge` as it's the same worker.

eve/main.yml
```diff
@@ -39,3 +30,20 @@ stages:
             npm run flow &&
             npm run lint
           haltOnFailure: false
+  post-merge:
+    worker:
```
tcarmet: could you add, at the same lever as the worker the option `simultaneous_builds`. Like the following:

```yaml
post-merge:
  simultaneous_builds: 1
  worker:
...
```
This will ensure that only one post-merge will be executed at a time. Per example if 5 merge happen, one after the other, they will not be running over each other, with this you ensure that the content of your page will be coherent.

eve/main.yml
```diff
@@ -4,25 +4,16 @@ version: "0.2"
...
-    worker:
-      type: local
-    steps:
-      - TriggerStages:
-          name: Trigger unit tests
-          stage_names:
-            - unit_tests
-          haltOnFailure: true
-  unit_tests:
     worker:
```
tcarmet: could you add a reference, here to the worker.
Like the following `worker: &worker_unit_tests` (you can name it however you want).

once you added it, modify the content of the worker on the `post-merge` stage with the following:
```yaml
post-merge:
  worker: *worker_unit_tests
```

eve/main.yml
```diff
@@ -39,3 +30,22 @@ stages:
...
+    steps:
+      - Git: *git_pull
+      - ShellCommand:
+          name: Deploy storybook
+          workdir: build/core-ui
+          command: >
+            npm ci &&
+            npm run storybook:deploy -- --ci --host-token-env-variable=GITHUB_TOKEN
+          env:
+            GIT_CONFIG_NOSYSTEM: "1"
```
tcarmet, line 48: Could you just add the following comment above the `GIT_CONFIG_NOSYSTEM`

```
# Our system gitconfig is setup to go through a gitcache service and
# as we're preforming write operations with git on this step, we don't want that the request
# go through the gitcache.
```

tcarmet: it's a reminder in case someone wants to remove this option in the future :P

tcarmet (approved):
one last request to add a comment. Otherwise it's looking good! I'm gonna run a little test and come back with results

PR #120, opened 2019-11-05
Button: improve inverted disabled icon button style

carlito-scality:
**Component**:Button

**Description**:
- improve inverted disabled icon button style because currently we have a background color that should not be there.

**Design**:
image

**Breaking Changes**:

- [] Breaking Changes

---

Closes: #ISSUE_NUMBER


ChengYanJin (approved):
LGTM!

PR #122, opened 2019-11-07
feat(navbar-tabs): Add tabs to the navbar

felipebrunetti94:
Gravação de tela de 07-11-2019 16_28_38

**Component**:

**Description**:
- Alter Navbar.component.js to include tabs
- Alter navbar.js to include tabs

Closes: #116

PR #126, opened 2019-11-08
Merge navbar tabs

### LINKED ISSUE #116: Navbar: Improvements
**Component**: Navbar

**Why this is needed**:

**What should be done**:

Actual Navbar:
image

Use case to be covered:
image

Expected Navbar:
navbar

```
- Add `tabs` in the center of the Navbar.  Navbar inputs would be:

type Tab = {
  title: string,
  selected?: boolean,
  onClick: any => void
};

export type Props = {
  onToggleClick?: () => void,
  toggleVisible?: boolean,
  productName?: string,
  applications?: Items,
  help?: Items,
  user?: User,
  logo?: Node,
  languages?: Items,
  tabs?:Tab,
};

```
**Test plan**:

carlito-scality:
**Component**:Navbar 

**Description**:
- Merge open source PR: https://github.com/scality/core-ui/pull/122

**Design**:

**Breaking Changes**:

- [] Breaking Changes

---

Closes: #116

### DIFF
--- a/src/lib/components/navbar/Navbar.component.js
+++ b/src/lib/components/navbar/Navbar.component.js
@@ -8,7 +8,7 @@ import Button from "../button/Button.component";
 import * as defaultTheme from "../../style/theme";
 import { mergeTheme } from "../../utils";
 import type { Item } from "../dropdown/Dropdown.component";
-import {padding} from "../../style/theme"; 
+
 
 type Items = Array<Item>;
 type User = {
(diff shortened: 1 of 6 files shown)


src/lib/components/navbar/Navbar.component.js
```diff
@@ -48,6 +54,43 @@ const NavbarMenu = styled.div`
...
+    props.selected &&
+    css`
+      border-bottom: 2px solid ${mergeTheme(props.theme, defaultTheme).primary};
+      span {
+        padding-top: 2px;
+      }
+    `};
+`;
+
 const NavbarMenuItem = styled.div`
```
ChengYanJin, line 94: I think we should give`.sc-dropdown {
    cursor: pointer;
  }` and add ```cursor: pointer;``` for button in NavbarMenuItem so that the user can know it's clickable.

ChengYanJin (changes requested):

src/lib/components/navbar/Navbar.component.js
carlito-scality: good idea but we should update the Dropdown component instead of overriding the style here. Let's create another ticket please

ChengYanJin (approved):
I added the `cursor:pointer` for the button in the NarbarMenu wondering if it's good for you?

ChengYanJin:
> I added the `cursor:pointer` for the button in the NarbarMenu wondering if it's good for you?

ok. Since it's the issue of button. I will do it in another PR.

PR #127, opened 2019-11-11
feat(textarea): Add text area component

### LINKED ISSUE #113: Input should support TextArea type
**Component**:Input

**Why this is needed**:
- this component currently supports `<input>` whose types are `text` and  `checkbox`. It should support `TextArea` type as well like below:

image

**What should be done**:

**Test plan**:

felipebrunetti94:
- Add new component with plop <TextArea/>

Fix #113

**Component**:

`<TextArea />` component

**Description**:
New input, text area, with disable and error options
**Design**
Add  `<TextArea />` component, with the plop commad, design follows the patterns from other components

---

Closes: #I113


src/lib/components/textarea/TextArea.component.js
```diff
@@ -0,0 +1,95 @@
...
+};
+
+const TextAreaContainer = styled.div`
+  display: flex;
+  align-items: flex-start;
+  justify-content: center;
+  flex-direction: column;
+
+  ${props => {      
+    console.log("defaultTheme: ", defaultTheme)
```
negreirosleo: :eyes:

src/lib/components/textarea/TextArea.component.js
```diff
@@ -0,0 +1,95 @@
...
+
+const LabelStyle = styled.label`
+  font-size: ${defaultTheme.fontSize.large};
+  padding-left: ${defaultTheme.padding.base};
+  vertical-align: middle;
+  color: ${props => mergeTheme(props.theme, defaultTheme).text};
+`;
+
+const TextareaErrorMessage = styled.span`
+  color: ${props => mergeTheme(props.theme, defaultTheme).danger};
```
negreirosleo: since you are using this: `mergeTheme(props.theme, defaultTheme)` a couple times why dont you assign its value to a a variable instead of multiple function calls?
```js
const brandingTheme = mergeTheme(props.theme, defaultTheme); // and use this in all other functions

const { text, danger, primary } = mergeTheme(props.theme, defaultTheme) //or already destructuring
```

and then just call the variable where you need it :ok_woman:

stories/index.js
```diff
@@ -19,3 +19,4 @@ import "./modal";
 import "./tooltip";
 import "./multiselect";
 import "./chips";
+import "./textarea";
```
laka3000: No new line at EOF

src/lib/components/textarea/TextArea.component.js
felipebrunetti94: ops...

src/lib/components/input/Input.component.js
```diff
@@ -4,6 +4,7 @@ import styled, { css } from "styled-components";
 import { DebounceInput } from "react-debounce-input";
 import Checkbox from "../checkbox/Checkbox.component";
 import Select from "../select/Select.component";
+import TextArea from "../textarea/TextArea.component";
```
carlito-scality, line 7: Where do you use it in this component?
you may add in the `InputRenderer ` like above so that `Input` can support the type `textarea`:
```
if (type === "select") {
...
} else if (type === "checkbox") {
...
} else if (type === "textarea") {
    < TextArea .../>
} else {
...
}
```

src/lib/components/textarea/TextArea.component.js
```diff
@@ -0,0 +1,94 @@
+//@flow
+import React from "react";
+import styled, { css } from "styled-components";
+import * as defaultTheme from "../../style/theme";
+import { mergeTheme } from "../../utils";
+
+type Props = {
+  rows?: int,
```
carlito-scality: `number` instead of `int`

src/lib/components/textarea/TextArea.component.js
```diff
@@ -0,0 +1,94 @@
...
+  cols?: int,
```
carlito-scality: number instead of int

src/lib/components/textarea/TextArea.component.js
```diff
@@ -0,0 +1,94 @@
...
+  placeholder?: string,
+  label?: string,
+  value?: string,  
+  error?: string,
+  disabled?: boolean,
+  onChange: () => void
```
carlito-scality: `any => void` instead

src/lib/components/textarea/TextArea.component.js
```diff
@@ -0,0 +1,94 @@
...
+  color: ${props => mergeTheme(props.theme, defaultTheme).text};
+`;
+
+const TextareaErrorMessage = styled.span`
+  color: ${props => mergeTheme(props.theme, defaultTheme).danger};
+  margin: ${defaultTheme.padding.smaller} 0;
+  font-size: ${defaultTheme.fontSize.small};
+`;
+
+function TextArea({ id,rows,cols, label, error, placeholder,disabled,...rest }: Props) {
```
carlito-scality, line 55: This component should only deal with `textarea`. Then you use it in `Input.component` as mentioned in above comment. Consequently, you do not need manage `label ` and `error ` here because they are already added in `Input.component`

carlito-scality: - `id` is not required so can be included in `...rest`
- we need to put some default values for `rows = 3` and `cols = 20` for example

carlito-scality (changes requested):

carlito-scality:
Merged in https://github.com/scality/core-ui/pull/133

PR #129, opened 2019-11-12
Dropdown: Add cursor pointer to DropdownStyled

ChengYanJin:
**Component**:
Dropdown, Button

**Description**:
When hover to the dropdown/button it's better to show the pointer so that the user can understand it's clickable.

**Design**:

### DIFF
--- a/src/lib/components/button/Button.component.js
+++ b/src/lib/components/button/Button.component.js
@@ -42,6 +42,7 @@ export const ButtonStyled = styled.button`
   &:focus,
   &:active {
     outline: none;
+    cursor:pointer;
   }
 
   ${props => {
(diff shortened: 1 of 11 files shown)


src/lib/components/button/Button.component.js
```diff
@@ -97,7 +98,8 @@ export const ButtonStyled = styled.button`
         border-color: ${brandingTheme[props.variant]};
         background-color: ${brandingTheme.background};
         color: ${brandingTheme[props.variant]};
-        
+        cursor:pointer;
```
carlito-scality: IMO, we do not need it here.

src/lib/components/button/Button.component.js
```diff
@@ -179,7 +181,8 @@ ${props => {
         border: none;
         background-color: transparent;
         color: ${props.disabled ? brandLight : brandingTheme[props.variant]};
-        
+        cursor:pointer;
```
carlito-scality: IMO, we do not need it here

carlito-scality (changes requested):

src/lib/components/button/Button.component.js
ChengYanJin: You are right, thanks!

carlito-scality (approved):

PR #136, opened 2019-11-16
Improvement/vegachart supports dark mode

ChengYanJin:
**Component**: LineChart, VegaChart

**Description**: we need to support VegaChart component to switch in between dark/light mode.

**Design**:
Light mode: 
image

Dark mode:
image


carlito-scality (approved):
LGTM but I wonder if we have to theme the tooltip and the "vertical cursor bar"

PR #137, opened 2019-11-19
Added <CloudProgressBar />

### LINKED ISSUE #124: Cloud ProgressBar
**Component**:Cloud ProgressBar

**Why this is needed**:

**What should be done**:
- We need a cloud shape progress bar in SVG with a text indicating the progress

image

**Test plan**:

edumoreira1506:
**Component**:

**Description**:
Created a CloudProgressBar component, related o issue #124 .

**Design**:
Peek 2019-11-19 14-44

---

Closes: #124

### DIFF
--- a/src/lib/components/cloudprogressbar/CloudProgressBar.component.js
+++ b/src/lib/components/cloudprogressbar/CloudProgressBar.component.js
@@ -27,13 +27,23 @@ const Container = styled.div`
 const ContainerProgress = styled.div`
   ${props => {
     return css`
+      @keyframes heightAnimation {
+        from {
+          height: 100%;
+        }
+        to {
+          height: ${100 - props.percentage}%;
+        }
+      }
+      animation-duration: 1s;
+      animation-fill-mode: both;
+      animation-name: heightAnimation;
       position: absolute;
       z-index: 222;
       top: 0;
       left: 0;
       display: block;
       width: 100%;
-      height: ${100 - props.percentage}%;
       overflow: hidden;
     `;
   }}
--- a/src/lib/components/cloudprogressbar/CloudProgressBar.component.js
+++ b/src/lib/components/cloudprogressbar/CloudProgressBar.component.js
@@ -1,0 +1,150 @@
+//@flow
+import React from 'react';
+import styled, { css } from "styled-components";
+import * as defaultTheme from "../../style/theme";
+
+export type CloudProgressBarProps = {
+  percentage: number,
+  borderSize: string,
+  color?: string,
+  progressColor?: string,
+  label?: string,
+  fontCOlor?: string,
+  fontSizePercentage?: string,
+  fontSizeLabel?: string
+};
+
+const Container = styled.div`
+  ${props => {
+    return css`
+      width: 100%;
+      text-align: center;
+      position: relative;
+    `;
+  }}
+`;
+
+const ContainerProgress = styled.div`
+  ${props => {
+    return css`
+      position: absolute;
+      z-index: 222;
+      top: 0;
+      left: 0;
+      display: block;
+      width: 100%;
+      height: ${100 - props.percentage}%;
+      overflow: hidden;
+    `;
+  }}
+`;
+
+const LabelPercentage = styled.span`
+  ${props => {
+    return css`
+      position: absolute;
+      top: 50%;
+      left: 50%;
+      margin-right: -50%;
+      transform: translate(-50%, -50%);
+      color: ${props.fontColor};
+      font-size: ${props.fontSize};
+      font-weight: 900;
+    `;
+  }}
+`;
+
+const Label = styled.span`
+  ${props => {
+    return css`
+      position: absolute;
+      top: 60%;
+      left: 50%;
+      margin-right: -50%;
+      transform: translate(-50%, -60%);
+      color: ${props.fontColor};
+      font-size: ${props.fontSize};
+    `;
+  }}
+`;
+
+const Cloud = ({
+  strokeColor,
+  borderSize
+}) =>
+  <svg 
+    xmlns="http://www.w3.org/2000/svg" 
+    viewBox="0 0 16 16"
+  >
+    <path 
+      d="m391.84 540.91c-.421-.329-.949-.524-1.523-.524-1.351 0-2.451 1.084-2.485 2.435-1.395.526-2.388 1.88-2.388 3.466 0 1.874 1.385 3.423 3.182 3.667v.034h12.73v-.006c1.775-.104 3.182-1.584 3.182-3.395 0-1.747-1.309-3.186-2.994-3.379.007-.106.011-.214.011-.322 0-2.707-2.271-4.901-5.072-4.901-2.073 0-3.856 1.202-4.643 2.925" 
+      transform="matrix(.77976 0 0 .78395-299.99-418.63)" 
+      fill="transparent"
+      stroke={strokeColor}
+      stroke-width={borderSize}
+      stroke-linejoin='round'
+    />
+  </svg>
+
+const CloudProgress = ({
+  strokeColor,
+  size,
+  percentage,
+  borderSize
+}) =>
+  <ContainerProgress
+    size={size}
+    percentage={percentage}
+  >
+    <svg 
+      xmlns="http://www.w3.org/2000/svg" 
+      viewBox="0 0 16 16"
+    >
+      <path 
+        d="m391.84 540.91c-.421-.329-.949-.524-1.523-.524-1.351 0-2.451 1.084-2.485 2.435-1.395.526-2.388 1.88-2.388 3.466 0 1.874 1.385 3.423 3.182 3.667v.034h12.73v-.006c1.775-.104 3.182-1.584 3.182-3.395 0-1.747-1.309-3.186-2.994-3.379.007-.106.011-.214.011-.322 0-2.707-2.271-4.901-5.072-4.901-2.073 0-3.856 1.202-4.643 2.925" 
+        transform="matrix(.77976 0 0 .78395-299.99-418.63)" 
+        fill="transparent"
+        stroke={strokeColor}
+        stroke-width={borderSize}
+        stroke-linejoin='round'
+      />
+    </svg>
+  </ContainerProgress>
+
+const CloudProgressBar = ({
+  percentage = 0,
+  borderSize = '2px',
+  color = defaultTheme.pink,
+  progressColor = defaultTheme.green,
+  label = '',
+  fontColor = defaultTheme.black,
+  fontSizePercentage = '10px',
+  fontSizeLabel = '5px'
+}: CloudProgressBarProps) =>
+  <Container
+    color={color}
+  >
+    <Cloud
+      strokeColor={progressColor}
+      borderSize={borderSize}
+    />
+    <CloudProgress
+      strokeColor={color}
+      percentage={percentage}
+      borderSize={borderSize}
+    />
+    <LabelPercentage
+      fontColor={fontColor}
+      fontSize={fontSizePercentage}
+    >
+      {`${percentage}%`}
+    </LabelPercentage>
+    <Label
+      color={fontColor}
+      fontSize={fontSizeLabel}
+    >
+      {label}
+    </Label>
+  </Container>
+
+export default CloudProgressBar;
--- a/src/lib/components/cloudprogressbar/CloudProgressBar.component.js
+++ b/src/lib/components/cloudprogressbar/CloudProgressBar.component.js
@@ -1,23 +1,30 @@
 //@flow
 import React from 'react';
-import styled, { css } from "styled-components";
+import styled, { css, keyframes } from "styled-components";
 import * as defaultTheme from "../../style/theme";
+import type { Node } from "react";
 
 export type CloudProgressBarProps = {
   percentage: number,
   borderSize: string,
-  color?: string,
-  progressColor?: string,
-  label?: string,
-  fontCOlor?: string,
-  fontSizePercentage?: string,
-  fontSizeLabel?: string
+  cloudColor?: string,
+  progressCloudColor?: string,
+  children: Node
+};
+
+const keyFrameCloud = props => {
+  return keyframes`
+    from {
+      height:100%
+    } 
+    to {
+      height:${100 - props.percentage}%
+    }`;
 };
 
 const Container = styled.div`
   ${props => {
     return css`
-      width: 100%;
       text-align: center;
       position: relative;
     `;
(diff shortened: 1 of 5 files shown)


src/lib/components/cloudprogressbar/CloudProgressBar.component.js
```diff
@@ -0,0 +1,160 @@
...
+      fill="transparent"
+      stroke={strokeColor}
+      stroke-width={borderSize}
+      stroke-linejoin='round'
+    />
+  </svg>
+
+const CloudProgress = ({
+  strokeColor,
+  size,
```
ChengYanJin: The `size`  prop looks unused, should be removed.

src/lib/components/cloudprogressbar/CloudProgressBar.component.js
```diff
@@ -0,0 +1,160 @@
+//@flow
+import React from 'react';
+import styled, { css } from "styled-components";
+import * as defaultTheme from "../../style/theme";
+
+export type CloudProgressBarProps = {
+  percentage: number,
+  borderSize: string,
+  color?: string,
```
ChengYanJin: IMO, Maybe rename the `color` to `cloudColor` and 
`progressColor` to `progressCloudColor` to make it more easy to understand?

src/lib/components/cloudprogressbar/CloudProgressBar.component.js
```diff
@@ -0,0 +1,160 @@
...
+        }
+        to {
+          height: ${100 - props.percentage}%;
+        }
+      }
+      animation-duration: 1s;
+      animation-fill-mode: both;
+      animation-name: heightAnimation;
+      position: absolute;
+      z-index: 222;
```
ChengYanJin: Why do we need z-index here?

ChengYanJin:
Hello, I am reviewing your PR.

src/lib/components/cloudprogressbar/CloudProgressBar.component.js
```diff
@@ -0,0 +1,160 @@
...
+        stroke-linejoin='round'
+      />
+    </svg>
+  </ContainerProgress>
+
+const CloudProgressBar = ({
+  percentage = 0,
+  borderSize = '2px',
+  color = defaultTheme.pink,
+  progressColor = defaultTheme.green,
```
ChengYanJin: It's better to get the default color from the theme 
```
color = defaultTheme.brand.primary
progressColor = defaultTheme.brand.backgroundContrast2
```

src/lib/components/cloudprogressbar/CloudProgressBar.component.js
```diff
@@ -0,0 +1,160 @@
...
+      text-align: center;
+      position: relative;
+    `;
+  }}
+`;
+
+const ContainerProgress = styled.div`
+  ${props => {
+    return css`
+      @keyframes heightAnimation {
```
ChengYanJin: It seems like all the CloudProgressBar has the same height in the storybook.
The problem is:

> If multiple keyframe sets exist for a given name, the last one encountered by the parser is used. @keyframes rules don't cascade, so animations never derive keyframes from more than one rule set.
 
I would propose that we use the helper `keyframes` from styled-component 
```
const keyFrameCloud = props => {
  return keyframes`
    from
    {
      height:100%
    } 
    to
    {
      height:${100 - props.percentage}%
    }`;
};
```

```
const ContainerProgress = styled.div`
...
  animation-name: ${keyFrameCloud};
...
`;
```

src/lib/components/cloudprogressbar/CloudProgressBar.component.js
```diff
@@ -0,0 +1,160 @@
...
+`;
+
+const Cloud = ({
+  strokeColor,
+  borderSize
+}) =>
+  <svg 
+    xmlns="http://www.w3.org/2000/svg" 
+    viewBox="0 0 16 16"
+  >
```
ChengYanJin, line 61: I notice an issue if we set the `percentage = 82%` or above, the cloud is almost full [screenshot](https://user-images.githubusercontent.com/18453133/69812298-217ec200-11f0-11ea-968c-fb982e3f7603.png)

We will need to change the attribute `viewBox` in order to set the same width/height as the path.

src/lib/components/cloudprogressbar/CloudProgressBar.component.js
```diff
@@ -0,0 +1,160 @@
...
+      position: absolute;
+      top: 60%;
+      left: 50%;
+      margin-right: -50%;
+      transform: translate(-50%, -60%);
+      color: ${props.fontColor};
+      font-size: ${props.fontSize};
+    `;
+  }}
+`;
```
ChengYanJin: We could remove both `Label` and `LabelPercentage` since we will specify it in the children. 
The idea is to design the component as generic as possible.

src/lib/components/cloudprogressbar/CloudProgressBar.component.js
```diff
@@ -0,0 +1,160 @@
...
+
+export type CloudProgressBarProps = {
+  percentage: number,
+  borderSize: string,
+  color?: string,
+  progressColor?: string,
+  label?: string,
+  fontCOlor?: string,
+  fontSizePercentage?: string,
+  fontSizeLabel?: string
```
ChengYanJin: In order to develop a more generic CloudProgressBar, I would propose to put all the stuff inside the cloud as children. So that the user can specify anything they want inside the cloud
`children: Node`

src/lib/components/cloudprogressbar/CloudProgressBar.component.js
```diff
@@ -0,0 +1,160 @@
...
+  </svg>
+
+const CloudProgress = ({
+  strokeColor,
+  size,
+  percentage,
+  borderSize
+}) =>
+  <ContainerProgress
+    size={size}
```
ChengYanJin: This looks unused, should be removed.

ChengYanJin (changes requested):
Thanks for the work!

It will be better if you can solve the issue described here:

-  The connection of the cloud is not very smooth[ on the top left, please see the red square](https://duckduckgo.com).
-  I noticed that all the CloudProgressBar in the storybook have the same height. I proposed one solution during the code review. https://github.com/scality/core-ui/pull/137#discussion_r351316924 
-  The percentage is above 80%, the cloud is almost full. You will need to figure out how to set the size for SVG with viewBox. Or maybe you have a better solution. [screenshot](https://user-images.githubusercontent.com/18453133/69812298-217ec200-11f0-11ea-968c-fb982e3f7603.png)

Let me know if you have any questions.

src/lib/components/cloudprogressbar/CloudProgressBar.component.js
```diff
@@ -0,0 +1,160 @@
...
+  borderSize = '2px',
+  color = defaultTheme.pink,
+  progressColor = defaultTheme.green,
+  label = '',
+  fontColor = defaultTheme.black,
+  fontSizePercentage = '10px',
+  fontSizeLabel = '5px'
+}: CloudProgressBarProps) =>
+  <Container
+    color={color}
```
ChengYanJin: Please remove the color prop if it's unused.
Please add `className='sc-cloudprogressbar'` as well

src/lib/components/cloudprogressbar/CloudProgressBar.component.js
edumoreira1506: Because we have "two layers" on the component, one is of complete cloud and other is of the progress area, so without this `z-index` the component doesn't work how expected.

edumoreira1506:
I believe than I fixed everything :thinking: 
Peek 2019-11-29 11-08

src/lib/components/cloudprogressbar/CloudProgressBar.component.js
ChengYanJin: I understand your point. But feel free to remove `z-index` to give a try.

IMO, in `<Cloud>` we use `progressCloudColor` which is the color of the complete cloud at the beginning. It will override by `<CloudProgress>` with the `cloudColor` which is the color of the bottom part of the cloud.

So I guess without using `z-index` is doable ;)

edumoreira1506: Nice! You was right, we don't need of `z-index`.

src/lib/components/cloudprogressbar/CloudProgressBar.component.js
```diff
@@ -0,0 +1,115 @@
...
+      />
+    </svg>
+  </ContainerProgress>
+
+const CloudProgressBar = ({
+  percentage = 0,
+  borderSize = '2px',
+  cloudColor = defaultTheme.brand.primary,
+  progressCloudColor = defaultTheme.brand.backgroundContrast2,
+  children = ''
```
ChengYanJin:
```suggestion
  children
```
IMO, it's not necessary to give a default value for children as it's a mandatory prop. 

Apart from that, it's better to specify the children in the storybook. The purpose of `storybook` is to give example/use case to the people who use this library.

src/lib/components/cloudprogressbar/CloudProgressBar.component.js
```diff
@@ -0,0 +1,115 @@
+//@flow
+import React from 'react';
+import styled, { css, keyframes } from "styled-components";
+import * as defaultTheme from "../../style/theme";
+
```
ChengYanJin, line 6: Please add
```suggestion
import type { Node } from "react";
```

stories/cloudprogressbar.js
```diff
@@ -0,0 +1,45 @@
...
+          cloudColor='#3d383a'
+          progressCloudColor='#83d136'
+        />
+      </div>
+      <div style={{ width: "500px", margin: '20px 0 20px 0' }}>
+        <CloudProgressBar
+          percentage={70}
+          borderSize='0.5px'
+          cloudColor='#BDFFB0'
+          progressCloudColor='#A3FFCB'
```
ChengYanJin, line 76: It's better to specify the `children` here
```
          <div
            style={{
              position: "absolute",
              top: "50%",
              left: "50%",
              transform: "translate(-50%,-50%)"
            }}
          >
            <Title>70%</Title>
            <SubTitle>of the quotas has been reached.</SubTitle>
          </div>
```

Don't forget to import the `SubTitle` so what the text can support light/dark mode ;)
`import { Wrapper, Title, SubTitle } from "./common";`

Please add children in all the clouds.

ChengYanJin (changes requested):
Great job!!! Almost all the issue have been addressed. 

Small reminder, there is still another issue we need to figure out( from last review)

-  The connection of the cloud is not very smooth[ on the top left, please see the red square](https://duckduckgo.com).

We should be good after this review! 
Good luck!

edumoreira1506:
Everything was fixed :smile: :rocket:
Peek 2019-12-02 09-53

stories/cloudprogressbar.js
```diff
@@ -0,0 +1,57 @@
...
+        <CloudProgressBar
+          percentage={50}
+          borderSize='0.5px'
+          cloudColor='#ccc'
+          progressCloudColor='#000'
+        >
+          <Title>50%</Title>
+          <SubTitle>of the quotas has been reached</SubTitle>
+        </CloudProgressBar>
+      </div>
```
ChengYanJin: Maybe I didn't make it clear in my last review. 
We still want to display the text `50% of the quotas has been reached` inside the cloud. So basically we will want to give a style to the text.

```     
    <div style={{ width: "300px", margin: "20px 0 20px 0" }}>
        <CloudProgressBar
          percentage={50}
          borderSize="0.5px"
          cloudColor="#ccc"
          progressCloudColor="#000"
        >
          <div
            style={{
              position: "absolute",
              top: "50%",
              left: "50%",
              transform: "translate(-50%,-50%)"
            }}
          >
            <Title>50%</Title>
            <SubTitle>of the quotas has been reached</SubTitle>
          </div>
        </CloudProgressBar>
      </div>
```

ChengYanJin (changes requested):
Great improvement!!! We are so close to the goal!!!
Two Minor changes we will be perfect!!!

1.  Last time maybe I didn't describe clearly what's our goal. I think you can easily tell the difference between the two screenshots. We will want the two parts of the clouds smoothly connect with each other.
[a. The screenshot from the issue description](https://user-images.githubusercontent.com/18453133/69981090-e18e4680-1531-11ea-87ac-4b350ba300fd.png)
[b. The screenshot of our current cloud](https://user-images.githubusercontent.com/18453133/69981012-b3106b80-1531-11ea-9c2b-54747ff2491b.png)

2. We still want to display the text in the cloud. Please take a look at my suggestion.

Thank you!

edumoreira1506:
@ChengYanJin . Thanks for the feedback! Look the new changes:
Peek 2019-12-02 15-32

src/lib/components/cloudprogressbar/CloudProgressBar.component.js
```diff
@@ -49,112 +48,85 @@ const ContainerProgress = styled.div`
...
   >
-    <path 
-      d="m391.84 540.91c-.421-.329-.949-.524-1.523-.524-1.351 0-2.451 1.084-2.485 2.435-1.395.526-2.388 1.88-2.388 3.466 0 1.874 1.385 3.423 3.182 3.667v.034h12.73v-.006c1.775-.104 3.182-1.584 3.182-3.395 0-1.747-1.309-3.186-2.994-3.379.007-.106.011-.214.011-.322 0-2.707-2.271-4.901-5.072-4.901-2.073 0-3.856 1.202-4.643 2.925" 
-      transform="matrix(.77976 0 0 .78395-299.99-418.63)" 
-      fill="transparent"
-      stroke={strokeColor}
-      stroke-width={borderSize}
-      stroke-linejoin='round'
-    />
+    <g id="_x2601__xFE0F__1_">
```
ChengYanJin: Please remove the id ;) same as another SVG please

src/lib/components/cloudprogressbar/CloudProgressBar.component.js
```diff
@@ -49,112 +48,85 @@ const ContainerProgress = styled.div`
...
 const Cloud = ({
   strokeColor,
   borderSize
 }) =>
   <svg 
-    xmlns="http://www.w3.org/2000/svg" 
-    viewBox="0 0 16 16"
+    xmlns="http://www.w3.org/2000/svg"
+    x="0px"
+    y="0px"
```
ChengYanJin: Could be removed I think ;) same as another SVG please

src/lib/components/cloudprogressbar/CloudProgressBar.component.js
```diff
@@ -0,0 +1,132 @@
...
+    } 
+    to {
+      height:${100 - props.percentage}%
+    }`;
+};
+
+const Container = styled.div`
+  ${props => {
+    return css`
+      width: 100%;
```
ChengYanJin: I guess `width:100%` can be removed

src/lib/components/cloudprogressbar/CloudProgressBar.component.js
```diff
@@ -0,0 +1,132 @@
...
+      display: block;
+      width: 100%;
+      overflow: hidden;
+    `;
+  }}
+`;
+
+const CloudContainer = styled.div`
+  position: relative;
+`;
```
ChengYanJin, line 51: I think we could remove the `<CloudContainer/>` and the style as well

src/lib/components/cloudprogressbar/CloudProgressBar.component.js
```diff
@@ -0,0 +1,132 @@
...
+const ContainerProgress = styled.div`
+  ${props => {
+    return css`
+      animation-duration: 1s;
+      animation-fill-mode: both;
+      animation-name: ${keyFrameCloud};
+      position: absolute;
+      top: 0;
+      left: 0;
+      display: block;
```
ChengYanJin: Why do we need it here? I guess could be removed.

ChengYanJin (changes requested):
Great!!! We achieved the goal already! The cloud looks good now!

If we could do some clean up it will be perfect!!! Please let me know if the changes I proposed makes sense to you. 

Thanks.

src/lib/components/cloudprogressbar/CloudProgressBar.component.js
edumoreira1506: Nop. We need of this to align the items, and for the two clouds be aligned as well.

src/lib/components/cloudprogressbar/CloudProgressBar.component.js
edumoreira1506: We need of this to align the items, and for the two clouds be aligned as well.

edumoreira1506:
:rocket:

src/lib/components/cloudprogressbar/CloudProgressBar.component.js
ChengYanJin: Position absolute means
> The element is positioned relative to its first positioned (not static) ancestor element.

If we remove the `<cloudContainer/>`, it's ancestor is `<container/>`which we already set ` position: relative;` So I think if we remove `<cloudContainer/>` should also work!

I will let you try it ;)

Issue #18, opened 2019-04-18
Collapsible Panel


nicolas2bert:
https://github.com/scality/pensieve-front/blob/master/src/react/ui-elements/Collapsible.jsx

Collapsed:
<img width="938" alt="Screenshot 2019-10-08 at 09 50 30" src="https://user-images.githubusercontent.com/47394132/66377131-3ccc0e80-e9b1-11e9-8b3a-f67ed167973d.png">

Expanded:
<img width="944" alt="Screenshot 2019-10-08 at 09 50 39" src="https://user-images.githubusercontent.com/47394132/66377142-43f31c80-e9b1-11e9-9acf-eb8fc6a05e9a.png">

![image](https://user-images.githubusercontent.com/47394132/69791726-67726080-11c5-11ea-8291-0a2a0dc80f96.png)

```
type Props = {
  header: Node,
  onHeaderClick: any => void,
  expanded: boolean,
  children: Node, //content
};

```

StheffanyHadlich:
Do you need help with this one? I can do it

carlito-scality:
> Do you need help with this one? I can do it

Yes! Definitely :) If you have any questions, let me know. Thanks

StheffanyHadlich:
@carlito-scality  is Node supposed to be the Header text?

carlito-scality:
> @carlito-scality is Node supposed to be the Header text?

@StheffanyHadlich good question :) We would like to cover this case as well:
![image](https://user-images.githubusercontent.com/47394132/69791611-28dca600-11c5-11ea-9737-0ce6e910b3fc.png)

Consequently, we should allow users to put what they want in the header :)

StheffanyHadlich:
:thinking: So maybe not a string, we could work with Node as an array and use flex to distribute its contents in Node container. What do you think?

carlito-scality:
> 🤔 So maybe not a string, we could work with Node as an array and use flex to distribute its contents in Node container. What do you think?

Yes good idea. It could be an array of Node :)

PR #141, opened 2019-11-28
Feature/collapsible panel

StheffanyHadlich:
**Component**:

Colapsible panel

**Description**:

Implements a collapsable panel

**Design**:

image

This component accepts a variant, a node and its children, a boolean prop to indicate if the panel is expanded or not and functions to controll `onClick` behaviours

**Breaking Changes**:

- [] Breaking Changes

---

See: #18


StheffanyHadlich:
@carlito-scality  I'm sendind this PR as WIP so you can have a look and comment while I keep going

package.json
```diff
@@ -19,6 +19,7 @@
   },
   "keywords": [],
   "devDependencies": {
+    "@scality/core-ui": "github:scality/core-ui.git",
```
carlito-scality: Why do you need this? Please remove it here and from the package-lock.json

src/lib/components/collapsiblepanel/CollapsiblePanel.component.js
```diff
@@ -0,0 +1,114 @@
+//@flow
+import React from "react";
+import { lighten } from "polished";
+import styled, { css } from "styled-components";
+import Button from "../button/Button.component";
+import { mergeTheme } from "../../utils";
+import * as defaultTheme from "../../style/theme";
+
+type Props = {
```
carlito-scality, line 7: Please stick with what is defined in the ticket description:
`header` could be an array of Node (since the content could be icon, string, or anything) as you suggested :)
```
type Props = {
  header: Array<Node>, 
  onHeaderClick: any => void,
  expanded: boolean,
  children: Node, //content
};
```

src/lib/components/collapsiblepanel/CollapsiblePanel.component.js
```diff
@@ -0,0 +1,114 @@
...
+      {
+        expanded && (
+          <ExpandedContainer>
+            {
+              children.map(item =>
+                <ExpandedItems key={item} onClick={onClick} variant={variant}>
+                  {item}
+                </ExpandedItems>
+              )
+            }
```
carlito-scality: As suggested, `children` could be anything. It is not necessary an array.
```
<ExpandedContainer> {children}</ExpandedContainer>
```
is enough :)

src/lib/components/collapsiblepanel/CollapsiblePanel.component.js
carlito-scality: `variant` is a good idea. We can manage it later in another PR :)

src/lib/components/collapsiblepanel/CollapsiblePanel.component.js
```diff
@@ -0,0 +1,114 @@
...
+  expanded = false,
+  variant =  'primary',
+  onHeaderClick,
+  onClick,
+  children = [],
+  node = '',
+}: Props) {
+  return (
+    <CollapsiblePanelContainer className="sc-collapsiblepanel" variant={variant} >
+      <HeaderContainer>
```
carlito-scality: `onClick={onHeaderClick}` should be here :) so if we can click anywhere in the header to open the panel.

src/lib/components/collapsiblepanel/CollapsiblePanel.component.js
```diff
@@ -0,0 +1,114 @@
...
+        <PanelText className="sc-chips-text">
+          {node}
+        </PanelText>
+        <Button
```
carlito-scality: Use `inverted` props ;)
```
 <Button
      icon={<i className={expanded ? ARROW.expanded : ARROW.colapsed} />}
      inverted
      size="larger"
    />
```

carlito-scality (changes requested):

src/lib/components/collapsiblepanel/CollapsiblePanel.component.js
StheffanyHadlich: When I use `inverted` my Icon gets the same colour as my background and it gets "invisible" only to appear on hover with a very light shade of gray. Those are the Button's default variant colours but they don't go well with my `background-color: defaultTheme.brand.primary ` , so I set this Button's `variant` as `base` because it gets a light shade of blue, from all the other `variants` I think this one is better but I don't know if it is the best approche here.

This is how it looks:

`inverted`

|` on hover default variant` |` variant set as base `|
|---------------------------------|-----------------------------|
| image image | image|

`!inverted`

| `on hover default variant` |
|---------------------------------|
| image image |

carlito-scality: oh! I think we don't need this Button, only icons are enough since, we moved `onHeaderClick` to `HeaderContainer`. Sorry I missed it

src/lib/components/collapsiblepanel/CollapsiblePanel.component.js
```diff
@@ -0,0 +1,88 @@
...
+import styled from "styled-components";
+import Button from "../button/Button.component";
+import * as defaultTheme from "../../style/theme";
+
+
+type Props = {
+  expand?: Boolean,
+  onHeaderClick?: any => void,
+  children?: any,
+  node: Array,
```
carlito-scality: rename `node ` to `header` since it is more meaningful

src/lib/components/collapsiblepanel/CollapsiblePanel.component.js
```diff
@@ -0,0 +1,88 @@
...
+  margin-right: 20px; 
+`;
+const HeaderIcon = styled.div`
+  margin: 0 20px;
+`;
+
+const CollapsiblePanelContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  border-radius: 8px 8px 4px 4px;
```
carlito-scality: `border-radius: 4px;` instead to be consistent

src/lib/components/collapsiblepanel/CollapsiblePanel.component.js
carlito-scality: rename `node` to `header` please

src/lib/components/collapsiblepanel/CollapsiblePanel.component.js
```diff
@@ -0,0 +1,88 @@
...
+  background-color: ${defaultTheme.brand.primary};
+  color: ${defaultTheme.brand.backgroundContrast1};
```
carlito-scality: `color: ${props => mergeTheme(props.theme, defaultTheme).text};` instead so we can support the dark mode

src/lib/components/collapsiblepanel/CollapsiblePanel.component.js
```diff
@@ -0,0 +1,88 @@
...
+`;
+const HeaderIcon = styled.div`
+  margin: 0 20px;
+`;
+
+const CollapsiblePanelContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  border-radius: 8px 8px 4px 4px;
+  background-color: ${defaultTheme.brand.primary};
```
carlito-scality: `background-color: ${props =>
    mergeTheme(props.theme, defaultTheme). backgroundContrast1};` to be themable

src/lib/components/collapsiblepanel/CollapsiblePanel.component.js
carlito-scality: They are all mandatory so please remove `?`

src/lib/components/collapsiblepanel/CollapsiblePanel.component.js
```diff
@@ -0,0 +1,88 @@
...
+function CollapsiblePanel({
+  expanded = false,
+  onHeaderClick,
+  children,
+  node = [],
+}: Props) {
+  return (
+    <CollapsiblePanelContainer className="sc-collapsiblepanel" >
+      <HeaderContainer onClick={onHeaderClick}>
+        <HeaderText className="sc-chips-text" >
```
carlito-scality: `sc-collapsiblepanel-text` instead of `sc-chips-text`

src/lib/components/collapsiblepanel/CollapsiblePanel.component.js
```diff
@@ -0,0 +1,88 @@
...
+            )
+          }
+        </HeaderText>
+        <HeaderIcon>
+          <Button
+            icon={<i className={expanded ? ARROW.expanded : ARROW.colapsed} />}
+            size="larger"
+          />
+        </HeaderIcon>
+      </HeaderContainer>
```
carlito-scality, line 73: please add `className="sc-collapsiblepanel-content" `

carlito-scality (changes requested):

src/lib/components/collapsiblepanel/CollapsiblePanel.component.js
StheffanyHadlich: No problem! I'll fix it later

stories/index.js
```diff
@@ -24,3 +24,4 @@ import "./vegachart";
 import "./progressbar";
 import "./textarea";
 import "./banner";
+import "./collapsiblepanel";
```
wenderjean: It seems you have a line ending problem here.

carlito-scality (approved):
Good job! thanks
FYI, I opened https://github.com/scality/core-ui/pull/142 to merge your PR.

PR #146, opened 2019-12-06
Feat/debounceinput

### LINKED ISSUE #103: DebounceInput
**Component**: DebounceInput

**Why this is needed**:
- We currently use ```react-debounce-input``` but it uses an old React version. This component will be deprecated soon. We need to get rid of this library and implement this component using the newest React version with Hooks

**What should be done**:

**Test plan**:

felipebrunetti94:
**Component**:
DebounceInput 

Gravação de tela de 06-12-2019 11_35_58

**Description**:
Only implemented the input with the delay and minLength trigger, if everything is okay I can substitute the "react-debounce-input" for the new component.

**Design**:
Text input with value label next to it.

Closes: #103


stories/index.js
```diff
@@ -19,3 +19,4 @@ import "./modal";
 import "./tooltip";
 import "./multiselect";
 import "./chips";
+import "./debounceinput";
```
wenderjean: You have a line ending problem here.

carlito-scality (changes requested):
I think this component has to be isofunctional as the `react-debounce-input` so we can remove `react-debounce-input` dependency . IMO, the simplest way to do that is to re-write [react-debounce-input](https://github.com/nkbt/react-debounce-input/blob/master/src/Component.js) using react Hooks

carlito-scality:
Thanks to https://github.com/nkbt/react-debounce-input/commit/3c6719a91dd2854b378987fba9ceedb83dff5bea, we do not need this PR anymore since it is compatible with React 16.8

carlito-scality:
No need anymore thanks to https://github.com/scality/core-ui/pull/155

PR #152, opened 2019-12-19
Feature/bar chart

ChengYanJin:
**Component**: BarChart

**Description**:
including:
- vertical stacked bar chart
- horizontal stacked bar chart 
- simple bar chart

**Design**:

Light mode:
image
image

Dark mode:
image
image


carlito-scality (changes requested):
As discuss IRL, it would be better to rename this component to `BarChart` and it should cover at least 3 charts: VerticalBarchart, HorizontalBarChart and StackedBarChart.

carlito-scality (approved):
LGTM

PR #155, opened 2019-12-30
Update the version of react-debounce-input

### LINKED ISSUE #103: DebounceInput

carlito-scality:
**Component**:Dependency

**Description**:
-Thanks to https://github.com/nkbt/react-debounce-input/pull/115, `react-debounce-input`  has no more deprecated lifecycle methods. We need to update the required version of `react-debounce-input` to `3.2.2`

**Design**:

**Breaking Changes**:

- [] Breaking Changes

---

Closes: https://github.com/scality/core-ui/issues/103


ChengYanJin (approved):
LGTM!

PR #180, opened 2020-04-23
Export only the sidebar

### LINKED ISSUE #178: New Layout - export only the sidebar
**Component**: Layout, Sidebar

**Why this is needed**:
We should modify our layout to meet the new Design. This layout should be applied by all the micro-apps.

**What should be done**:
- should remove the Navbar

TBD

**Test plan**:
TBD

ChengYanJin:
**Component**: lateralNavbarLayout

**Description**:

**Design**:
Hided Sidebar:
image

Expanded Sidebar:
image

How does it look like in the real application:
image

image

---

Closes: #178

### DIFF
--- a/src/lib/components/lateralnavbarlayout/LateralNavbarLayout.component.js
+++ b/src/lib/components/lateralnavbarlayout/LateralNavbarLayout.component.js
@@ -1,0 +1,41 @@
+//@flow
+import React from "react";
+import styled from "styled-components";
+import type { Node } from "react";
+import type { Props as SidebarProps } from "../sidebar/Sidebar.component";
+import Sidebar from "../sidebar/Sidebar.component";
+import { getThemePropSelector } from "../../utils";
+
+type Props = {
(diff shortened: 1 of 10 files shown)


gdemonet:
Great stuff! I like how it looks :)
Didn't test yet, but I was wondering, can we open the sidebar when hovering to the left?
Also, since the toggle is part of the navbar, does it mean we'll need to manage display of the sidebar of a micro-app from the Shell?

ChengYanJin:
> Great stuff! I like how it looks :)
> Didn't test yet, but I was wondering, can we open the sidebar when hovering to the left?
> Also, since the toggle is part of the navbar, does it mean we'll need to manage display of the sidebar of a micro-app from the Shell?

Hello Guillaume, 
Yeah. I can make the sidebar swipeable to display/hide when we hover to the left.
And, I think you are right. Anyway, we need to have a toggle to indicate the users to trigger the sidebar.

nicolas2bert (approved):

PR #185, opened 2020-05-06
Improvement/151 status bar component

### LINKED ISSUE #151: Status Bar Component
**Component**: StatusBar

**Why this is needed**:  We want to have a view of the global status

**What should be done**:
- different color represent different status that we can define by ourselves( For example, green stand for good, yellow stand for warning and red stand for error)
- support light/dark mode
- the labels to show recent several days
- have a title located in the top-left
- show the current status in the top-right
- when you hover a yellow or red zone, I think you should see a simplified list of alerts, when clicking on one, you are redirected to the corresponding enity and corresponding alert is highlighted in alert table 

image
image

**Test plan**:
TBD

ChengYanJin:
**Component**:
StatusBar

**Description**:
- Tooltip to display the date and alerts
- Click on the certain bar will be redirected to the alerts table
- Display the real date instead of the -7days...-6...today(but those labels is also doable)

- Expose the event handling in Vega-Embed. Currently, we are using hyperlink to redirect to the alert page. However, we would lose all the context. It's better to use the event handler.
> If you're using Vega-Embed, it returns a promise with a reference to the view which allows you to use addEventListener - explained in the docs here.
https://github.com/vega/vega-view#event-handling

**Design**:
image
image
image

---

Closes: #151


Cuervino:
Some remarks to discuss:

- For the tooltip, there is not starting point indicating what time-period is described, it might be useful for the user:
81199892-8b7f6c80-8fc3-11ea-8b4b-2db6403f21f9
Another solution could be that the focused time-period is circled.

- About "rounding" the bar, would that be possible to do it on the full bar level, and not on individual bar level ? (in order to have a closer look with use.design screen)
For instance : 
81199910-91754d80-8fc3-11ea-906d-f3e34c76457c

- About the tooltip, maybe the "formatting" indication could be removed, with information more simple:
81199892-8b7f6c80-8fc3-11ea-8b4b-2db6403f21f91

- In case that the tooltip is displayed from a time-period with more than 1 alert, it could notify how many alerts there are. Maybe an additional field, such as "Number of alert : 2"


- Just for checking the behaviour of the user action :
"Hoovering" with the mouse will display the tooltip.
"Clicking" will bring the user to the alert screen, in case it's a time-period with at least an alert.
In this case, I don't know how "obvious" it is for the user that he could click for going to the alert screen. Maybe a simple and discreet "_click to have more information about this alert_" (need to be discussed")

ChengYanJin:
Hello @Cuervino,
Thanks for the remarks!
Please find the improvement based on your comments:
image
Without choosing anything:
image
When I hover on one bar:
image


What I've changed:
- Change the opacity when hovering on one specific bar in order to indicate what time-period is described
- Tooltip: Remove the date formatting and rename it with `Date and Time`
- Tooltip: Add the `Number of alerts` field.

What's remain to be done:

> About "rounding" the bar, would that be possible to do it on the full bar level, and not on individual bar level ? (in order to have a closer look with use.design screen)

Yeah. This point was raised during the meeting as well, but I didn't figure out a way to change `cornerRadius` for only one first&last rectangular. 

> The behaviour of the user action

Good point! So I was considering to change the cursor to `pointer` when there is alert existed. 
However, I didn't find a way to change the status based on the condition. I will send a question to `Vega-lite` Slack channel to check if others have solution.

I'll write down those two points as an enhancement in a separate ticket. 

Don't hesitate to raise more remarks/improvements! Thanks!

MonPote (approved):
LGTM

PR #190, opened 2020-07-11
progressbar: Unable the default animation

ChengYanJin:
**Component**: progressbar

**Description**:
We don't want to have the animation for progress by default

**Design**:
Add a prop `isAnimation` to control the animation which is `false` by default


nicolas2bert:
Did you mean: "disable the default animation"?

nicolas2bert (approved):

PR #199, opened 2020-09-12
Tabs: Change the default color of the selected tabs and add activeTabColor

ChengYanJin:
**Component**: Tabs

**Description**:
And add `activeTabColor` property in the component to change at the app level.

Note that:
We need to keep the commit `23e3c8f904f0c4ab9ed77e8dfbcd7112cf4bc195` because it already merged in the Metalk8s
repo

### DIFF
--- a/src/lib/components/tabs/Tabs.component.js
+++ b/src/lib/components/tabs/Tabs.component.js
@@ -40,8 +40,7 @@ const TabItem = styled.div`
       css`
         &:hover {
           cursor: pointer;
-          background-color: ${props.activeTabColor ||
-          getTheme(props).primaryDark1};
+          background-color: ${props.activeTabColor || getTheme(props).primary};
         }
       `
     );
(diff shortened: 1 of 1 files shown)


src/lib/components/tabs/Tabs.component.js
```diff
@@ -33,13 +34,14 @@ const TabItem = styled.div`
...
-  ${props => {
+  ${(props) => {
     return (
       !props.selected &&
       css`
         &:hover {
           cursor: pointer;
-          background-color: ${getTheme(props).primary};
+          background-color: ${props.activeTabColor ||
+          getTheme(props).primaryDark1};
```
nicolas2bert: Why don't we keep `primary` for our default background color? This change will most likely break scality/zenko-ui.

src/lib/components/tabs/Tabs.component.js
```diff
@@ -50,12 +52,12 @@ const TabItemTitle = styled.p`
...
   padding: ${defaultTheme.padding.base} 0 16.5px;
-  ${props => {
-    const { textPrimary } = getTheme(props);
+  ${(props) => {
+    const { textPrimary, primaryDark1 } = getTheme(props);
     return props.selected
       ? css`
           color: ${textPrimary};
-          background-color: ${getTheme(props).primary};
+          background-color: ${props.activeTabColor || primaryDark1};
```
nicolas2bert: Same comment here.

src/lib/components/tabs/Tabs.component.js
ChengYanJin: ok. thanks for the review!

I changed back to `primary` as the default background color.
I will set `activeTabColor` in my case.

nicolas2bert (approved):

PR #210, opened 2020-12-03
HealthSelector: Add HealthSelector component

### LINKED ISSUE #204: Health Severity selection
**Component**:
Health Severity selection

**Why this is needed**:
We will have this component quite often in different applications and pages.

**What should be done**:

https://scality.invisionapp.com/console/share/W91PSTCEMX/495059238
image
image

Note that: 
- Compare with the design, we should remove the "Health" text for the selected item. 

- After choosing an item, the dropdown should be **closed** automatically. (We should not reuse the current `dropdown component`, because we don't want to display the items after choosing one item. But you can get some inspiration from there. )
- The three dots icon (red/yellow/green) ideally should come from CSS, since it doesn't exist in font awesome, and we should try to minimize the assets. 

**Test plan**:

alexis-ld:
**Component**:

HealthSelector

**Description**:

This adds a HealthSelector component as specified in #204 .

Upgrade Storybook to v5.2.8 to get bugfix of React Hook in story.

**Design**:
health-selector-demo

---

Closes: #204


src/lib/components/healthselector/Healthselector.component.js
```diff
@@ -0,0 +1,322 @@
...
+			<CSSCircle variant="danger" />
+		</CircleWrapper>,
+	];
+
+	return (
+		<HealthselectorContainer className="sc-healthselector">
+			<TriggerStyled
+				variant={variant || 'base'}
+				size={size || 'base'}
+				className="trigger"
```
ChengYanJin: the className would be better to rename to `sc-healthselector-trigger`, so it's kind of unique.

src/lib/components/healthselector/Healthselector.component.js
```diff
@@ -0,0 +1,322 @@
...
+						triggerSize={triggerSize}
+					>
+						<HealthSelectorMenuItem
+							onClick={items && items[0] && items[0].onClick}
+						>
+							<LeftRowWrapper>{icons[0]}</LeftRowWrapper>
+							<RightRowWrapper>
+								{(items[0] && items[0].label) || 'All'}
+							</RightRowWrapper>
+						</HealthSelectorMenuItem>
```
ChengYanJin: When we select an item, it should be removed from the `<HealthSelectorMenu/>`. So not all of the items should appear in the list/

src/lib/components/healthselector/Healthselector.component.js
```diff
@@ -0,0 +1,324 @@
...
+export type Item = {
+  label: string,
+  selected?: boolean,
+  onClick: (any) => void,
+};
+type Items = Array<Item>;
+
+type Props = {
+  size?: string,
+  variant?: string,
```
ChengYanJin: I believe we don't need this `variant` prop. We can remove it.

src/lib/components/healthselector/Healthselector.component.js
```diff
@@ -0,0 +1,324 @@
...
+      background-color: ${getThemePropSelector('primaryDark2')};
+    }
+    &:active {
+      background-color: ${getThemePropSelector('primaryDark2')};
+    }
+  `};
+`;
+
+function Healthselector(props: Props) {
+  const { variant, size, items } = props;
```
ChengYanJin: Please remove the variant too.

src/lib/components/healthselector/Healthselector.component.js
```diff
@@ -0,0 +1,324 @@
...
+            <HealthSelectorMenuItem
+              onClick={items && items[1] && items[1].onClick}
+            >
+              <LeftRowWrapper>{icons[1]}</LeftRowWrapper>
+              <RightRowWrapper>
+                {(items[1] && items[1].label) || 'Ok'}
+              </RightRowWrapper>
+            </HealthSelectorMenuItem>
+            <HealthSelectorMenuItem
+              onClick={items && items[2] && items[2].onClick}
```
ChengYanJin:
```suggestion
              onClick={items[2] && items[2].onClick}
```
We actually don't need to check items because we are sure it exists.

.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
+  "singleQuote": true,
+  "tabWidth": 2,
+  "trailingComma": "all"
+}
```
ChengYanJin: Please enter a new line.

ChengYanJin (changes requested):
Good job! 

There are two behaviors we need to improve: 
image

1. I feel like it makes sense to have this arrow telling me there are more choices. 
2. When I select an item, it should be removed from the list and appear as the triggered item.

src/lib/components/healthselector/__snapshots__/Healthselector.component.test.js.snap
```diff
@@ -0,0 +1,282 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
ChengYanJin, line 1: Would be better to separate the test into a specific commit and make sure to update the test every time change the code. 

I updated the test and pushed a commit^^

alexis-ld: Ok will do for next time :) Thanks for updating

ChengYanJin (approved):
LGTM!

PR #213, opened 2020-12-15
Input: Number type input

### LINKED ISSUE #211: Number selection Input
**Component**: Input

**Why this is needed**:
Since we introduce the number selection input in MetalK8s, we may want to add this component in this core-ui.
https://user-images.githubusercontent.com/18453133/101155678-bf894180-3627-11eb-9662-e1d0e247c7ab.png

**What should be done**:
Add number selection input 

Use the native <input/> HTML tag styled with our customized style. 

**Test plan**:
Test in the application.

alexis-ld:
**Component**:
Input

**Description**:

This adds a number type input as specified in #211  .

**Design**:

I kept the existing Input design for consistency and made it work with type="number" cases.

Screenshot 2020-12-15 at 15 56 14

---

Closes: #211

### DIFF
--- a/src/lib/components/input/Input.component.js
+++ b/src/lib/components/input/Input.component.js
@@ -1,5 +1,5 @@
 //@flow
-import React from "react";
+import React, { useRef } from "react";
 import { DebounceInput } from "react-debounce-input";
 import Checkbox from "../checkbox/Checkbox.component";
 import Select from "../select/Select.component";
@@ -8,12 +8,12 @@ import {
   InputContainer,
   LabelStyle,
   InputErrorMessage,
-  InputWrapper
+  InputWrapper,
 } from "./Input.component.style";
 
 export type Item = {
   label: string,
-  value: string | number
+  value: string | number,
 };
 
 type Items = Array<Item>;
@@ -26,8 +26,10 @@ type Props = {
   error?: string,
   id?: string,
   checked?: boolean,
-  onChange: () => void,
-  options?: Items
+  onChange: (e?: any) => void,
+  options?: Items,
+  min?: string,
+  max?: string,
 };
 
 const InputRenderer = ({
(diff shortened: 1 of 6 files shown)


stories/input.js
```diff
@@ -95,18 +96,27 @@ const ExampleInput = () => {
...
-        onChange={action("onChange")}
+        onChange={action('onChange')}
         type="textarea"
         placeholder="Write here..."
         error="error"
         value="Hello"
         cols={50}
         rows={5}
       />
+      <Title>Number Inputl</Title>
```
ChengYanJin:
```suggestion
      <Title>Number Input</Title>
```

src/lib/components/input/Input.component.js
```diff
@@ -1,19 +1,19 @@
 //@flow
-import React from "react";
-import { DebounceInput } from "react-debounce-input";
-import Checkbox from "../checkbox/Checkbox.component";
-import Select from "../select/Select.component";
-import TextArea from "../textarea/TextArea.component";
+import React, { Fragment } from 'react';
```
ChengYanJin: FYI, To simplify the code, we could use short syntax. See below,
https://reactjs.org/docs/fragments.html#short-syntax

But looks like you did not use Fragment in the code, should be removed.

src/lib/components/input/Input.component.js
```diff
@@ -1,19 +1,19 @@
...
+import { DebounceInput } from 'react-debounce-input';
+import Checkbox from '../checkbox/Checkbox.component';
+import Select from '../select/Select.component';
+import TextArea from '../textarea/TextArea.component';
```
ChengYanJin: Ha!! I realized that in core-ui, we use double quotes. 
I would suggest to remove   `"singleQuote": true,` from .prettierrc. Otherwise, it will bring too much noisy.

src/lib/components/input/Input.component.style.js
```diff
@@ -28,12 +28,35 @@ export const InputContainer = styled.div`
...
     box-shadow: inset 0 1px 1px rgba(0, 0, 0, 0.075),
       0 0 0 1px rgba(0, 126, 255, 0.1);
     outline: none;
   }
 
+  input.sc-number-input {
+    &::-webkit-inner-spin-button {
+      opacity: 0; // Disable default number input arrows
+    }
+  }
```
ChengYanJin: We should take care of Firefox as well.

src/lib/components/input/Input.component.style.js
```diff
@@ -28,12 +28,35 @@ export const InputContainer = styled.div`
...
+          color: ${textSecondary};
+        `;
+      }};
+      position: absolute;
+      right: ${defaultTheme.padding.smaller};
+      top: 50%;
+      transform: translate(-50%, -50%);
+    }
+  }
+
```
ChengYanJin: I understand you tried to set the opacity:0 to the original arrows so that we can't see them, but the functionality is still there. 

But it does not work well on my side. I need to be at a little left to the new icon `<i className="fas fa-sort"></i>` I am wondering is it working well on your side?

stories/input.js
```diff
@@ -95,18 +96,27 @@ const ExampleInput = () => {
...
+      <Input
+        id="id8"
+        value={numberValue}
+        onChange={(e) => setNumberValue(e.target.value)}
```
ChengYanJin: This cause a flow error in CI.
Something need to be change in the props definition.

  ```js
  onChange: (any) => void,
  ```
  
> Cannot get `e.target` because property `target` is missing in undefined [1].
>    stories/input.js:111:41
>    111|         onChange={(e) => setNumberValue(e.target.value)}

ChengYanJin (changes requested):
Some changes required. 

Most importantly, please test the arrows to see if increment and decrement work.

src/lib/components/input/Input.component.js
```diff
@@ -37,8 +39,44 @@ const InputRenderer = ({
...
+      window.HTMLInputElement.prototype,
+      "value",
+    );
+    if (valuePropDescriptor && inputEl.current) {
+      const nativeInputValueSetter = valuePropDescriptor.set;
+      if (nativeInputValueSetter) {
+        if (e.target.dataset.core === "up")
+          nativeInputValueSetter.call(
+            inputEl.current,
+            max && parseInt(value) + 1 <= parseInt(max)
```
ChengYanJin: if I don't specify `max`, I may not be able to increment?

alexis-ld: Indeed... fixing

ChengYanJin (changes requested):
Please add an error state number input in the storybook, and fix the bug of the carets are outside the input component. 
 
 Thanks

ChengYanJin (approved):
LGTM!
Great solution `align-items: baseline;`!

PR #215, opened 2020-12-16
Bugfix/214 fix memory leak on vega component

### LINKED ISSUE #214: VegaChart component tries to render the graph even if the component has been unmounted leading to memory leak issues
**Component**:

VegaChart

**What happened**:

While quickly navigating in MetalK8s ui I noticed following error (among several similar others due to several instance of the component being rendered in the page) :

```
index.js:1 Error: #node_cpu_usage_id does not exist
    at _callee4$ (embed.ts:263)
    at tryCatch (runtime.js:63)
    at Generator.invoke [as _invoke] (runtime.js:293)
    at Generator.next (runtime.js:118)
    at tslib.es6.js:74
    at new Promise (<anonymous>)
    at __awaiter (tslib.es6.js:70)
    at _embed (embed.ts:251)
    at _callee$ (embed.ts:224)
    at tryCatch (runtime.js:63)
    at Generator.invoke [as _invoke] (runtime.js:293)
    at Generator.next (runtime.js:118)
    at fulfilled (tslib.es6.js:71)
```

**What was expected**:

´VegaChart´ should handle the unmount of the component correctly and stop rendering of the chart if the component is unmounted.

**Steps to reproduce**

Run Metalk8s and quickly switch between ´Alerts´ and ´Metrics´ tab on nodes page.

**Resolution proposal** (optional):

We should add a cleanup function in this effect https://github.com/scality/core-ui/blob/development/1.0/src/lib/components/vegachart/VegaChart.component.js#L76 to interrupt vegaEmbed promise when the component is unmounted. However I didn't found a way to assign an abort controller or similar mechanism to vegaEmbed library to allow this so we might have to study another option maybe based on a ref to the element where we want vegaEmbed to actually render the graphs and in this way we would be able to abort the rendering in case of unmount.

JBWatenbergScality:
**Component**:

VegaChart and dependencies

**Description**:

Fix #214. Add a cleanup function to `VegaChart` component to avoid memory leaks on component unmount.

Closes: #214

### DIFF
--- a/.storybook/addons.js
+++ b/.storybook/addons.js
@@ -1,4 +1,4 @@
 import "@storybook/addon-actions/register";
 import "@storybook/addon-links/register";
 import "@storybook/addon-knobs/register";
-import "storybook-addon-styled-component-theme/dist/src/register";
+import "storybook-addon-styled-component-theme/dist/register";
--- a/src/lib/components/vegachart/VegaChart.component.js
+++ b/src/lib/components/vegachart/VegaChart.component.js
@@ -1,7 +1,8 @@
 //@flow
-import React, { useEffect, useContext } from "react";
+import React, { useEffect, useContext, useRef } from "react";
 import styled, { ThemeContext } from "styled-components";
-import vegaEmbed from "vega-embed";
+import vegaEmbed, {Result} from "vega-embed";
+
 
 type Props = {
   id: string,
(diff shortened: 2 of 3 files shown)


src/lib/components/vegachart/VegaChart.component.js
```diff
@@ -60,21 +61,36 @@ function VegaChart({ id, spec, theme = "light" }: Props) {
...
+        actions: false,
+      })
+      .then(result => {
+        vegaInstance.current = result;
       })
-      .catch(console.error);
-  }, [id, themedSpec, theme]);
+      .catch((...args) => {
+        if (isMounted) {
+          console.error(...args) // todo sould handle this with a retry or an error state of the component
```
ChengYanJin: A minor typo here
// TODO: we `should` handle this ...

src/lib/components/vegachart/VegaChart.component.js
```diff
@@ -1,7 +1,8 @@
 //@flow
```
ChengYanJin, line 1: Thanks for the fix. Did you check if it fixes the issue in MetalK8s UI. FYI, you can simply take this specific commit and replace the package.json in MetalK8s  "@scality/core-ui": "github:scality/core-ui.git#3a7cd45",

In terms of the format of the commit message, I would put the ticket number in the last end. 
> VegaChart: Add cleanup function in effect to avoid memory leaking 
>  
> (Here you can describe more what's the issue/ how to solve the issue.)
> 
> Refs: #214

.storybook/addons.js
```diff
@@ -1,4 +1,4 @@
 import "@storybook/addon-actions/register";
 import "@storybook/addon-links/register";
 import "@storybook/addon-knobs/register";
-import "storybook-addon-styled-component-theme/dist/src/register";
+import "storybook-addon-styled-component-theme/dist/register";
```
ChengYanJin, line 4: I would put this [link](https://github.com/echoulen/storybook-addon-styled-component-theme#legacy-configuration-v52-and-v4) in the commit message detail to explain why I need to perform this change.

package.json
```diff
@@ -68,10 +68,10 @@
     "@fortawesome/fontawesome-free": "5.7.2",
     "polished": "3.4.1",
     "react": "^16.8.0",
-    "react-dom": "^16.8.0",
     "react-debounce-input": "3.2.2",
-    "react-virtualized": "9.21.0",
```
ChengYanJin: If I understand correctly, there is no change in `package.json`, only in `package-lock.json`
I am wondering what's the reason you wanted to commit the changes in package.json? Maybe write a sentence in the commit message?

ChengYanJin (commented):
Bravo for the first PR !!! 

The code looks good for me, if you want to know my opinion on the commit message please see below ^^

package.json
JBWatenbergScality: I committed this file by mistake

ChengYanJin (approved):
LGTM! 

Bravo for the first PR at Scality!

src/lib/components/vegachart/VegaChart.component.js
```diff
@@ -60,21 +61,36 @@ function VegaChart({ id, spec, theme = "light" }: Props) {
...
-  }, [id, themedSpec, theme]);
+      .catch((...args) => {
+        if (isMounted) {
+          console.error(...args) // TODO: we should handle this with a retry or an error state of the component
+        }
+      });
+      
+      return () => {
+        isMounted = false;
+        if (vegaInstance.current) {
```
nicolas2bert, line 89: You might also want to make sure that the view property exists.

nicolas2bert (approved):
Congrats on your first PR! 🥇

MonPote:
Since you already got 2 approvals, you can just comment /approve :D

src/lib/components/vegachart/VegaChart.component.js
JBWatenbergScality: Thanks for sharing this however according to the typing of vega Result (https://github.com/vega/vega-embed/blob/master/src/embed.ts#L119) and the implementation view will always be defined on this object so I think we can safely skip this check thanks to flow/ts typings

PR #216, opened 2020-12-18
Feature/hoverable sidebar

### LINKED ISSUE #181: Add the hover possibility to expand the sidebar without the content resizing
**Component**: sidebar / lateral navbar

**Why this is needed**: 
The behaviour of the sidebar today is when we expend the sidebar, the size of the content becomes smaller.
We want to keep this behaviour at the same time, add another possibility to expand the sidebar. 

**What should be done**:
image
It's basically the same as Gmail does if you want to have a look. 

(New) When hovering over on the left, the sidebar overlap on the page without the content resizing 
(Already the case)When clicking on the expand button, the content resizes. 

Note: We may need to check if **layout** component of Core-ui  need to be changed.

TBD

**Test plan**:
Need to test the bebaviour in the application uses the sidebar from core-ui (MetalK8s UI)

alexis-ld:
**Component**:

Sidebar

**Description**:
This adds a 'hoverable' props to the sidebar. This prop makes it expand on mouse hover. It will still keep its 'click to toggle expand' functionality. The behavior is as defined in #181 : it doesn't resize the content on hover expand but does resize the content on click expand.

**Design**:
core-ui-hoverable-sidebar

Closes: #181

### DIFF
--- a/src/lib/components/sidebar/Sidebar.component.js
+++ b/src/lib/components/sidebar/Sidebar.component.js
@@ -1,5 +1,5 @@
 //@flow
-import React from "react";
+import React, { useState } from "react";
 import styled, { css } from "styled-components";
 import type { Node } from "react";
 import * as defaultTheme from "../../style/theme";
@@ -16,10 +16,12 @@ type Items = Array<Item>;
 export type Props = {
   expanded?: boolean,
   actions: Items,
+  hoverable?: boolean,
   onToggleClick?: () => void,
 };
 
-const SidebarContainer = styled.div`
+const Wrapper = styled.div`
+  flex-shrink: 0;
   ${(props) => {
     const { background, textPrimary } = getTheme(props);
     return css`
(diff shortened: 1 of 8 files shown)


src/lib/components/sidebar/Sidebar.component.js
```diff
@@ -16,10 +16,11 @@ type Items = Array<Item>;
 export type Props = {
   expanded?: boolean,
   actions: Items,
+  hoverable?: boolean,
   onToggleClick?: () => void,
 };
 
-const SidebarContainer = styled.div`
+const Wrapper = styled.div`
```
ChengYanJin, line 23: Looks like the Wrapper here doesn't include the entire height, only the items. 
I would suggest using zIndex here to make sure the sidebar always hover on the content of the page.

ChengYanJin (changes requested):
When I try the sidebar component in MetalK8s UI, I found some issues we may want to improve. 

https://user-images.githubusercontent.com/18453133/102770054-caf89e80-4383-11eb-95d4-0a9c49a3dbfa.mov

- The content of the page move to the left whereas we expect it doesn't move
- We should set the `z-index` of the sidebar to make sure it always visible. 
- And I need to check with UX too see if we need to have another background color for the sidebar.

alexis-ld:
@ChengYanJin I investigated the issue with MetalK8s integration.
Indeed there was a missing background and z-index on the Sidebar hoverable component, it is now fixed.

However the placement issue is specific to the Volumes and Nodes pages in metalk8s, it is something we should fix in metal when implementing this sidebar not in core-ui directly.

For instance here is the hoverable sidebar on the alerts page, it works as expected : 

https://user-images.githubusercontent.com/3480526/103890814-07006300-50e9-11eb-9aa1-091c179c5d68.mov

src/lib/components/sidebar/Sidebar.component.js
```diff
@@ -42,6 +44,44 @@ const SidebarContainer = styled.div`
...
+  ${(props) => {
+    const { background } = getTheme(props);
+
+    if (props.hoverable && props.hovered && !props.expanded) {
+      return css`
+        .sc-sidebar {
+          position: absolute;
+          height: 100%;
+          background-color: ${background};
+          z-index: 100;
```
ChengYanJin: I suggest we add this to `zIndex` in `theme.json`.
```js
export const zIndex = {
  tooltip: 9990,
  notification: 9000,
  modal: 8500,
  overlay: 8000,
  dropdown: 7000,
  nav: 500,
  sidebar: 100,
  base: 1,
};
```

ChengYanJin (approved):
LGTM! 
Thanks for testing it on MetalK8s UI side as well. 

Last one little suggestion remains, please define the zindex in theme.json/

src/lib/components/sidebar/Sidebar.component.js
alexis-ld: Done :)

PR #223, opened 2021-01-11
[Jest] Set maxWorkers to 1 in the CI to avoid tests randomly failing

ChengYanJin:
**Component**: tests, CI

**Description**:
We encoutered tests randomly failling on CI
ENOMEM: not enough memory, read

      at ScriptTransformer._transformAndBuildScript (node_modules/@jest/transform/build/ScriptTransformer.js:509:43)
      at ScriptTransformer.transform (node_modules/@jest/transform/build/ScriptTransformer.js:579:25)

Try the solution:
https://github.com/dequelabs/cauldron-react/issues/190#issuecomment-569962657

> For this issue, I think we should explore the maxWorkers jest cli option. After quick googling it appears that tons of people have "fixed" this issue by running the tests like with the flag -w 1 to allow for only 1 worker

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -26,7 +26,7 @@ stages:
           workdir: build/core-ui
           command: >
             npm ci &&
-            npm run test &&
+            npm run test -- -w 1 &&
             npm run flow &&
             npm run lint
           haltOnFailure: false


gdemonet:
(or bump the memory allocated to the CI worker running those tests?)

package.json
```diff
@@ -13,7 +13,7 @@
     "lint": "eslint src/* --ext .js",
     "prepare": "npm run build",
     "plop": "plop",
-    "test": "jest",
+    "test": "jest -w 1",
```
JBWatenbergScality: As this is only an issue occurring in the ci I would add this parameter as part of the main.yaml eve pipeline file (https://github.com/scality/core-ui/blob/development/1.0/eve/main.yml#L29 that would became `npm run test -- -w 1` )instead in order to keep fast test execution in local dev env.

ChengYanJin: Make sense!

JBWatenbergScality (approved):

PR #226, opened 2021-01-29
LineChart: Make Linechart config more flexible

alexis-ld:
**Component**: LineChart

**Description**:
This allows to be more flexible in LineChart configuration.
It allows to provide custom specs for :

- strokeDash
- opacity
- tooltip

If we don't provide any of those, the component will default to its default behavior, no breaking changes.

**Breaking Changes**:
None

### DIFF
--- a/src/lib/components/linechart/LineChart.component.js
+++ b/src/lib/components/linechart/LineChart.component.js
@@ -12,6 +12,10 @@ type Props = {
   width?: number,
   height?: number,
   displayTrendLine?: boolean,
+  strokeDashEncodingConfig?: Object,
+  opacityEncodingConfig?: Object,
+  tooltipConfig?: Object,
+  tooltipTheme?: string,
 };
 
 function LineChart({
@@ -25,6 +29,10 @@ function LineChart({
   height = 300,
   width = 1000,
   displayTrendLine = false,
+  strokeDashEncodingConfig,
+  opacityEncodingConfig,
+  tooltipConfig,
+  tooltipTheme,
   ...rest
 }: Props) {
   // hardcode the trendline configuration for tooltip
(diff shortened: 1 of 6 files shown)


src/lib/components/linechart/LineChart.component.js
```diff
@@ -12,6 +12,9 @@ type Props = {
   width?: number,
   height?: number,
   displayTrendLine?: boolean,
+  strokeDashConfig: Object,
```
ChengYanJin:
```suggestion
  strokeDashEncodingConfig?: Object,
```
Those new props should be optional IMO.
And I would rename it with `strokeDashEncodingConfig` to indicate this is the strokeDash channel for Encoding Property.

src/lib/components/linechart/LineChart.component.js
```diff
@@ -12,6 +12,9 @@ type Props = {
...
+  opacityConfig: Object,
```
ChengYanJin:
```suggestion
  opacityEncodingConfig?: Object,
```

src/lib/components/linechart/LineChart.component.js
```diff
@@ -12,6 +12,9 @@ type Props = {
...
+  tooltipConfig: Object,
```
ChengYanJin:
```suggestion
  tooltipConfig?: Object,
```

src/lib/components/linechart/LineChart.component.js
```diff
@@ -12,6 +12,9 @@ type Props = {
...
 };
```
ChengYanJin, line 19: According to the design, the background color of the tooltip should be dark in dark mode.

You may need to add optional `theme` property and pass it to <VegaChart/> component. 
`theme?: "light" | "dark",`


```js
return <VegaChart id={id} spec={spec} theme={theme}></VegaChart>
```

ChengYanJin (changes requested):
Hello @alexis-ld,
Thanks for taking care of the changes in core-ui!

Since the stroke dash line chart is missing today in core-ui and you spent a lot of effort to figure out the specs. I would suggest you create a story for that, as well as some necessary documentation. 

WDYT?

Another point you would need to take care of is the background color of the tooltip. 
Can you try the default dark theme of the tooltip? 
If the default dark theme is not great, we can even customize it. Please see the [documentation](https://github.com/vega/vega-tooltip/blob/master/docs/customizing_your_tooltip.md) if it's needed.

ChengYanJin (approved):
LGTM! 
Cheers!

PR #249, opened 2021-02-24
Improvement/upgrade storybook to 6.1

MonPote:
**Component**: All

**Description**:
- Update Storybook 5.2.8 -> 6.1.20
- Migrate stories from `storiesOf()` format to Component Story Format
- Add simple standalone MDX example (Introduction.mdx)
- Change the location of snapshot (use the "standard way")
- To change the Theme, you need to click on the circle on the top right of the top bar.
- Fix snapshot tests

**Notes**:

During the migration, I notice that we are barely using some add-ons in our `package.json`, we might want to invest some time to use it correctly later.

For example : 
- `@storybook/addon-knobs` are not correctly use and we might want to remove it later or use [`Controls`](https://storybook.js.org/docs/react/essentials/controls) instead.
- `@storybook/addon-links` are not currently used however we might need it in the future so I've keep it.
- `@storybook/theming` can also be an interesting add-ons if we want to brand a little bit more our storybooks.

To handle to theme, we are using a Wrapper in `stories/common.js`, we might want to double check if there is a better way to handle this with storybook.

We also might want to rework later the structure of the stories. But this will be done later when we have a clear picture of how it should look.
Moreover, most of stories are named `Default`, we might want to change/break them to a more explicit and specific one.

**Breaking Changes**: No breaking change for user. You will however need to follow the new way to write future stories
https://storybook.js.org/docs/react/api/csf
For the MDX/CSF stories, you can check what are the possibilities for the documentation here : https://github.com/storybookjs/storybook/blob/next/addons/docs/docs/recipes.md


stories/bannertest.stories.mdx
```diff
@@ -0,0 +1,39 @@
...
+      <Banner
+        variant="danger"
+        icon={<i className="fas fa-exclamation-triangle" />}
+        title={"Error"}
+      >
+        {"There is an error."}
+      </Banner>
+    </Wrapper>
+  </Story>
+</Canvas>
```
JBWatenbergScality: Discussion more than something related to the implementation itself : it could be interresting to be able to have a live playground in the documentation to see the impacts of the component when playing with a property. MDX support this thanks to a special code annotation  (https://mdxjs.com/guides/live-code)
``` 
```js live=true ```
```

Not sure through if this is supported out of the box by storybook but it may worth the try

.storybook/main.js
```diff
@@ -5,45 +5,3 @@ module.exports = {
     "storybook-addon-styled-component-theme/dist/register",
   ],
 };
-
-// import { configure, setAddon, addDecorator } from "@storybook/react";
```
JBWatenbergScality: This commit can be fixup with the first one

MonPote: I should have fixup some cleaning commits indeed sorry.

ChengYanJin (approved):
Thanks for the work! 
As you said there are a lot of cool features we may want to embed in the storybook. 
Let's merge this PR first!

JBWatenbergScality (approved):
LGTM !

PR #253, opened 2021-03-02
Improvement/1st improvement design system

Cuervino:
**Component**: storybook, design system, tests

**Description**:

A big PR but not that difficult to review. On the 6 commits only 2 are really important :
- `storybook: update storybook settings`
- `mdx: add first design system bricks`

You also might want to review it by launching the storybook.
You will need to `npm install` before `npm run storybook` because we added some changes to handle Windows environnement.

The rests are cosmetics changes and fix of tests.


.storybook/preview.js
```diff
@@ -32,3 +32,30 @@ const withThemeProvider = (Story, context) => {
...
+
+export const parameters = {
+  options: {
+    storySort: {
+      order: [
+        "Style",
+        ["Color", "Icons"],
+        "Components",
+        [
+          "Navigation",
```
nicolas2bert, line 44: Should it be ordered alphabetically for better readability?

package.json
```diff
@@ -8,7 +8,7 @@
   "mainSrc": "src/lib/index.js",
   "scripts": {
     "analyze": "source-map-explorer 'dist/*.js'",
-    "build": "rimraf dist && NODE_ENV=production babel src/lib --out-dir dist --copy-files && rimraf dist/**/*.test.js && rimraf dist/**/__snapshots__",
+    "build": "rimraf dist && cross-env NODE_ENV=production babel src/lib --out-dir dist --copy-files && rimraf dist/**/*.test.js && rimraf dist/**/__snapshots__",
```
nicolas2bert, line 11: What about using `envname` option to avoid having another dependency? https://babeljs.io/docs/en/options#envname

.prettierrc
```diff
@@ -1,4 +1,5 @@
 {
   "tabWidth": 2,
-  "trailingComma": "all"
```
nicolas2bert, line 3: Why do we need two different linter tools? ESLint should handle code formatting concerns.

stories/sparkline.stories.js
```diff
@@ -29,7 +29,7 @@ const id_multi = "multiVis";
 const lineConfig = { interpolate: "monotone" };
 
 export default {
-  title: "SparkLine",
+  title: "Components/Chart/SparkLine",
```
nicolas2bert, line 32: For maintainability purposes, we could assign variables for each topic/subtopic.
These values should be aligned with the `.storybook/preview.js` ones.

stories/icons.stories.mdx
```diff
@@ -0,0 +1,445 @@
...
+    <td>Sort-down</td>
+    <td>FA</td>
+    <td>Sort-down</td>
+    <td><i class="fas fa-sort-down"></i></td>
+    <td>solid Style (fas)</td>
+    <td>yes</td>
+    <td></td>
+    <td></td>
+  </tr>
+  <tr>
```
nicolas2bert, line 430: Not sure to understand this last row.

stories/guideline/button-guideline.stories.mdx
```diff
@@ -0,0 +1,140 @@
...
+
+Note for updating the coreUI sizes of buttons (22FEB21)
+
+|         | current | new |
+| ------- | ------- | --- |
+| Smaller | 27      | 24  |
+| Small   | 30      | 28  |
+| default | 37      | 32  |
+| large   | 40      | 36  |
+| larger  | 48      | 40  |
```
JBWatenbergScality: I think maybe this should be a github issue and we should only refers the current implemented variants in the actual design system documentation

stories/guideline/chart-guideline.stories.mdx
```diff
@@ -0,0 +1,64 @@
...
+### Bar and column charts
+
+- should start at a baseline of zero.
+
+<ChartExample />
+
+### Line charts
+
+- No more than 5–7 different lines
+- avoid randomly chosen color (otherwise risk if contrast issues)
```
JBWatenbergScality: nit: try to unify case of the first later for list items, eg: here `No` uses an upper case whereas `avoid` is lowercased

JBWatenbergScality:
```suggestion
- avoid randomly chosen color (otherwise risk of contrast issues)
```

stories/guideline/mdxExampleComponents.js
```diff
@@ -0,0 +1,54 @@
...
+  legend: {
+    direction: "horizontal",
+    orient: "top",
+    title: null,
+    symbolType: "circle",
+    columnPadding: 50,
+  },
+  scale: {
+    domain: ["2XX", "401", "404", "4XX", "503", "5XX"],
+    range: ["#4BE4E2", "#E45834", "#FEFA52", "#968BFF", "#BE2543", "#DC90F1"],
```
JBWatenbergScality, line 37: Shall we use a well defined color palette coming from the theme for charts ?

.storybook/preview.js
Cuervino: For headers on level 1 (`STYLE`, `COMPONENT`), the order should be predefined.
For the Intro and Style section, the order of levels 2 should be predefined.
For the library itself (inside `COMPONENTS`), it could be alphabetical order, however it makes sense to put `Navigation` at the beginning, to set `Input` and Selector next to each other, and also `Chart` and `Progress & loading` aside, as they are rather visual components.
At the end/bottom, it could be miscellaneous components.

image

stories/sparkline.stories.js
Cuervino: Yes, agree, and it would be easier to change/to maintain.
However, as it's a limited number of items for the moment, and as it might not save time to define variables, we can keep it like that for this version.

stories/icons.stories.mdx
Cuervino: Agree, not clear: it's not a real icon. 
In the case of a cell with no value in a table, we use a simple "hyphen" (`&#45;`) from the font, for still displaying there is no value, without having too much visual noise.
As it's still useful to know about this specific use case, this info could be added in the "Table" component/guideline.
image

stories/guideline/button-guideline.stories.mdx
Cuervino: Fully agree.

stories/guideline/mdxExampleComponents.js
Cuervino: Indeed, we do have a specific chart color palette.

.prettierrc
MonPote: On `metalk8s` and `xcore`. We use ESLint for code quality rules and prettier for formatting rules.
ESLint can indeed do both but I (subjective opinion) prefer prettier.

package.json
MonPote: We can, but I don't think it's an issue. `cross-env` is a widely use stable dependency to handle this use case.
And because it's only use during the build, it doesn't have an impact on the size on the build itself.

stories/guideline/button-guideline.stories.mdx
Cuervino: Issue #258 opened.

stories/guideline/mdxExampleComponents.js
Cuervino: We'll need to add the chart color palette, issue #259 opened.

package.json
nicolas2bert: A big reason why you use a library in the first place is to get the benefits of the code without having to worry about all the details. The drawback is that you’ve completely tied your code to these dependencies that you don’t own and don’t control. Instead, you could have used an existing babel option.

stories/Introduction.stories.mdx
```diff
@@ -8,4 +8,25 @@ Storybook Design System is a reusable component library that helps Storybook con
...
+
+## Objective
+
+A Design System is an ecosystem of guidelines, tools, components, which support teams ship more efficiently consistent design.
+
+- synchronising designers, product teams, and developers
+- building a shared vocabulary for reducing communication issues
+- having one solution for one component
+- easier testing at component level
+- faster iterations with established design patterns
```
JBWatenbergScality:
```suggestion
- Synchronising designers, product teams, and developers
- Building a shared vocabulary for reducing communication issues
- Having one solution for one component
- Easier testing at component level
- Faster iterations with established design patterns
```

stories/guideline/button-guideline.stories.mdx
```diff
@@ -0,0 +1,138 @@
...
+| larger  | 40  |
+
+## Padding
+
+Button’s label measurement:
+
+- line height: 20px
+- vertical padding: 8px
+- horizontal padding of 16px
+- space between icon & label: 8px
```
JBWatenbergScality:
```suggestion
- Line height: 20px
- Vertical padding: 8px
- Horizontal padding of 16px
- Space between icon & label: 8px
```

stories/guideline/button-guideline.stories.mdx
```diff
@@ -0,0 +1,138 @@
...
+_to produce_
+
+### Disabled button
+
+Disabled state only if an action on the current screen can enable it (action possibility).
+
+## Placement
+
+-Primary buttons placed on the right of a group.  
+-Dismissive actions placed on the left. It allows the user to return to the previous screen or step in the process.
```
JBWatenbergScality:
```suggestion
- Primary buttons placed on the right of a group.  
- Dismissive actions placed on the left. It allows the user to return to the previous screen or step in the process.
```

.prettierrc
nicolas2bert: I have never user Prettier.io and would love to have your opinion on "what is the advantage of using Prettier for formatting concerns over Eslint?". We could have a side discussion on this topic if you want.

stories/guideline/chart-guideline.stories.mdx
```diff
@@ -0,0 +1,64 @@
...
+- Avoid randomly chosen color (otherwise risk of contrast issues)
+
+Here, the colors for Get and List are too close.
+
+Components/Chart/LineChart
+
+Area charts
+
+- could be a single area, overlapping area, or stacked area
+- max 3-4 categories
```
JBWatenbergScality:
```suggestion
- Could be a single area, overlapping area, or stacked area
- Max 3-4 categories
```

stories/guideline/chart-guideline.stories.mdx
```diff
@@ -0,0 +1,64 @@
...
+- Should be a 100% total
+- Categories in the desc order
+
+_produce an example WIP_
+
+_(Currently, there is no Pie chart in the UI)_
+
+### Bar and column charts
+
+- should start at a baseline of zero.
```
JBWatenbergScality:
```suggestion
- Should start at a baseline of zero.
```

.storybook/preview.js
nicolas2bert: Thank you for this explanation.

stories/guideline/button-guideline.stories.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+<Story id="components-button--size" />
+
+|         | px  |
+| ------- | --- |
+| Smaller | 24  |
+| Small   | 28  |
+| default | 32  |
+| large   | 36  |
+| larger  | 40  |
```
JBWatenbergScality:
```suggestion
| Smaller | 24  |
| Small   | 28  |
| Default | 32  |
| Large   | 36  |
| Larger  | 40  |
```

JBWatenbergScality (approved):
Looks good to me 👍 , I purposed a set of suggestions to unify the case of the list items in each documentation but I let you see if it makes sense.

MonPote (approved):

PR #260, opened 2021-03-06
docs: Fix regression on props documentation

MonPote:
**Component**: documentation

**Description**:

A pretty simple PR, I've only add this kind of change for all stories.
```diff
// areachart.stories.js

export default {
  title: 'Components/Chart/AreaChart',
+  component: AreaChart,
};
```

The rest is prettier noise because we added `"singleQuote": true`

In the Docs tab you can see : 

alt text

In the Canvas tab 
alt text

You can also directly see the source in the Story tab
alt text

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,4 +1,8 @@
 module.exports = {
-  stories: ["../stories/**/*.stories.@(js|mdx)"],
-  addons: ["@storybook/addon-essentials", "@storybook/addon-knobs"],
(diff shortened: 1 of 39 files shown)


nicolas2bert (approved):
I would just have renamed the first commit.

MonPote:
> I would just have renamed the first commit.

Oh indeed thx, I forgot to change it after I added `@storybook/addon-storysource`.

PR #267, opened 2021-03-12
fix: Add aria-selected to navbar when options is selected

JBWatenbergScality:
**Component**:

Navbar

**Description**:

Add `aria-selected` to navbar options when options is selected


MonPote (approved):
LGTM

PR #274, opened 2021-03-19
Improvement: Custom scrollbar

alexis-ld:
**Component**: **Layout**

**Description**:

This enables scrollbar styling for every children within the **Layout** component.
We wanted to have a thinner scrollbar fitting the design better.

**For Chrome/Safari/Edge** we use the non-standard ::webkit-scrollbar CSS properties which are well supported across these browsers. The scrollbar is 4px wide and has a very small left padding to avoid the scrollbar being directly sticked to the scrollable data.

Before : 
chrome_before
After:
chrome_after

**For Firefox** we use the scrollbar-width and scrollbar-color properties. However these give less option than webkit and the only options for scrollbar-width are 'auto | thin | none' . There is no way to specify a pixel width for now. This has been set to 'thin', it is slightly wider than its webkit counterpart but is still way smaller than the original one.

Before :
firefox_before
After :
firefox_after

**Breaking Changes**: none

---

### DIFF
--- a/src/lib/components/layout/Layout.component.js
+++ b/src/lib/components/layout/Layout.component.js
@@ -1,24 +1,64 @@
 //@flow
-import React from "react";
-import styled from "styled-components";
-import type { Node } from "react";
-import { navbarHeight } from "../../style/theme";
-import Navbar from "../navbar/Navbar.component";
-import type { Props as NavbarProps } from "../navbar/Navbar.component";
-import type { Props as SidebarProps } from "../sidebar/Sidebar.component";
-import Sidebar from "../sidebar/Sidebar.component";
-import { getThemePropSelector } from "../../utils";
+import React from 'react';
+import styled, { css } from 'styled-components';
+import type { Node } from 'react';
+import { navbarHeight } from '../../style/theme';
+import Navbar from '../navbar/Navbar.component';
+import type { Props as NavbarProps } from '../navbar/Navbar.component';
+import type { Props as SidebarProps } from '../sidebar/Sidebar.component';
+import Sidebar from '../sidebar/Sidebar.component';
+import { getThemePropSelector, getTheme } from '../../utils';
 type Props = {
   navbar?: NavbarProps,
   sidebar?: SidebarProps,
   navbarElement?: Node,
-  children: Node
+  children: Node,
 };
 
 const LayoutContainer = styled.div`
   display: flex;
   flex-direction: column;
   height: 100vh;
+
+  ${(props) => {
+    const brand = getTheme(props);
+    return css`
+      // Custom scrollbar
+      * {
+        // Chrome / Safari / Edge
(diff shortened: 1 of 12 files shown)


src/lib/components/layout/Layout.component.js
```diff
@@ -1,24 +1,64 @@
...
+          display: none;
+        }
+        ::-webkit-scrollbar-corner {
+          background-color: transparent;
+        }
+
+        // Firefox
+        scrollbar-color: ${brand.border} ${brand.primary};
+        scrollbar-width: thin;
+      }
```
JBWatenbergScality: Ok for the approach ! But I think we should add it to `/src/lib/index.css` in order to make it available to applications that don't use this `Layout` component (we may have this use case for embedded applications)

alexis-ld: @JBWatenbergScality Ok, make sense. Do you have any suggestions on how to use the color variables (in this case primary and border) in a .css file ?

Cuervino:
Nice. On the screenshot, I like the Firefox one. 
It looks like the Chrome version is cut, especially as the round effect at the end is not complete. Maybe it's just a feeling.

Something like that:
image

alexis-ld:
> Nice. On the screenshot, I like the Firefox one.
> It looks like the Chrome version is cut, especially as the round effect at the end is not complete. Maybe it's just a feeling.
> 
> Something like that:
> image

@Cuervino indeed, probably linked to the background-clip :) will fix it

src/lib/components/layout/Layout.component.js
JBWatenbergScality: I see 2 options : 
1. Create a Scrollbar component independent from the layout and document that it should be used as a top level wrapper if we want an UI to be able to benefit from this scrollbar. For this approach restraining the `*` matcher to the root of this wrapper component could make sense
2. Introduce a theme component wrapper that, when introduced in the app, define css variables (https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) for each of the theme color. We then would be able to use those css variables directly in `index.css` 

IMO the first one will be easier to maintain and brings some benefits (decoupling the custom navbar to anything else) despite forcing us to add a wrapper in the UI where we want this custom navbar (but I think its ok and makes it explicit if one day we wonder where does this customisation comes from).

alexis-ld:
@ChengYanJin @JBWatenbergScality Rebased and moved the Scrollbar rules to their own wrapper component.

JBWatenbergScality (approved):

ChengYanJin (commented):
Hello @alexis-ld 
The changes look good to me!
Maybe add a note in the README to explain what change should we proceed with in order to have the styled scrollbar in all the components?

ChengYanJin (approved):
LGTM! Thanks for adding the README to explain how to use the customized scrollbar.

Issue #259, opened 2021-03-04
Add the Chart color palette and related guideline


Cuervino:
**Component**:

`color`, or a dedicated MDX page

**Why this is needed**:

Specific colors are used for charts lines or bars.

**What should be done**:

Having a guideline that is explaining what color to select, depending on the use case.

**Implementation proposal** (strongly recommended):

Explaining the use cases
- Type of chart,
- Meaning of data (positive, neutral),
- Number of lines

![image](https://user-images.githubusercontent.com/28654802/110005067-5f826400-7d18-11eb-93b3-47e4597fec8a.png)

**Test plan**:

PR #275, opened 2021-03-23
Update global health component

### LINKED ISSUE #186: Global Health Component enhancement
**Component**:
StatusBar

**Why this is needed**:
image
Need to improve the UX and appearance of the whole StatusBar.

**What should be done**:

Two points need to be addressed as an enhancement:
1. Set cornerRadius for only one first&last rectangular.
2. Change the cursor to pointer when there is alert existed.
3. We are using hyperlink to redirect to the alert page. However, we would lose all the context. It's better to use the event handler.

> If you're using Vega-Embed, it returns a promise with a reference to the view which allows you to use addEventListener - explained in the docs here.
https://github.com/vega/vega-view#event-handling
```
const width = 600
const color = blue
embed(element, {
    $schema: 'https://vega.github.io/schema/vega-lite/3.0.0-rc6.json',
    data: { 'values': data },
    mark: {
        type: 'line',
        color,
        point: {
            color,
        }
    },
    width,
    height: width / 2,
    encoding: {
        'x': {
            field: 'label',
            type: 'temporal',
        },
        'y': {
            field: 'value',
            type: 'quantitative',
        },
    }
}).then(({spec, view}) => {
    view.addEventListener('mouseover', function (event, item) {
        console.log(item.datum)
    })
})
``` 

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v5.json",
  "width": 650,
  "height": 18,
  "data": {
    "values": [
      {
        "severity": "warning",
        "start": "2021-02-02T15:00:00Z",
        "end": "2021-02-03T00:00:00Z",
        "description": "Logging service"
      },
      {
        "severity": "critical",
        "start": "2021-02-03T00:00:00Z",
        "end": "2021-02-04T00:00:00Z",
         "description": "Logging service"
      },
      {
        "severity": "warning",
        "start": "2021-02-04T10:00:00Z",
        "end": "2021-02-06T00:00:00Z"
      },
      {
        "severity": "warning",
        "start": "2021-02-06T12:00:00Z",
        "end": "2021-02-06T20:00:00Z"
      }
    ]
  },
  "config": {
    "view": {
      "stroke": "transparent"
    }
  },
  "transform": [  {"calculate": "datum.description !== '' ? 'View details from alart page' : ''", "as": "title"} ],
  "layer": [
    {
      "mark": {"type": "rect", "cornerRadius": 15},
      "encoding": {"color": {"value": "#28a745"}}
    },
    {
      "mark": {"type": "rect", "cursor": "pointer"},
      "encoding": {
        "x": {
          "field": "start",
          "type": "temporal",
          "title": null,
          "stack": null,
          "scale": {
            "domain": [
              {"date": 1, "month": 2, "year": 2021},
              {"date": 7, "month": 2, "year": 2021}
            ]
          },
          "axis": {
            "format": "%d %b",
            "ticks": false,
            "grid": false,
            "domainWidth": 0,
            "tickCount": 7,
            "labelExpr": "[timeFormat(datum.value, '%d'), timeFormat(datum.value, '%d') == '01' ? timeFormat(datum.value, '%b') : '']",
            "labelColor": "#a8b5c1"
          }
        },
        "x2": {"field": "end"}
      }
    },
    {
      "mark": {
        "type": "rect",
        "tooltip": true,
        "cornerRadius": 0,
        "cursor": "pointer"
      },
      "params": [{"name":"highlight","select":{"type":"point","on":"mouseover"}}],
      "encoding": {
        "x": {
          "timeUnit": "yearmonthdatehoursminutes",
          "field": "start",
          "type": "temporal",
          "title": null,
          "axis": {
            "format": "%d %b",
            "labelExpr": "timeFormat(datum.value, '%H') == '00'? datum.label:''",
            "labelColor": "#a8b5c1"
          }
        },
        "x2": {"field": "end"},
        "color": {
          "field": "severity",
          "type": "nominal",
          "title": "null",
          "scale": {"range": ["#dc3545", "#fefa52"]},
          "legend": null
        },
        "tooltip": [
          {"field": "severity", "title": "Severity"},
          {"field": "start", "title": "Start", "type": "temporal","timeUnit": "yearmonthdatehoursminutes"},
          {"field": "end", "title": "End", "type": "temporal","timeUnit": "yearmonthdatehoursminutes"},
          {"field": "title", "title": "title"},
          {"field":"description","title": "Description"}
        ],
        "opacity": {
          "condition": {"param": "highlight", "value": 1},
          "value": 0.6
        }
      }
    }
  ]
}
```

Design update with tooltip:
image

image

TBD

**Test plan**:
TBD

ChengYanJin:
**Component**: Global Health Component 

**Description**:
 As the first interaction, we decided to only display the global health alert exclude the alerts one layer below.

**Design**:
image

Update:
Screenshot 2021-03-29 at 10 55 57

**Update**: 
image

---

Closes: #186


src/lib/components/globalhealthbar/GlobalHealthBar.component.js
```diff
@@ -0,0 +1,139 @@
...
+  const spec = {
+    width,
+    height,
+    data: {
+      values: alertSegments,
+    },
+    transform: [
+      {
+        calculate:
+          "datum.description !== '' ? 'View details on alart page' : ''",
```
JBWatenbergScality:
```suggestion
          "datum.description !== '' ? 'View details on alert page' : ''",
```

src/lib/components/globalhealthbar/GlobalHealthBar.component.js
```diff
@@ -0,0 +1,139 @@
...
+
+type Props = {
+  id: string,
+  alertSegments: {
+    severity: 'critical' | 'warning' | 'none',
+    startSegment: string,
+    endSegment: string,
+    description: string,
+  }[],
+  start: { date: number, month: number, year: number },
```
JBWatenbergScality: Date format taken as an input should be uniformed between startSegement/endSegment and start/end likely use a JS Date object or an ISO string

src/lib/components/globalhealthbar/GlobalHealthBar.component.js
```diff
@@ -0,0 +1,139 @@
+//@flow
+import React from 'react';
+import VegaChart from '../vegachart/VegaChart.component.js';
+
+type Props = {
+  id: string,
+  alertSegments: {
+    severity: 'critical' | 'warning' | 'none',
+    startSegment: string,
```
JBWatenbergScality: I'm not sure we really want to pass a startSegment and endSegment, maybe giving the alert object directly makes more sense ? It would avoid each component using this component to reformat the alerts before passing them to the global health component

src/lib/components/globalhealthbar/GlobalHealthBar.component.js
```diff
@@ -0,0 +1,139 @@
...
+    endSegment: string,
+    description: string,
+  }[],
+  start: { date: number, month: number, year: number },
+  end: { date: number, month: number, year: number },
+  height?: number,
+  width?: number,
+  theme?: 'light' | 'dark' | 'custom',
```
JBWatenbergScality: It should be aware of the currently selected theme using useTheme hook, so I think we can drop this theme

src/lib/components/globalhealthbar/GlobalHealthBar.component.js
```diff
@@ -0,0 +1,139 @@
+//@flow
+import React from 'react';
+import VegaChart from '../vegachart/VegaChart.component.js';
+
+type Props = {
```
JBWatenbergScality: exporting prop type could be useful to ease app integration 
```suggestion
export type GlobalHealthProps = {
```

src/lib/components/globalhealthbar/GlobalHealthBar.component.js
```diff
@@ -0,0 +1,139 @@
...
+        calculate:
+          "datum.description !== '' ? 'View details on alart page' : ''",
+        as: 'title',
+      },
+    ],
+    layer: [
+      // Paint the entire bar with green
+      {
+        mark: { type: 'rect', cornerRadius: 15 },
+        encoding: { color: { value: '#28a745' } },
```
JBWatenbergScality: We should use a well known theme color (https://github.com/scality/core-ui/issues/259)

src/lib/components/globalhealthbar/GlobalHealthBar.component.js
```diff
@@ -0,0 +1,139 @@
...
+            },
+            axis: {
+              format: '%d %b',
+              ticks: false,
+              grid: false,
+              domainWidth: 0,
+              tickCount: 7,
+              labelExpr:
+                "[timeFormat(datum.value, '%d'), timeFormat(datum.value, '%d') == '01' ? timeFormat(datum.value, '%b') : '']",
+              labelColor: '#a8b5c1',
```
JBWatenbergScality: We should use a well known theme color (#259)

src/lib/components/globalhealthbar/GlobalHealthBar.component.js
```diff
@@ -0,0 +1,139 @@
...
+  id,
+  alertSegments,
+  start,
+  end,
+  height = 16,
+  width = 650,
+  theme = 'dark',
+  tooltipPosition = 'up',
+}: Props) {
+  const spec = {
```
JBWatenbergScality, line 56: You may be missing a view entry in your spec to clip the segment between the border radius  (I believe that in the current state if a segment is raised at the beginning or end of the bar the border will be squared). You can refer to https://github.com/scality/core-ui/issues/186#issuecomment-792960267 and more specifically this spec section : 
```
"view": {"cornerRadius": 15},
```
and also the layer should then be clipped :
```
mark: {
          type: 'rect',
          tooltip: true,
          cornerRadius: 0,
          cursor: 'pointer',
          clip: true /// <------- here 
        },
        params: [
          { name: 'highlight', select: { type: 'point', on: 'mouseover' } },
        ],
```

src/lib/components/vegachart/VegaChart.component.js
```diff
@@ -1,20 +1,23 @@
...
 
+/* How to theme tooltip:
+   https://github.com/vega/vega-tooltip/blob/master/docs/customizing_your_tooltip.md */
 type Props = {
   id: string,
   spec: Object,
-  theme?: string, // predefine theme light/dark, or 'custom'
-  // How to theme tooltip:
-  // https://github.com/vega/vega-tooltip/blob/master/docs/customizing_your_tooltip.md)
+  theme?: 'light' | 'dark' | 'custom', //predefine tooltip theme
```
JBWatenbergScality: same as before I believe deducing the theme form useTheme is a better approach

src/lib/components/globalhealthbar/GlobalHealthBar.component.js
JBWatenbergScality: The global health component would then be responsible of segment splitting

JBWatenbergScality (changes requested):

Cuervino:
1. Time legend - Nicer to have "01 Feb" on the same line (and not on 2 lines)

2. Proposal: maybe add super small stick bars for having a better period division
image

3. A little bit disturbing for me that there is no action/popup on the green sections (even with a "0 alert" tooltip), but that's just a feeling

4. Hovering behavior issue/bug:
When hovering on, the focus (color highlight) + the tooltip
When hovering out, the focus stays (it should come back to normal), and the tooltip disappears (correct). Currently, no way to switch back to the "original" status.
https://user-images.githubusercontent.com/28654802/112181861-1463be80-8bfd-11eb-9799-12a2282b295c.mp4

5. For the popup/tooltip

- Nicer to change the border to 313131
- I would propose to remove the opacity of the bg (on the PR it's 90%, but as we are not using it currently anywhere else in the UI, I would avoid it)
- The font Family is not Lato, I'd change it for Lato
image

The design of the pop-up might be enhanced later on (for instance, adding an icon next to severity).

src/lib/components/globalhealthbar/GlobalHealthBar.component.js
ChengYanJin: Yes. I changed to ISO string.  
And then format it to this { date: number, month: number, year: number } which is required by the vega-lite library.

ChengYanJin:
> 1. Time legend - Nicer to have "01 Feb" on the same line (and not on 2 lines)

Haven't found a solution for this. But I think it's not a block for the moment, maybe we can create a ticket to tract all the improvement need to be done around global health component.

> 2. Proposal: maybe add super small stick bars for having a better period division
>    image

(Please check the change). I try to achieve the best I can do.
Screenshot 2021-03-29 at 10 55 57

> 3. A little bit disturbing for me that there is no action/popup on the green sections (even with a "0 alert" tooltip), but that's just a feeling

What I can do is display the same tooltip(without description and title, because no alert trigger in this period)

> 4. Hovering behavior issue/bug:
>    When hovering on, the focus (color highlight) + the tooltip
>    When hovering out, the focus stays (it should come back to normal), and the tooltip disappears (correct). Currently, no way to switch back to the "original" status.
>    https://user-images.githubusercontent.com/28654802/112181861-1463be80-8bfd-11eb-9799-12a2282b295c.mp4

Issue solved. Thanks for pointing this issue!

> 5. For the popup/tooltip
> 
> * Nicer to change the border to 313131
> * I would propose to remove the opacity of the bg (on the PR it's 90%, but as we are not using it currently anywhere else in the UI, I would avoid it)
> * The font Family is not Lato, I'd change it for Lato
>   image
> 
> The design of the pop-up might be enhanced later on (for instance, adding an icon next to severity).

Customized the tooltip.

src/lib/components/globalhealthbar/GlobalHealthBar.component.js
```diff
@@ -0,0 +1,167 @@
...
+              domain: [startTimeObject, endTimeObject],
+            },
+            axis: {
+              format: '%d %b',
+              ticks: true,
+              grid: false,
+              domainWidth: 0,
+              tickCount: 7,
+              labelExpr:
+                "[timeFormat(datum.value, '%d'), timeFormat(datum.value, '%d') == '01' ? timeFormat(datum.value, '%b') : '']",
```
JBWatenbergScality: shouldn't we always display the month for the first point ? (for example a graph starting on 7 February and ending on 14th won't display any month information so I'm wondering if displaying the month on the first displayed day could help to understand that we are displaying dates here )

ChengYanJin: Update with:
image

JBWatenbergScality (approved):

PR #281, opened 2021-03-29
Doc - format for number and dates

Cuervino:
**Component**:

MDX documentation on Format
Style>Format

**Description**:
Add 1 page with a guideline about number format and date format, in order to have consistency in the UI

**Design**:
image
image

**Breaking Changes**:
no
- [] Breaking Changes

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/stories/format.stories.mdx
+++ b/stories/format.stories.mdx
@@ -19,22 +19,21 @@ import { Meta } from '@storybook/addon-docs/blocks';
 | EEE  | Weekday on 3 letters, first with uppercase | "Mon"   |
 | DD   | Day of the month with leading zero         | 08      |
 | MM   | Numerical month with leading zero          | 07      |
-| MMM  | Month abbreviation on 3 letters uppercase  | "JUL"   |
+| MMM  | Month abbreviation on 3 letters uppercase  | "Jul"   |
 | HH   | Hour [0-23]                                | 13      |
 | mm   | Minute with leading zero                   | 05      |
 | ss   | Second with leading zero                   | 06      |
 
 ### Format list
 
-| Alias   | code                     | Example                  | Length | Context                                              |
-| ------- | ------------------------ | ------------------------ | ------ | ---------------------------------------------------- |
-| Short#1 | DD MMM                   | 20 Jul                   | 5      | Chart time axis                                      |
-| Short#2 | DDMMM HH:mm              | 20JUL 09:00              | 11     | Limited space - Dense tables, charts                 |
-| Short#3 | YYYY-MM-DD               | 2020-07-20               | 10     |                                                      |
-| Mid#1   | EEE YYYY-MM-DD HH:mm     | Mon 2020-07-20 09:00     | 20     | Checking with agenda for day of the week             |
-| Mid#2   | YYYY-MM-DD HH:mm         | 2020-07-20 09:00         | 16     | No need of day of the week                           |
-| Mid#3   | YYYY-MM-DD HH:mm:ss      | 2020-07-20 09:00:00      | 19     | When the seconds are needed                          |
-| Full#1  | EEE MMM DD YYYY HH:mm:ss | Mon Jul 20 2020 09:00:00 | 24     | When a lot of space (hover) When precision is needed |
+| Alias   | code                     | Example                  | Length | Context                                                |
+| ------- | ------------------------ | ------------------------ | ------ | ------------------------------------------------------ |
+| Short#1 | DD MMM                   | 20 Jul                   | 5      | Chart time axis                                        |
+| Short#2 | DDMMM HH:mm              | 20Jul 09:00              | 11     | Limited space, year not needed                         |
+| Short#3 | YYYY-MM-DD               | 2020-07-20               | 10     | Tables                                                 |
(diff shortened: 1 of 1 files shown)


stories/format.stories.mdx
```diff
@@ -0,0 +1,44 @@
...
+
+| Alias   | code                     | Example                  | Length | Context                                              |
+| ------- | ------------------------ | ------------------------ | ------ | ---------------------------------------------------- |
+| Short#1 | DD MMM                   | 20 Jul                   | 5      | Chart time axis                                      |
+| Short#2 | DDMMM HH:mm              | 20JUL 09:00              | 11     | Limited space - Dense tables, charts                 |
+| Short#3 | YYYY-MM-DD               | 2020-07-20               | 10     |                                                      |
+| Mid#1   | EEE YYYY-MM-DD HH:mm     | Mon 2020-07-20 09:00     | 20     | Checking with agenda for day of the week             |
+| Mid#2   | YYYY-MM-DD HH:mm         | 2020-07-20 09:00         | 16     | No need of day of the week                           |
+| Mid#3   | YYYY-MM-DD HH:mm:ss      | 2020-07-20 09:00:00      | 19     | When the seconds are needed                          |
+| Full#1  | EEE MMM DD YYYY HH:mm:ss | Mon Jul 20 2020 09:00:00 | 24     | When a lot of space (hover) When precision is needed |
```
JBWatenbergScality: Shouldn't this be internationalised and for example use `MM/DD/YYYY` in the US, and `DD/MM/YYYY` in Europe ?

Cuervino: Yes, from the long-term point of view. For that, we would need to add more "clues" about whether we are displaying US format or non-US format.
Having an English language and just for example "12/11/2021" is not enough for knowing if it's December or November of 2021. So we might need to add more information (letters, like 12NOV2021, etc.).
To avoid ambiguity, the easier and most straight forward way would be to use the YYYY-MM-DD format for the moment.

JBWatenbergScality (approved):

stories/format.stories.mdx
```diff
@@ -0,0 +1,44 @@
...
+| ss   | Second with leading zero                   | 06      |
+
+### Format list
+
+| Alias   | code                     | Example                  | Length | Context                                              |
+| ------- | ------------------------ | ------------------------ | ------ | ---------------------------------------------------- |
+| Short#1 | DD MMM                   | 20 Jul                   | 5      | Chart time axis                                      |
+| Short#2 | DDMMM HH:mm              | 20JUL 09:00              | 11     | Limited space - Dense tables, charts                 |
+| Short#3 | YYYY-MM-DD               | 2020-07-20               | 10     |                                                      |
+| Mid#1   | EEE YYYY-MM-DD HH:mm     | Mon 2020-07-20 09:00     | 20     | Checking with agenda for day of the week             |
```
thomasdanan: Do we really need Mid#1, not sure where we are using it currently?
I guess we should try to reduce this list as much as possible ...

stories/format.stories.mdx
```diff
@@ -0,0 +1,44 @@
...
+| HH   | Hour [0-23]                                | 13      |
+| mm   | Minute with leading zero                   | 05      |
+| ss   | Second with leading zero                   | 06      |
+
+### Format list
+
+| Alias   | code                     | Example                  | Length | Context                                              |
+| ------- | ------------------------ | ------------------------ | ------ | ---------------------------------------------------- |
+| Short#1 | DD MMM                   | 20 Jul                   | 5      | Chart time axis                                      |
+| Short#2 | DDMMM HH:mm              | 20JUL 09:00              | 11     | Limited space - Dense tables, charts                 |
```
thomasdanan: Is it really the one we want to use in dense tables? I was thinking the MMM was causing some alignment issues and that it was preferable to use numeric format?

stories/format.stories.mdx
```diff
@@ -0,0 +1,44 @@
...
+| DD   | Day of the month with leading zero         | 08      |
+| MM   | Numerical month with leading zero          | 07      |
+| MMM  | Month abbreviation on 3 letters uppercase  | "JUL"   |
+| HH   | Hour [0-23]                                | 13      |
+| mm   | Minute with leading zero                   | 05      |
+| ss   | Second with leading zero                   | 06      |
+
+### Format list
+
+| Alias   | code                     | Example                  | Length | Context                                              |
```
thomasdanan: more generally, I tend to think we should limit as much as possible the number of formats we use in the UIs ...

ChengYanJin (commented):
I feel it would interesting for me to some examples, meaning in which case I should use this format.

stories/format.stories.mdx
Cuervino: Correct, I reduced the list to 6.

stories/format.stories.mdx
Cuervino: Correct, I change the "table" use cases.

stories/format.stories.mdx
Cuervino: I still think that it makes sense to have the standard date with 2 formats: with and without seconds.
Space can be very scarce.

Cuervino:
> 
> 
> I feel it would interesting for me to some examples, meaning in which case I should use this format.
What do you mean exactly?
You mean associated screens? Visual exemples? It may be in a next version of this page.

ChengYanJin (approved):

PR #290, opened 2021-04-06
feat: add possibility nested dropdown

### LINKED ISSUE #277: Navbar : Dropdown in a dropdown
**Component**:

Dropdown

**Why this is needed**:

We should be able to add a dropdown as an item of a dropdown to enable nested list in navbar (theme and language example below).

image

romainlancia:
**Component**:
Dropdown
**Description**:
Add possibility to put a Dropdown as an item of Dropdown
**Design**:
https://user-images.githubusercontent.com/47044686/115277099-c37fb100-a13b-11eb-80bc-bb2687e69221.mp4

Closes: #277

### DIFF
--- a/src/lib/components/dropdown/Dropdown.component.js
+++ b/src/lib/components/dropdown/Dropdown.component.js
@@ -1,5 +1,5 @@
 //@flow
-import React, { useState, useCallback } from "react";
+import React, { useState, useCallback, forwardRef } from "react";
 import type { Node } from "react";
 import styled, { css } from "styled-components";
 import {
@@ -8,23 +8,45 @@ import {
   ButtonText
 } from "../button/Button.component";
 import * as defaultTheme from "../../style/theme";
-import { getThemePropSelector } from "../../utils";
+import { getThemePropSelector } from '../../utils';
+import { getPositionDropdownMenu } from './utils';
 
 export type Item = {
   label: string,
   name?: string,
   selected?: boolean,
-  onClick: any => void
+  onClick?: any => void,
+  submenuIcon?: Node,
+  submenuItems?: Array<Item>
 };
+
+type DropdownTriggerContainerProps = {
+  isItem?: boolean,
+  dataIndex: number,
+  open?: boolean,
+  size?: string,
+  variant?: string,
+  title?: string,
+  onBlur: any => void,
+  onFocus: any => void,
+  onClick: any => void,
+  onMouseEnter: any => void,
+  onMouseLeave: any => void,
+  children: Node,
+}
+
 type Items = Array<Item>;
 type Props = {
+  isItem?: boolean,
   text?: string,
   size?: string,
   variant?: string,
   title?: string,
   items: Items,
   icon?: Node,
-  caret?: boolean
+  caret?: boolean,
+  dataIndex?: number,
+  onClick?: any => void,
 };
 
 const DropdownStyled = styled.div`
@@ -41,43 +63,19 @@ const DropdownMenuStyled = styled.ul`
   position: absolute;
   margin: 0;
   padding: 0;
-  border: 1px solid ${getThemePropSelector("primary")};
   z-index: ${defaultTheme.zIndex.dropdown};
   max-height: 200px;
   min-width: 100%;
-  overflow: auto;
 
-  ${props => {
-    if (
-      props.size &&
-      props.triggerSize &&
-      props.triggerSize.x + props.size.width > window.innerWidth
-    ) {
-      return css`
-        right: 0;
-        top: 100%;
-      `;
-    } else if (
-      props.size &&
-      props.triggerSize &&
-      props.triggerSize.y + props.size.height > window.innerHeight
-    ) {
-      return css`
-        left: 0;
-        bottom: ${props.triggerSize.height + "px"};
-      `;
-    } else {
-      return css`
-        left: 0;
-        top: 100%;
-      `;
-    }
+  ${({ triggerSize, isItem, size, itemIndex = 0, nbItems}) => {
+    return getPositionDropdownMenu({ isItem, triggerSize, size, nbItems, itemIndex })
   }};
 `;
 
 const DropdownMenuItemStyled = styled.li`
   display: flex;
   align-items: center;
+  justify-content: space-between;
   padding: ${defaultTheme.padding.base};
   white-space: nowrap;
   cursor: pointer;
@@ -101,7 +99,43 @@ const Caret = styled.span`
 
 const TriggerStyled = ButtonStyled.withComponent("div");
 
+const DropdownTriggerContainer = forwardRef<DropdownTriggerContainerProps, Element>(({isItem, dataIndex, open, size , variant, title, onBlur, onFocus, onClick, onMouseEnter, onMouseLeave, children, ...rest}, ref) => {
+  return isItem ? (
+    <DropdownMenuItemStyled
+      onMouseEnter={onMouseEnter}
+      onMouseLeave={onMouseLeave}
+      data-index={dataIndex}
+      onClick={onClick}
+      ref={ref}
+    >
+      {children}
+    </DropdownMenuItemStyled>
+  ) : (
+    <DropdownStyled
+      active={open}
+      variant={variant}
+      className="sc-dropdown"
+      {...rest}
+    >
+    <TriggerStyled
+      variant={variant}
+      size={size}
+      className="trigger"
+      onBlur={onBlur}
+      onFocus={onFocus}
+      onClick={onClick}
+      tabIndex="0"
+      title={title}
+      ref={ref}
+    >
+      {children}
+    </TriggerStyled>
+    </DropdownStyled>
+  )
+})
+
 function Dropdown({
+  isItem = false,
   items,
   text,
   icon,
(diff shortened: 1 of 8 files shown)


src/lib/components/dropdown/Dropdown.component.js
```diff
@@ -8,23 +8,45 @@ import {
...
-import { getThemePropSelector } from "../../utils";
+import { getThemePropSelector } from '../../utils';
+import { getPositionDropdownMenu } from './utils';
 
 export type Item = {
   label: string,
   name?: string,
   selected?: boolean,
-  onClick: any => void
+  onClick?: any => void,
```
JBWatenbergScality: You might want to type this to `SyntheticMouseEvent<HTMLDivElement>` (ref : https://flow.org/en/docs/react/events/)

src/lib/components/dropdown/Dropdown.component.js
```diff
@@ -8,23 +8,45 @@ import {
...
+  dataIndex: number,
+  open?: boolean,
+  size?: string,
+  variant?: string,
+  title?: string,
+  onBlur: any => void,
+  onFocus: any => void,
+  onClick: any => void,
+  onMouseEnter: any => void,
+  onMouseLeave: any => void,
```
JBWatenbergScality: You might want to type this to SyntheticMouseEvent<HTMLDivElement> (ref : https://flow.org/en/docs/react/events/)

src/lib/components/dropdown/Dropdown.component.js
```diff
@@ -8,23 +8,45 @@ import {
...
   text?: string,
   size?: string,
   variant?: string,
   title?: string,
   items: Items,
   icon?: Node,
-  caret?: boolean
+  caret?: boolean,
+  dataIndex?: number,
+  onClick?: any => void,
```
JBWatenbergScality: You might want to type this to SyntheticMouseEvent<HTMLDivElement> (ref : https://flow.org/en/docs/react/events/)

src/lib/components/dropdown/utils.js
```diff
@@ -0,0 +1,72 @@
...
+    else {
+      return css`
+          left: ${triggerSize.width}px;
+          top: ${triggerSize.height * itemIndex}px;
+        `;
+    }
+  }
+  else {
+    // Check collision for dropdown acting as a root button*
+  if (isBottomHit) {
```
JBWatenbergScality: nit: Please format this using prettier

src/lib/components/dropdown/utils.js
```diff
@@ -0,0 +1,72 @@
...
+
+  return {
+    isTopHit,
+    isRightHit,
+    isBottomHit,
+    isLeftHit
+  }
+}
+
+export const getPositionDropdownMenu = ({ isItem, triggerSize, size, nbItems, itemIndex }) => {
```
JBWatenbergScality: This method should be unit tested

src/lib/components/dropdown/utils.js
```diff
@@ -0,0 +1,72 @@
...
+  const { isTopHit, isRightHit, isBottomHit, isLeftHit } = getBorderCollisionDetection(isItem, triggerSize, size)
+
+  if (isItem) {
+    // Check collision for dropdown acting as an item
+    if (isRightHit && isBottomHit) {
+      return css`
+          right: ${size.width}px;
+          bottom: ${-triggerSize.height * (itemIndex - nbItems + 1)}px;
```
JBWatenbergScality: What if some items takes more than one line ?

src/lib/components/dropdown/Dropdown.component.js
```diff
@@ -101,86 +99,125 @@ const Caret = styled.span`
...
+      ref={ref}
+    >
+      {children}
+    </TriggerStyled>
+    </DropdownStyled>
+  )
+})
+
 function Dropdown({
+  isItem = false,
```
JBWatenbergScality: I think separating exposed Dropdown component and an internal non-exposed DropdownAsAnItem component would be a better design. It would for example avoid confusion of having an `isItem` or `dataIndex`(which are implementation details) props available when trying to use the Dropdown component. I think it is very important to keep our exposed API very clear and not add props linked to the implementation in what we expose.

JBWatenbergScality (changes requested):
Few comments below that we may want to address before merging this. 
Another important fact I noticed is that the dropdown is not navigable through keyboard interactions (tab or directional arrows). It is not specific to this feature but this dropdown in dropdown implementation makes this an even farer goal to achieve, maybe it is something we should keep in mind as well when working on such features.

src/lib/components/dropdown/Dropdown.component.js
```diff
@@ -41,43 +63,19 @@ const DropdownMenuStyled = styled.ul`
   position: absolute;
   margin: 0;
   padding: 0;
-  border: 1px solid ${getThemePropSelector("primary")};
```
ChengYanJin, line 44: Please double-check with the design, we should have a border for each of the items.

src/lib/components/dropdown/Dropdown.component.js
```diff
@@ -101,86 +99,125 @@ const Caret = styled.span`
...
-      variant={variant}
-      className="sc-dropdown"
-      {...rest}
-    >
-      <TriggerStyled
+      <DropdownTriggerContainer
+        open={open}
+        isItem={isItem}
+        onMouseEnter={(e) => {
+          setItemIndex(e && e.target && e.target.getAttribute('data-index') || 0)
```
ChengYanJin: We got a flow error. https://eslint.org/docs/rules/no-mixed-operators
```setItemIndex(e && e.target && e.target.getAttribute('data-index')) || 0```

File: src/lib/components/dropdown/Dropdown.component.js [hunk 10]
```diff
@@ -101,86 +99,125 @@ const Caret = styled.span`
 
 const TriggerStyled = ButtonStyled.withComponent("div");
 
+const DropdownTriggerContainer = forwardRef<DropdownTriggerContainerProps, Element>(({isItem, dataIndex, open, size , variant, title, onBlur, onFocus, onClick, onMouseEnter, onMouseLeave, children, ...rest}, ref) => {
```
ChengYanJin: I am wondering do you have the prettier extension installed? Please enable it to make sure the code follows the `.prettierrc` config.
Thanks!

src/lib/components/dropdown/Dropdown.component.js
```diff
@@ -8,23 +8,45 @@ import {
...
+import { getPositionDropdownMenu } from './utils';
 
 export type Item = {
   label: string,
   name?: string,
   selected?: boolean,
-  onClick: any => void
+  onClick?: any => void,
+  submenuIcon?: Node,
+  submenuItems?: Array<Item>
```
ChengYanJin: As shown in the design, the documentation and support item also have another external link icon.
Maybe we miss another `iconExternal` prop?

src/lib/components/dropdown/Dropdown.component.js
```diff
@@ -101,86 +99,125 @@ const Caret = styled.span`
...
           <ButtonIcon text={text} size={size}>
             {icon}
           </ButtonIcon>
         )}
         {text && <ButtonText className="sc-trigger-text">{text}</ButtonText>}
-        {caret && (
+        {caret && items.length > 0 && (
           <Caret>
-            <i className="fas fa-caret-down" />
+            <i className={`fas fa-caret-${isItem ? 'right' : 'down'}`} />
```
ChengYanJin: The right arrow icon should be `chevron` instead of `caret`
`<i className="fas fa-chevron-right" />`

ChengYanJin (changes requested):

src/lib/components/dropdown/Dropdown.component.js
romainlancia: done

src/lib/components/dropdown/Dropdown.component.js
romainlancia: done

src/lib/components/dropdown/Dropdown.component.js
romainlancia: done

src/lib/components/dropdown/Dropdown.component.js
romainlancia: done

src/lib/components/dropdown/Dropdown.component.js
romainlancia: done

src/lib/components/dropdown/Dropdown.component.js
romainlancia: done

src/lib/components/dropdown/utils.js
romainlancia: done

src/lib/components/dropdown/utils.js
romainlancia: done

src/lib/components/dropdown/Dropdown.component.js
romainlancia: done

src/lib/components/dropdown/Dropdown.component.js
romainlancia: done

src/lib/components/dropdown/Dropdown.component.js
romainlancia: done

src/lib/components/dropdown/utils.js
romainlancia: @JBWatenbergScality Trying to find a solution but this one is a bit tricky to handle because indeed if an item have a height superior to other item, the new menu generated will display with a wrong offset. The problem I have to get each item size of the previous menu to place the new menu correctly but it will add a lot of complexity to the code. I can figure out a way to implement that in a simple way, do you have any suggestions ?

romainlancia:
@JBWatenbergScality @ChengYanJin Regarding prettier, I think it is best to format the code before someone pushes his work. I just launch ```npx prettier --check ./src``` and a lot of files are not following the prettier config.
For now, I just run prettier on my files in order to not push files that are not related to the scope of the ticket.
But, in a future ticket, we can maybe do something like this:
```
  "husky": {
    "hooks": {
      "pre-commit": "npx prettier --write ./src && npm run lint && npm run flow && git-diff --exit-code"
    }
  },
```
 What do you think ?

romainlancia:
here's a video of a demo: https://user-images.githubusercontent.com/47044686/115277099-c37fb100-a13b-11eb-80bc-bb2687e69221.mp4

PR #291, opened 2021-04-07
feat: icon wrapper

### LINKED ISSUE #283: Implement FontAwesome Icon Wrapper
**Component**: Icon

**Why this is needed**:
The purpose of is icon wrapper is to help us choose the good icon base on the usage instead of choosing a random one directly from FontAwesome. 

**What should be done**:
Create a new component `Icon`.
Following the [Iconography](https://github.com/scality/core-ui/pull/276), we want to implement a wrapper for all the icon lists in the PR.

Example:
image
when I call`<Icon name={"Theme"}/>` I will get directly the theme icon. 
Instead of `<i className="fas fa-palette" />`

Would be great if the implementation of <Icon name={"Theme"}/> lazy loads only the required icon

**Test plan**:
**Next step**: Will need to replace all the current icons directly from `fontawsome` by the iconwrapper from `core-ui`.

theocerutti:
**Component**: Icon

**Description**:

Icon Wrapper (props: color, size, name)

**Breaking Changes**:

- [NO] Breaking Changes: No breaking changes, but in the future it will be necessary to remove @fortawesome/fontawesome-free to replace it with the IconWrapper

Closes: #283


src/lib/components/icon/Icon.component.js
```diff
@@ -0,0 +1,136 @@
...
+    if (props.color && theme[props.color]) {
+      return css`
+          color: ${theme[props.color]};
+      `;
+    }
+  }}
+`;
+
+type Props = {
+  name: string,
```
JBWatenbergScality: This should be typed to `"Account" | "Backend" | ...`. Would be great if the type is derived from `iconTable`

src/lib/components/icon/Icon.component.js
```diff
@@ -0,0 +1,136 @@
...
+  size?: string,
+  color?: string,
```
JBWatenbergScality: This should be typed using `keyof theme` type

src/lib/components/icon/Icon.component.js
```diff
@@ -0,0 +1,136 @@
...
+  return React.lazy(async () => {
+    try {
+      let icon = null;
+      if (iconType === "fas") {
+        icon = await import(`@fortawesome/free-solid-svg-icons/${iconClass}.js`);
+      } else if (iconType === "far") {
+        icon = await import(`@fortawesome/free-regular-svg-icons/${iconClass}.js`);
+      } else {
+        icon = await import(`@fortawesome/free-solid-svg-icons/${iconClass}.js`);
+      }
```
JBWatenbergScality: This could be simplified as follow
```suggestion
     const fontAwesomeType = iconType === "far" ? "free-regular-svg-icons" : "free-solid-svg-icons";
     const icon = await import(`@fortawesome/${fontAwseomeType}/${iconClass}.js`);
```

src/lib/components/icon/Icon.component.js
```diff
@@ -0,0 +1,136 @@
...
+      return css`
+          color: ${theme[props.color]};
+      `;
+    }
+  }}
+`;
+
+type Props = {
+  name: string,
+  size?: string,
```
JBWatenbergScality: This should be typed using FontAwesomeIcon type definition

src/lib/components/icon/Icon.component.js
```diff
@@ -0,0 +1,136 @@
...
+      }
+      return { default: ({ color, size, ...rest }) => <IconStyled color={ color } icon={ icon[iconClass] } size={ size } { ...rest } /> }
+    } catch {
+      return null;
+    }
+  })
+}
+
+function Icon({
+  name = "Theme",
```
JBWatenbergScality: I'm not sure we want a default icon, the name prop should be required and we should log / throw when no name is provided. Picking a default value could render an unexpected icon and then confused the user

src/lib/components/icon/Icon.component.js
```diff
@@ -0,0 +1,136 @@
...
+  "Account": "fas faWallet",
+  "Backend": "fas faNetworkWired",
+  "Tape": "fas faTape",
+  "Node-backend": "fas faServer",
+  "Volume-backend": "fas faHdd",
+  "Node-pdf": "fas faDatabase",
+  "Volume-pdf": "fas faCompactDisc",
+  "Network": "fas faProjectDiagram",
+  "Bucket": "fas faGlassWhiskey",
+  "Cloud-backend": "far faCloud",
```
JBWatenbergScality: This icon seems not loading

theocerutti: The Cloud icon required a PRO font awesome
We want to use: https://fontawesome.com/icons/cloud?style=regular

Maybe I can use this: https://fontawesome.com/icons/cloud?style=solid (free icon) ?

Cuervino: Yes, fully correct. 
Let's use the free version now, we might have our own SVG in a mid-term objective.

src/lib/components/icon/Icon.component.js
theocerutti: Done

src/lib/components/icon/Icon.component.js
theocerutti: Done

src/lib/components/icon/Icon.component.js
theocerutti: Done

src/lib/components/icon/Icon.component.js
theocerutti: Done

src/lib/components/icon/Icon.component.js
theocerutti: Done

src/lib/components/icon/Icon.component.js
theocerutti: Done

JBWatenbergScality (commented):
looks good to me LGTM, can you solve the conflicts ?

JBWatenbergScality:
The build is failing when executing flow checks. Can you check this please ?

```
> flow check
Error ------------------------------------------------------------------- src/lib/components/icon/Icon.component.js:7:15
Importing a type from an untyped module makes it `any` and is not safe! Did you mean to add `// @flow` to the top of
`@fortawesome/react-fontawesome`? (`untyped-type-import`)
   7| import type { SizeProp } from '@fortawesome/react-fontawesome';
                    ^^^^^^^^
Error ----------------------------------------------------------------- src/lib/components/icon/Icon.component.js:102:21
Cannot call `React.lazy` with async function bound to `component` because null [1] is incompatible with object type [2]
in type argument `R` [3] of the return value.
   src/lib/components/icon/Icon.component.js:102:21
                            v------------
   102|   return React.lazy(async () => {
   103|     try {
   104|       const fontAwesomeType = iconType === "far" ? "free-regular-svg-icons" : "free-solid-svg-icons";
   105|       const icon = await import(`@fortawesome/${fontAwesomeType}/${iconClass}.js`);
   106|       return { default: ({ color, size, ...rest }) => <IconStyled color={ color } icon={ icon[iconClass] } size={ size } { ...rest } /> }
   107|     } catch {
   108|       return null;
   109|     }
   110|   })
          ^
References:
   src/lib/components/icon/Icon.component.js:108:14
   108|       return null;
                     ^^^^ [1]
   /tmp/flow/flowlib_2c01daa1/react.js:313:30
   313|     component: () => Promise<{ default: React$AbstractComponent<Config, Instance>, ... }>,
                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ [2]
   /tmp/flow/flowlib_2c01daa1/core.js:659:24
   659| declare class Promise<+R> {
                               ^ [3]
Error ----------------------------------------------------------------- src/lib/components/icon/Icon.component.js:105:26
The parameter passed to `import` must be a string literal.
   105|       const icon = await import(`@fortawesome/${fontAwesomeType}/${iconClass}.js`);
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

theocerutti:
> Importing a type from an untyped module makes it `any` and is not safe! Did you mean to add `// @flow` to the top of
`@fortawesome/react-fontawesome`? (`untyped-type-import`)
   7| import type { SizeProp } from '@fortawesome/react-fontawesome';
   
I have to generate libdef with flow-typed install.
The problem is its generate a LOTS of issues in others components.
I can keep only libdef of @fortawesome/react-fontawesome to resolve this or should I put a //$FlowFixMe ?

> Error ----------------------------------------------------------------- src/lib/components/icon/Icon.component.js:105:26
The parameter passed to `import` must be a string literal.
   105|       const icon = await import(`@fortawesome/${fontAwesomeType}/${iconClass}.js`);
   
Should I update flowconfig.json (and add module.ignore_non_literal_requires=true) to allow this import or should I put a //$FlowFixMe ?

JBWatenbergScality:
> I can keep only libdef of @fortawesome/react-fontawesome to resolve this or should I put a //$FlowFixMe ?

I would go for keeping only libdef of @fortawesome/react-fontawesome else we won't benefit from the typing at all.

> Should I update flowconfig.json (and add module.ignore_non_literal_requires=true) to allow this import or should I put a //$FlowFixMe ?

I'm ok for adding `module.ignore_non_literal_requires=true` in flowconfig.

ChengYanJin (approved):

theocerutti:
@bert-e approve

PR #292, opened 2021-04-07
feat: card component

### LINKED ISSUE #164: Card component
**Component**: card

**Why this is needed**:
We would like to reuse the card component
( TBD ) 
**What should be done**:
should contain the following props:
- width
- height
- header
- children
- headerBackgroundColor
- childrenBackgroundColor
- borderColor
- onClick 

image

**Test plan**:
TBD

theocerutti:
**Component**: Card

**Description**:

**Design**:

image

**Breaking Changes**:

- [NO] Breaking Changes

Closes: #164


src/lib/components/card/Card.component.js
```diff
@@ -0,0 +1,105 @@
...
+      return css`
+        border: 2px solid ${borderColor};
+        border-radius: 8px;
+      `;
+    }}
+  }
+  
+  &:hover > div {
+    background-image: linear-gradient(rgba(${props => hex2RGB(getTheme(props)[props.borderColor]).join(',')}, 0.05) 0 0);
+  }
```
JBWatenbergScality: I think it is missing an active and focus state as well as keyboard navigation. You can get inspired from this gist https://gist.github.com/JBWatenbergScality/b5975905defdcab9c4300da2f7e19cc5

ChengYanJin (commented):
You may want to address the conflicts, thanks

JBWatenbergScality:
Thanks for the changes. We are waiting for https://github.com/scality/core-ui/pull/295 to land before merging this with the new theme.

JBWatenbergScality:
Hi @theocerutti ,
We changed a bit the design of the card. Now the hover/focused border should be an outline and use the `highlight`  theme color. Could you please sync the card implementation with those changes ?

Thanks,
Jean-Baptiste

theocerutti:
@JBWatenbergScality 

> Hi @theocerutti ,
> We changed a bit the design of the card. Now the hover/focused border should be an outline and use the `highlight` theme color. Could you please sync the card implementation with those changes ?
> 
> Thanks,
> Jean-Baptiste

Done (i updated pictures in the PR description).

I also modified the default color value to match darkRebrand theme:
```js
  headerBackgroundColor = 'backgroundLevel1',
  bodyBackgroundColor = 'backgroundLevel2',
  borderColor = 'highlight', // hover and border
```
is it okay ?

src/lib/components/card/Card.component.js
```diff
@@ -0,0 +1,134 @@
...
+      cursor: pointer;
+  
+      &:hover {
+        box-shadow: 0 0 0 2px ${getTheme(props).highlight};
+      }
+  
+      &:focus {
+        outline: 0;
+        box-shadow: 0 0 0 1px ${getTheme(props)[props.status || 'highlight']};
+      }
```
JBWatenbergScality: @Cuervino what do you think of this `focus` state for the cards ?

src/lib/components/card/Card.component.js
```diff
@@ -0,0 +1,134 @@
...
+`;
+
+type CardProps = {
+  width?: string,
+  height?: string,
+  header?: Node,
+  body?: Node,
+  headerBackgroundColor?: string,
+  bodyBackgroundColor?: string,
+  status?: 'healthy' | 'warning' | 'critical',
```
JBWatenbergScality, line 123: An healthy card should use `backgroundLevel4` for the header background (current default if no status is provided)

Cuervino: Correct, the logic is that we'll not focus too much on the Green Healthy part, that could distract the user, but we'll highlight any non-healthy state with Warning and Critical colors.

src/lib/components/card/Card.component.js
Cuervino: IMO, the focus state like that is too close to the hover state.
There should be a "additive" logic.
I think that an outline 2 px out of the card with the color buttonSecondary should be good.
navigation-status=Focus, Health=healthy

JBWatenbergScality (approved):

theocerutti:
@bert-e approve

PR #307, opened 2021-04-20
bugfix: unclickable checkbox

theocerutti:
**Component**: Checkbox

**Description**:

There is a very specific bug when several checkboxes are in a react-table and the table row allow to toggle the checkbox when clicked. They can't be clickable anymore. It seems that changing the icon of the checkbox to the top would fix the bug.

**Breaking Changes**:

- [NO] Breaking Changes


ChengYanJin (approved):

theocerutti:
@bert-e approve

PR #312, opened 2021-04-25
Remove the `aria-label` in Error page

ChengYanJin:
**Component**: Error Pages

**Description**:
When label text is visible on screen, we should use `aria-labelledby` instead of `aria-label`.
And the value of `aria-labelledby` should not link to the content

**Design**:

**Breaking Changes**:

- [] Breaking Changes
no

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/src/lib/components/error-pages/ErrorPage404.component.js
+++ b/src/lib/components/error-pages/ErrorPage404.component.js
@@ -77,10 +77,10 @@ function ErrorPage404({ btnLink = '/', locale = 'en', ...rest }: Props) {
       </Row>
       <Row>
         <Description>
-          <DescriptionContent aria-label="page not found">
+          <DescriptionContent>
             {translations[locale].error_desc}
           </DescriptionContent>
-          <DescriptionContent aria-label="check url or contact administrator">
+          <DescriptionContent>
             {translations[locale].should_do}
           </DescriptionContent>
         </Description>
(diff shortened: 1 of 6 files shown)


src/lib/components/error-pages/ErrorPage404.component.js
```diff
@@ -77,10 +77,10 @@ function ErrorPage404({ btnLink = '/', locale = 'en', ...rest }: Props) {
       </Row>
       <Row>
         <Description>
-          <DescriptionContent aria-label="page not found">
+          <DescriptionContent aria-labelledby="page-not-found">
```
JBWatenbergScality: I'm not sure this is correct/ nor needed. 
According to MDN https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute  aria-labelledby attribute accepts a list of id of other dom element. So here the id tag is missing. Also as you mentioned as the text is readable and explicit there is no need of any extra aria property. Screen readers would just read the content of the element naturally

ChengYanJin: You are right!
I'll remove the `aria-label` in the error pages

JBWatenbergScality (approved):

JBWatenbergScality:
the branchname shall have been bugfix/fix-the-aria-label to trigger the bot

PR #319, opened 2021-05-03
feat: restyle tabs component

### LINKED ISSUE #316: Restyle Tabs component
**Component**:
Tabs

**Why this is needed**:
- new theme for core-ui, colors need to be aligned
- consistency

**What should be done**:
1. Update the Tab's style
2. Check the status of tabs (default, hover, selected)
3. Change the separator between Tabs and the color of the tabs' line
4. Adjust core-ui story (in Storybook) accordingly

Note that the background colors could be overridden in the UIs (not always backgroundLevel3, or 4)

Few notes and examples:

The tabs themselves
image

- There is a separator element between every tab (1px wide, 16 px high, or rem equivalent).
- there are 2px space between the top of the tab and the top of the tab-line
- The full "line" of the tab has a bg color (so it's the same color for non-active tabs and the tab-line color)
tabs_proto
image

For deeper inspection, check the Figma file:
https://www.figma.com/file/qrsV3Oh7XZHDhZfdk5jZWI/core-ui?node-id=32%3A360

theocerutti:
**Component**: Tabs

**Description**: Restyle Tabs component

**Design**:

image

**Breaking Changes**:

- [] Breaking Changes: Modify tabs default color

Closes: #316

### DIFF
--- a/src/lib/components/tabs/Tabs.component.js
+++ b/src/lib/components/tabs/Tabs.component.js
@@ -1,12 +1,12 @@
 //@flow
 import React from "react";
-import styled, { css } from "styled-components";
+import styled from "styled-components";
 import type { Node } from "react";
 import * as defaultTheme from "../../style/theme";
 import { getTheme } from "../../utils";
 
 type Item = {
-  title: string,
+  title: Node | string,
   selected?: boolean,
   onClick: (any) => void,
 };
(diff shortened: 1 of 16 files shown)


Cuervino (commented):
1. It might be a little issue with the separator, in 2 places it looks like it's larger. 
image

2. The font for the non-active tab should be textTertiary
image

3. Would it be possible to have a clickable example in the story? Currently, there is only a hover.

Cuervino (approved):
LGTM !

src/lib/components/tabs/Tabs.component.js
```diff
@@ -1,52 +1,87 @@
...
 const TabsContainer = styled.div`
+  background-color: ${props => props.tabLineColor || getTheme(props).backgroundLevel3};
   margin: ${defaultTheme.padding.base} 0;
+  padding-top: 2px;
   * {
     box-sizing: border-box;
   }
-`;
+
+  & .sc-tabs-item {
```
JBWatenbergScality: nit: Prefer use ${TabItem} directly in order to avoid classnames conflict (event through sc prefix here should be enough)

JBWatenbergScality (approved):

src/lib/components/tabs/Tabs.component.js
theocerutti: done

theocerutti:
@bert-e approve

PR #321, opened 2021-05-04
Improvement: rework toggle

### LINKED ISSUE #273: Toggle new design
**Component**:

Toggle

**Why this is needed**:
- The current style could be improved
- It's not always recognized as being a toggle. 

**What should be done**:
1. Update the design for the OFF/ON
2. Having a smooth animation
3. Create a variant with `text`

Switch

Inspect this Figma file.
https://www.figma.com/file/qrsV3Oh7XZHDhZfdk5jZWI/core-ui?node-id=65%3A44

Prototype:
https://www.figma.com/proto/qrsV3Oh7XZHDhZfdk5jZWI/core-ui?page-id=65%3A0&node-id=65%3A61&viewport=-329%2C408%2C4.856637001037598&scaling=contain

**Test plan**:
- The full toggle box is hoverable - clickable (and not only the visible part of the toggle)
- The cursor should change to action mode when hovered

theocerutti:
**Component**: Toggle

**Description**: Rework Toggle styling

**Design**:
image

image

**Breaking Changes**:

- [] Breaking Changes: I modified the wrapper ".sc-toggle" from div to label. 

Closes: #273

### DIFF
--- a/src/lib/components/toggle/Toggle.component.js
+++ b/src/lib/components/toggle/Toggle.component.js
@@ -9,85 +9,78 @@ type Props = {
   onChange: (e: SyntheticEvent<HTMLInputElement>) => void,
   label?: string,
 };
-
-const width = 20;
-const height = 10;
-
-const Switch = styled.label`
-  cursor: pointer;
-  & input {
-    width: 0;
-    height: 0;
-    margin: 0;
-    position: absolute;
-  }
-
-  & span {
-    display: block;
-    position: relative;
-    width: ${width}px;
-    height: ${height}px;
-    border-radius: ${width}px;
-    padding: 2px;
-    background-color: ${(props) =>
-  getTheme(props)[props.toggle ? 'selectedActive' : 'infoPrimary']};
-    transition: 0.4s;
-  }
-
-  & span:before {
-    content: '';
-    position: absolute;
-    inset: 0;
-    border: 1px solid transparent;
-    background-clip: padding-box;
-    background-color: ${(props) =>
-  getTheme(props)[props.toggle ? 'selectedActive' : 'backgroundLevel1']};
-    border-radius: ${width}px;
-    transition: 0.4s;
-  }
-
-  & span:after {
-    content: '';
-    display: block;
-    width: ${height}px;
-    height: 100%;
-    background-color: ${(props) =>
-  getTheme(props)[props.toggle ? 'textSecondary' : 'textTertiary']};
-    border-radius: 50%;
-    position: relative;
-    right: 0;
-    transition: 0.4s;
-  }
-
-  & input:checked + span:after {
-    right: calc(-100% + ${height}px);
-  }
-`;
-
 const ToggleContainer = styled.label`
   display: inline-flex;
   align-items: center;
   position: relative;
+  cursor: pointer;
+`;
+
+const Switch = styled.label`
+  cursor: pointer;
+  position: relative;
+  width: 24px;
+  height: 14px;
+`;
+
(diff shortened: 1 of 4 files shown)


src/lib/components/toggle/Toggle.component.js
```diff
@@ -18,78 +18,82 @@ const ToggleContainer = styled.div`
...
-        top: -4px;
-        background-color: ${defaultTheme.grayLight};
-        transition: 0.5s;
-        border-radius: 50%;
-      }
+  &:before {
+    border-radius: 50%;
+    position: absolute;
+    content: '';
+    height: 16.5px;
```
Cuervino: It would be better to have a fully round circle (same height and width)

src/lib/components/toggle/Toggle.component.js
```diff
@@ -18,78 +18,82 @@ const ToggleContainer = styled.div`
 const Switch = styled.label`
   position: relative;
   display: inline-block;
-  width: 24px;
+  width: 42px;
```
Cuervino: The target size of the toggle itself is 24px*14px (width * height).

With 42px width, the label is way too small compared to the toggle.

Cuervino (changes requested):
1. The toggle itself is twice too big;
2. the circle is not round (issue on Figma also)
3. The action should be on the toggle (or also on the label), but not on the full line/container
image

The rest is nice!

Cuervino (changes requested):
Seems almost good!
Maybe it's my browser, but it looks like it's not centered. Could you please check?
image

theocerutti:
@Cuervino 
The problem is that if I center it totally at this zoom level the circle becomes non-centered when zoomed normally. It seems that the browser has trouble when the sizes are very small.

Currently, if we zoom "normally" the circle is centered:
image
image

src/lib/components/toggle/Toggle.component.js
```diff
@@ -19,77 +19,81 @@ const Switch = styled.label`
...
-      }
+const Slider = styled.span`
+  position: absolute;
+  cursor: pointer;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  background-color: ${(props) => getTheme(props).backgroundLevel1};
+  -webkit-transition: 0.4s;
```
JBWatenbergScality: Styled component automatically add vendor prefixing to css rules (https://styled-components.com/docs/basics#motivation) so this line can be removed

src/lib/components/toggle/Toggle.component.js
```diff
@@ -19,77 +19,81 @@ const Switch = styled.label`
...
+    border-radius: 50%;
+    position: absolute;
+    content: '';
+    height: 10.5px;
+    width: 10.5px;
+    left: 1px;
+    top: 0.5px;
+    background-color: ${(props) =>
+      getTheme(props)[props.toggle ? 'textSecondary' : 'textTertiary']};
+    -webkit-transition: 0.4s;
```
JBWatenbergScality: This line can be removed

Cuervino:
It looks like the previous version, even if not aligned, was better.
1. On Firefox, there is a strange white corner on the top left
image

2. On Chrome, the rounding seems rather strange. On 100% zoom, one circle in the toggle looks like top aligned, the other looks like bottom aligned.
image

Cuervino (approved):
Top, LGTM!

JBWatenbergScality (approved):

theocerutti:
@bert-e approve

PR #328, opened 2021-05-20
Improvement/text component

MonPote:
**Component**:

**Description**:

**Design**:

**Breaking Changes**:

- [] Breaking Changes

---

Closes: #ISSUE_NUMBER


src/lib/index.css
```diff
@@ -1,5 +1,6 @@
 html, body {
   font-family: "Lato";
+  font-size: max(14px, calc(0.2em + 0.8vw));
```
JBWatenbergScality: to be 14 px on a 1440 px we might need to change this to `max(14px, 0.972vw)`

MonPote: Ok let's do that.

JBWatenbergScality (approved):

PR #329, opened 2021-05-20
Div component

MonPote:
**Component**:

**Description**:

**Design**:

**Breaking Changes**:

- [] Breaking Changes

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -1,5 +1,5 @@
 ---
-version: "0.2"
+version: '0.2'
 
 branches:
   user/*, feature/*, improvement/*, bugfix/*, w/*, q/*, hotfix/*, dependabot/*:
@@ -17,7 +17,7 @@ stages:
     steps:
       - Git: &git_pull
           name: git pull
-          repourl: "%(prop:git_reference)s"
+          repourl: '%(prop:git_reference)s'
           method: clobber
           retryFetch: true
           haltOnFailure: true
@@ -26,7 +26,6 @@ stages:
           workdir: build/core-ui
           command: >
             npm ci &&
-            npm run test -- -w 1 &&
             npm run flow &&
             npm run lint
           haltOnFailure: false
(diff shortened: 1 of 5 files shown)


src/lib/components/div/Div.js
```diff
@@ -0,0 +1,112 @@
...
+  return result;
+};
+
+const m = createSpacingFn('m', ['margin']);
+const mt = createSpacingFn('mt', ['margin-top']);
+const mr = createSpacingFn('mr', ['margin-right']);
+const mb = createSpacingFn('mb', ['margin-bottom']);
+const ml = createSpacingFn('ml', ['margin-left']);
+const mx = createSpacingFn('mx', ['margin-top', 'margin-bottom']);
+const my = createSpacingFn('my', ['margin-left', 'margin-right']);
```
JBWatenbergScality: I think x and y axis are reversed here. In the common spirit x should stand for the horizontal axis I believe. (Tailwind also implemented the x axis against left/right properties : https://tailwindcss.com/docs/padding#add-horizontal-padding)

JBWatenbergScality (commented):
Additionally to the raw component we might want to document/show some example of the composability of this approach when it comes to use those spacing utils with other components

src/lib/components/div/Div.js
MonPote: Indeed that was a mistake

stories/div.stories.js
```diff
@@ -0,0 +1,136 @@
...
+      </Div>
+      <Div as={BoldP} my={14} p={14}>
+        Bold {'"</p>"'} with my 14 and p 14
+      </Div>
+      <Div
+        as={DisplayTextComp}
+        text={'Div + React component + nested styled component'}
+        m={12}
+        p={14}
+      ></Div>
```
JBWatenbergScality: This is great, I'm wondering how does it affect the typing ? Do we get props autocompletion/validation for both Div and  DisplayTextComp component ?

JBWatenbergScality (approved):

PR #333, opened 2021-06-07
Button V2

ChengYanJin:
**Component**: button

**Description**:
image

Including keyboard navigation: 
image

@Cuervino 
I didn't find a way to add radius for the outline...
Do we open to `box-shadow` or we're OK with the current implementation?

**Design**:
You will need to check out the branch to play with the interaction.

**Breaking Changes**:

- [] Breaking Changes
none

In order to use the new button you will need to import it from this path
` import {Button} from "@scality/core-ui/dist/next"`

I removed all the snapshot in this PR, since it's not useful for us anymore. 

---

Closes: #ISSUE_NUMBER


src/lib/components/buttonv2/Buttonv2.component.js
```diff
@@ -0,0 +1,231 @@
...
+function Button({
+  variant,
+  disabled,
+  label,
+  icon,
+  onClick,
+  tooltip,
+  ...rest
+}: Props) {
+  if (!icon && !label)
```
JBWatenbergScality: nit: add missing curly braces here

src/lib/components/buttonv2/Buttonv2.component.js
```diff
@@ -0,0 +1,231 @@
...
+    console.warn(
+      'Please specify either icon or label prop for the button component.',
+    );
+
+  if (icon && !tooltip)
```
JBWatenbergScality: nit add missing curly braces here

src/lib/components/buttonv2/Buttonv2.component.js
```diff
@@ -0,0 +1,231 @@
...
+import Tooltip from '../tooltip/Tooltip.component.js';
+import type { Props as TooltipProps } from '../tooltip/Tooltip.component.js';
+import { getTheme } from '../../utils';
+import * as defaultTheme from '../../style/theme';
+import { spacing } from '../../style/theme.js';
+
+export type Props = {
+  variant?: 'primary' | 'secondary' | 'danger' | 'outline',
+  disabled?: boolean,
+  onClick?: (any) => void,
```
JBWatenbergScality: onClick takes a MouseEvent as an attribute (https://flow.org/en/docs/react/events/ , SyntheticMouseEvent<HTMLButtonElement>)

src/lib/components/buttonv2/Buttonv2.component.js
```diff
@@ -0,0 +1,231 @@
...
+import type { Node } from 'react';
+import styled, { css } from 'styled-components';
+import Tooltip from '../tooltip/Tooltip.component.js';
+import type { Props as TooltipProps } from '../tooltip/Tooltip.component.js';
+import { getTheme } from '../../utils';
+import * as defaultTheme from '../../style/theme';
+import { spacing } from '../../style/theme.js';
+
+export type Props = {
+  variant?: 'primary' | 'secondary' | 'danger' | 'outline',
```
JBWatenbergScality, line 12: Variant is mandatory in your implementation

JBWatenbergScality (commented):
It may be missing the loading/pending state

Cuervino (approved):
I could not find any flaw, looks very nice, thanks!

src/lib/components/buttonv2/Buttonv2.component.js
ChengYanJin: oh. Yes. I missed that!

src/lib/components/buttonv2/Buttonv2.component.js
ChengYanJin: Good catch! 
After talking with @Cuervino, we realized we still need the `ghost button`, meaning the variant should be optional. So I removed the warning in 8e4f8d35

PR #339, opened 2021-06-15
table: Add simple content table

ChengYanJin:
**Component**: table 

**Description**:
Try to render the simple table content for the pod list table in Metalk8s UI.
image

Remain issue to solve: 
The scrollbar breaks the alignment of the header and body :(
still trying to figure out the solution. 

**Design**:
https://docs.google.com/document/d/1gZgZtcJtUZQlpN7H4zJhyeDK-c2w6y2UlaPGyGdqTpk/edit
PoC branch: `improvement/poc-new-table-component`

export the component in the next.js
`import { Table } from '@scality/core-ui/dist/next';`

**Breaking Changes**:

- [] Breaking Changes

---

Closes: #ISSUE_NUMBER


src/lib/components/tablev2/SimpleContent.js
```diff
@@ -0,0 +1,134 @@
...
+  TableBody,
+  TableHeader,
+  SortCaretWrapper,
+  SortIncentive,
+} from './Tablestyle';
+import { TableContext } from './Tablev2.component';
+
+export type SimpleContentProps = {
+  rowHeight: number,
+  outerTableHeight: number, //in pixel, the sum of height outside the table, pass it to virtualized table to calculate the height of the table
```
JBWatenbergScality: Can we compute this in the component ? Maybe we can wrap the table content with a div with display flex so that it fills parent height and use element.getBoundingClientRect() to retrieve the height.
This behavior could be shared with the multi select content so it could make sense to be extracted to a component that won't be exported (implementation detail of SimpleContent / MutliSelectContent components)

src/lib/components/tablev2/SimpleContent.js
```diff
@@ -0,0 +1,134 @@
...
+  TableRow,
+  TableBody,
+  TableHeader,
+  SortCaretWrapper,
+  SortIncentive,
+} from './Tablestyle';
+import { TableContext } from './Tablev2.component';
+
+export type SimpleContentProps = {
+  rowHeight: number,
```
JBWatenbergScality: Do we still need to give the control over the row height ? If so I think providing a set of options would be a better approach in order to unify the height across the UIs (a bit like the spacing component in a sense, provide a well-known pre defined scale). 
Additionally I think this could be an optional prop and default to a single line row height.

src/lib/components/tablev2/SimpleContent.js
```diff
@@ -0,0 +1,134 @@
...
+        <TableRow
+          {...row.getRowProps({
+            /* Note:
+            We need to pass the style property to the row component.
+            Otherwise when we scroll down, the next rows are flashing because they are re-rendered in loop. */
+            style: { ...style },
+          })}
+          row={row}
+          separationLineVariant={separationLineVariant}
+          className="tr"
```
JBWatenbergScality: Is this className required ? To me it could makes sense that the TabelRow component renders to an actual tr component

src/lib/components/tablev2/SimpleContent.js
```diff
@@ -0,0 +1,134 @@
...
+                ...cell.column.cellStyle,
+                // Vertically center the text in cells.
+                display: 'flex',
+                flexDirection: 'column',
+                justifyContent: 'center',
+              },
+            });
+
+            return (
+              <div {...cellProps} className="td">
```
JBWatenbergScality: I would suggest to switch to a real td component `<td {...cellProps}>`

src/lib/components/tablev2/SimpleContent.js
```diff
@@ -0,0 +1,134 @@
...
+          })}
+        </TableRow>
+      );
+    },
+    [prepareRow, rows, separationLineVariant],
+  );
+
+  return (
+    <>
+      <div className="thead">
```
JBWatenbergScality: I would suggest to switch to a real thead component <thead>

src/lib/components/tablev2/SimpleContent.js
```diff
@@ -0,0 +1,134 @@
...
+        {headerGroups.map((headerGroup) => (
+          <HeadRow
+            {...headerGroup.getHeaderGroupProps()}
+            style={{
+              display: 'flex',
+            }}
+            separationLineVariant={separationLineVariant}
+            className="tr"
```
JBWatenbergScality: Same here HeadRow could render to a tr

src/lib/components/tablev2/SimpleContent.js
```diff
@@ -0,0 +1,134 @@
...
+  );
+
+  return (
+    <>
+      <div className="thead">
+        {headerGroups.map((headerGroup) => (
+          <HeadRow
+            {...headerGroup.getHeaderGroupProps()}
+            style={{
+              display: 'flex',
```
JBWatenbergScality: Can we move this style to styled-component definition of HeadRow ?

src/lib/components/tablev2/SimpleContent.js
```diff
@@ -0,0 +1,134 @@
...
+        {...getTableBodyProps()}
+        className="tbody"
+        outerTableHeight={outerTableHeight}
+      >
+        <AutoSizer>
+          {({ height, width }) => (
+            <List
+              height={height}
+              itemCount={rows.length} // how many items we are going to render
+              itemSize={rowHeight} // height of each row in pixel
```
JBWatenbergScality: I understand the List component from react-window takes this value in pixel but in our component we should compute a value taking in account the screen size to ease readability on monitoring screens and tablets.

src/lib/components/tablev2/TableUtil.js
```diff
@@ -0,0 +1,14 @@
+const STATUS_WARNING = 'warning';
+const STATUS_CRITICAL = 'critical';
+const STATUS_NONE = 'none';
+const STATUS_HEALTH = 'healthy';
+
+export const compareHealth = (status1, status2) => {
```
JBWatenbergScality: Please add some typing here and unit test it

JBWatenbergScality: Passing an invalid status will result in `undefined - undefined` which will give NaN to the user. Throwing an error if the provided status are invalid would enhance the developer experience

src/lib/components/tablev2/Tablestyle.js
```diff
@@ -0,0 +1,48 @@
...
+  height: ${(props) => props.headerHeight};
+  &:hover {
+    ${SortIncentive} {
+      display: block;
+    }
+  }
+`;
+
+export const HeadRow = styled.div`
+  height: 32px;
```
JBWatenbergScality: This should change depending on the screen height and could be considered as a minimum value eventually

JBWatenbergScality: We probably want to derive this value form the font size of the headers and add some margin around instead of a fixed value

src/lib/components/tablev2/Tablev2.component.js
```diff
@@ -0,0 +1,142 @@
...
+import SimpleContent from './SimpleContent';
+import { compareHealth } from './TableUtil.js';
+
+export type TableProps = {
+  columns: {
+    Header: string,
+    accessor: string,
+    sortType?:
+      | string
+      | ((rowA: any, rowB: any, columnId: string, desc: boolean) => number),
```
JBWatenbergScality: any can be replaced by generics deduced from the type of `data`

src/lib/components/tablev2/Tablev2.component.js
```diff
@@ -0,0 +1,142 @@
...
+    cellStyle?: { [key: string]: any },
+    Cell?: (cellProps: any) => any,
+  }[],
+  defaultSortingKey: string, //we don't display the default sort key in the URL, so we need to specify here
+  data: { [key: string]: any }[],
```
JBWatenbergScality: I would type it using `ROW[]` with `ROW` a generic type

src/lib/components/tablev2/Tablev2.component.js
```diff
@@ -0,0 +1,142 @@
...
+import { compareHealth } from './TableUtil.js';
+
+export type TableProps = {
+  columns: {
+    Header: string,
+    accessor: string,
+    sortType?:
+      | string
+      | ((rowA: any, rowB: any, columnId: string, desc: boolean) => number),
+    cellStyle?: { [key: string]: any },
```
JBWatenbergScality: Use React.CSSProperties type

src/lib/components/tablev2/Tablev2.component.js
```diff
@@ -0,0 +1,142 @@
...
+    Cell?: (cellProps: any) => any,
```
JBWatenbergScality: It returns a React.Node

src/lib/components/tablev2/Tablev2.component.js
```diff
@@ -0,0 +1,142 @@
...
+  }[],
+  defaultSortingKey: string, //we don't display the default sort key in the URL, so we need to specify here
+  data: { [key: string]: any }[],
+  rowIDKey: string, // define which column represent the unique id of the table
+  children: React.Node,
+  isMultiRowSelection?: boolean,
+  isCollapsible?: boolean,
```
JBWatenbergScality: I would remove those props for now as we decided to control multi row selection with a different child component and they are not yet implemented

src/lib/components/tablev2/SimpleContent.js
```diff
@@ -0,0 +1,134 @@
...
+  rowHeight: number,
+  outerTableHeight: number, //in pixel, the sum of height outside the table, pass it to virtualized table to calculate the height of the table
+  separationLineVariant:
+    | 'backgroundLevel1'
+    | 'backgroundLevel2'
+    | 'backgroundLevel3'
+    | 'backgroundLevel4',
+};
+
+export default function SimpleContent({
```
JBWatenbergScality: rename this to SingleSelectableContent to match the tech design

src/lib/components/tablev2/Tablev2.component.js
```diff
@@ -0,0 +1,142 @@
...
+  rows: rows,
+  prepareRow: prepareRow,
+  selectedRowIds: selectedRowIds,
+  selectedFlatRows: selectedFlatRows,
+  preGlobalFilteredRows: preGlobalFilteredRows,
+  setGlobalFilter: setGlobalFilter,
+  globalFilter: globalFilter,
+  setFilter: setFilter,
+};
+export const TableContext = React.createContext<TableContextValue>({});
```
JBWatenbergScality: A common practice is to initialize context with a null value in this way we can wrap their usage with a custom hook that check their value and throw an error if they are used outside of the context provider.

src/lib/components/tablev2/Tablev2.component.js
```diff
@@ -0,0 +1,142 @@
...
+  children,
+  ...rest
+}: TableProps) {
+  const sortTypes = {
+    health: (row1, row2) => {
+      return compareHealth(row2.values.health, row1.values.health);
+    },
+  };
+
+  const getRowId = (row, relativeIndex) => {
```
JBWatenbergScality: nit: unify case of Id. In the props we have `rowIDKey` and here `getRowId`

JBWatenbergScality: I would opt to get this function from the props directly in order to give more control to the developer and address larger cases

src/lib/components/tablev2/Tablev2.component.js
```diff
@@ -0,0 +1,142 @@
...
+    return row[rowIDKey];
+  };
+
+  // Use the state and functions returned from useTable to build your UI
```
JBWatenbergScality: We can remove this comment

src/lib/components/tablev2/Tablev2.component.js
```diff
@@ -0,0 +1,142 @@
...
+    },
+    useBlockLayout,
+    useFilters,
+    useGlobalFilter,
+    useSortBy,
+    useExpanded,
+    useRowSelect,
+  );
+
+  // Render the UI for your table
```
JBWatenbergScality: We can remove this comment

src/lib/components/tablev2/Tablev2.component.js
```diff
@@ -0,0 +1,142 @@
...
+        prepareRow,
+        selectedRowIds,
+        selectedFlatRows,
+        preGlobalFilteredRows,
+        setGlobalFilter,
+        globalFilter,
+        setFilter,
+      }}
+    >
+      {/* we need to use <div/> because of the virtualized table*/}
```
JBWatenbergScality: is this due to a styling issue ? Can a css reset on the table element address this issue ? To me if possible it will be better to use a table component to keep the html semantic and benefit from browser rendering optimisations

JBWatenbergScality (changes requested):

src/lib/components/tablev2/SimpleContent.js
ChengYanJin: After checking the design from @Cuervino , I see the following possibilities.
```js
export const tableRowHeight = {
  h32: '2.286rem',
  h40: '2.858rem',
  h48: '3.428rem',
  h64: '4.572rem',
};
```
but I am not sure if we want to have the default row height => to be discussed.

Cuervino: IMO, offering those 4 predefined row height is rather good.
From a visual design point of view, these multiple of 8-px give good balance between components.

src/lib/components/tablev2/SimpleContent.js
ChengYanJin: It's a great idea. But unfortunately, without explicitly define the height of the table body, the height will be 0, hence no content can be displayed.

I don't see any other way than expose a property to ask the developer to calculate the sum height of outerTableBody.

src/lib/components/tablev2/SimpleContent.js
ChengYanJin: so we maybe should define the `rowHeight` in `vh` and convert it to pixel?

src/lib/components/tablev2/Tablestyle.js
ChengYanJin: for the moment, I will just change it to `height: 2.286rem;` and vertical aligned

src/lib/components/tablev2/SingleSelectableContent.js
```diff
@@ -10,32 +10,50 @@ import {
...
+  if (['h32', 'h40', 'h48', 'h64'].indexOf(rowHeight) === -1) {
+    console.error('Invalid table row height defined.');
+  }
+  if (
+    [
+      'backgroundLevel1',
+      'backgroundLevel2',
+      'backgroundLevel3',
+      'backgroundLevel4',
+    ].indexOf(separationLineVariant) === -1
```
JBWatenbergScality: can be replaced by `includes`

src/lib/components/tablev2/SingleSelectableContent.js
```diff
@@ -10,32 +10,50 @@ import {
...
+  ) {
+    console.error('Invalid table seperation line color defined.');
```
JBWatenbergScality: I would reword the error message to `Invalid separationLineVariant props, expected backgroundLevel1, backgroundLevel2, backgroundLevel3 or backgroundLevel4 but received ${separationLineVariant}`

src/lib/components/tablev2/TableUtil.js
```diff
@@ -1,14 +1,39 @@
...
 const STATUS_HEALTH = 'healthy';
 
-export const compareHealth = (status1, status2) => {
+// some common customized sortTypes
+export function compareHealth(
+  status1: STATUS_WARNING | STATUS_CRITICAL | STATUS_NONE | STATUS_HEALTH,
+  status2: STATUS_WARNING | STATUS_CRITICAL | STATUS_NONE | STATUS_HEALTH,
+): number {
+  if (
+    [STATUS_WARNING, STATUS_CRITICAL, STATUS_NONE, STATUS_HEALTH].indexOf(
```
JBWatenbergScality: Can be replaced by `includes`

src/lib/components/tablev2/SingleSelectableContent.js
ChengYanJin: Yes. indeed. it's more developer friendly ;)

JBWatenbergScality (approved):

PR #340, opened 2021-06-16
tabs: Document the next tab design

JBWatenbergScality:
**Component**:

Tabs

**Description**:

We are reviewing core-ui component and studying API and implementations of those next components

### DIFF
--- a/stories/tabs.stories.mdx
+++ b/stories/tabs.stories.mdx
@@ -1,0 +1,46 @@
+import { Meta } from "@storybook/addon-docs/blocks";
+
+<Meta title="Next Version Implementation Guides/Components/Navigation/Tabs" />
+
+
+## Implementation guidelines for next tabs component 
+
+It should use text basic  component to display the labels 
+The latest tabs should display the separation line at the end to mark the end of the tab
+
+Expected component API : 
+
+```
+<Tabs variant={backgroundLevel3|backgroundLevel4}>
+    <Tab path={string} label={} textBadge={} >{children}</Tab>
+    <Tab path={string} label={} textBadge={} >{children}</Tab>
+    <Tab path={string} label={} textBadge={} >{children}</Tab>
+</Tabs>
+```
+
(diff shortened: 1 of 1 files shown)


stories/tabs.stories.mdx
```diff
@@ -0,0 +1,46 @@
...
+    <Tab path={string} label={} textBadge={} >{children}</Tab>
+    <Tab path={string} label={} textBadge={} >{children}</Tab>
+    <Tab path={string} label={} textBadge={} >{children}</Tab>
+</Tabs>
+```
+
+Tabs are based on react router and extends the current route architecture of the application with child routes.
+Each path is taken as a child path to the current router path.
+
+Tab size depend on the label, minimum size is 5rem.
```
Cuervino, line 24: Tab **height** is fixed, it's the tab **width** that is dependent on the label string length.

stories/tabs.stories.mdx
```diff
@@ -0,0 +1,46 @@
+import { Meta } from "@storybook/addon-docs/blocks";
+
+<Meta title="Next Version Implementation Guides/Components/Navigation/Tabs" />
+
+
+## Implementation guidelines for next tabs component 
+
+It should use text basic  component to display the labels
```
Cuervino, line 8: TextBasic
color TextPrimary for Active tab
color TextSecondary for Non-active tabs

stories/tabs.stories.mdx
```diff
@@ -0,0 +1,46 @@
+import { Meta } from "@storybook/addon-docs/blocks";
+
+<Meta title="Next Version Implementation Guides/Components/Navigation/Tabs" />
+
+
+## Implementation guidelines for next tabs component 
+
+It should use text basic  component to display the labels 
+The latest tabs should display the separation line at the end to mark the end of the tab
```
Cuervino, line 9: There should be also a separation line between each tabs

ChengYanJin:
@Cuervino can you please send us the figma design link so we can attach it to the design doc?

PR #342, opened 2021-06-18
select: SelectV2 component

### LINKED ISSUE #334: New select component
**Component**:

Select

**Why this is needed**:

We want to review the select behavior throughout our applications.

**What should be done**:

Variants :

Rounded :

image

Default :

image

When an option is selected we display all the other options with the selected one decorated with the selected state.

Options state :

image

By default show max 4.5 items and then add a scrollbar if there is more options.

Move up the options if there is not enough space to display them under the select components.

If we have more than 8 options we display a search bar. When opening the select, we focus the search bar but scroll to the selected option (decorated with selected state). If less than 8 options focus on the first option or selected one.

When typing in the search bar, highlight the first matching option, hide the options that are not matching.

API :

```
<Select>
    <Option></Option>
</Select>
```

Select props :
- placeholder: string, optional
- disabled: boolean, optional, default false
- children: array of Option (only Option are accepted as children), required
- defaultValue: string, optional
- value: string, optional
- onChange: (newValue: string) => void
- variant: 'default' | 'rounded' - used to pick the theme of the select component

Option props :
- disabled: boolean, default false
- icon: optional, icon component only accepted
- children: string - used as a label
- value: string, required

Keyboard navigation and accessiblity :

It should follow the ARIA design pattern for a select (https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html) with these extensions/overrides :
 - When there are more than 8 options and the search bar is displayed focus the search bar when opening but scroll to the selected option. Typing arrow down should focus the option available right after the selected one. 
 - Starting to search for an item will focus the first item of the list
 - Removing everything in the search will move back the focus to the selected entry except if arrows or mouse was hovering an entry. We then preserve the focus/highlight on that option.

Could be based on react select or downshift js. 

**Test plan**:

Interactions should be tested, (you might want to use user-events testing library)
All states and variants should be tested

theocerutti:
**Component**: SelectV2

**Design**:

image
image

Closes: #334


JBWatenbergScality:
Few comments on the behavior :
 - Disable options should not be focusablee
 - When searching then validating, we can't use left/right arrows to modify the search
 - When options have an icon the icon should be rendered on the left of the label and should be displayed in the component when the option is selected

Cuervino (changes requested):
1. When expandind the select, the "focus" state should appear on the option lines only when the keyboard is pressed, otherwise it's adding to much _noise_ to the component behaviour.
(The button V2 comonent is a good example of focus state appearing only when using keyboard)
image

2. the cursor should change when hovering on a disable component (such as the button v2 component).
image
in the case of the disabled component, it would be interesting to see in Storybook the case of the Disabled Select Component already filled (with "item1", instead of "Select", that would not be a frequent use case)

3. For the search logic, once the option has been selected, it should not act anymore as a filter (that need's to be erased for seeing the other option lines).
Moreover, this wrong filtering logic preventing an easy other choice is also applied when the search feature was not used, but the option was just chosen in the list.
https://user-images.githubusercontent.com/28654802/123603452-2eafc880-d7fa-11eb-884f-8dca180ad4fe.mp4

3. There are 2 different cases for icons
     - the disabled option icon, at the end of the line (far right)
     - the option with a standard icon, with the icon on the left of the label
image
So the disabled icon should be moved to the far right of the line.

Cuervino (approved):
Top! 👍

src/lib/components/selectv2/Selectv2.component.js
```diff
@@ -0,0 +1,280 @@
...
+          };
+        });
+    } else {
+      return [];
+    }
+  }, [children]);
+
+  useEffect(() => {
+    setOptions(childrenToOptions());
+  }, [childrenToOptions]);
```
JBWatenbergScality: I think this can be refactored to a simple useMemo with children as a dependency array. It will avoid an unnecessary render linked to the useEffect 

```suggestion
  const options = useMemo((): Array<SelectOptionProps> => {
    if (children) {
      return React.Children.toArray(children)
        .filter((child) => child.type === Option)
        .map((child) => {
          const {
            value,
            children,
            disabled,
            icon,
            ...rest
          }: OptionProps = child.props;
          return {
            value: value,
            label: children,
            isDisabled: disabled || false,
            icon: icon,
            optionProps: { ...rest },
          };
        });
    } else {
      return [];
    }
  }, [children]);
```

Then the state options could be dropped.

src/lib/components/selectv2/Selectv2.component.js
```diff
@@ -0,0 +1,280 @@
...
+  ...rest
+}: SelectProps) {
+  const [keyboardFocusEnabled, setKeyboardFocusEnabled] = useState(false);
+  const [searchSelection, setSearchSelection] = useState('');
+  const [searchValue, setSearchValue] = useState('');
+  const [customPlaceholder, setPlaceholder] = useState(placeholder);
+  const [options, setOptions] = useState<Array<SelectOptionProps> | null>(null);
+  const isDefaultVariant = variant === 'default';
+  const [isMenuBottom, setIsMenuBottom] = useState(true);
+  const [selectRef, setSelectRef] = useState(null);
```
JBWatenbergScality: Is there a specific reason not to use `useRef` here to hold the select reference ?

JBWatenbergScality (commented):
Nice behaviour and implementation overall 👍 .
I've seen that the select component becomes slower when it starts to handle a lot of data (> 1000). We may have some use cases for such large selects and I think we could benefit from virtualisation to handle those use cases with better performances (maybe using [react-window](https://github.com/bvaughn/react-window)).

stories/selectv2.stories.js
```diff
@@ -105,6 +107,8 @@ export const Rounded = () => (
       <Select variant="rounded" onChange={null} />
       <Title>With scrollbar (more than 4 items)</Title>
       <CustomSelect variant="rounded" opts={optionsWithoutSearchBar} />
+      <Title>1000 items</Title>
+      <CustomSelect variant="rounded" opts={generateOptions(1000)} />
```
JBWatenbergScality, line 111: image

It seems that some items are duplicated, seems to be a bug introduced when switching to virtualized list, could you please check ?

JBWatenbergScality: Another bug is that we not scroll anymore to the selected entry when changing the focus from an item to another one

theocerutti: It also break the scroll on keyboard navigation, icons are blinking on keyboard navigation.
I will check that.

JBWatenbergScality: Additionally the scrollbar is invisible when there are 1000 items, I guess that's because the scroll handle is too small. To me we should add a min-height for the scroll-handle to provide context to the user about the scroll context in large documents.
For example here is the min height of default chrome scroll handle :
image

src/lib/components/selectv2/Selectv2.component.js
```diff
@@ -114,21 +119,69 @@ const Menu = (props) => {
...
+      listRef.current.scrollTo(
+        getScrollOffset(
+          listRef.current,
+          focusedIndex,
+          children.length,
+          optionHeight / 4,
+        ),
+      );
+    }
+  }, [children.length, focusedIndex, optionHeight]);
```
JBWatenbergScality: listRef seems to be missing in the dependency array

src/lib/components/selectv2/Selectv2.component.js
```diff
@@ -114,21 +119,69 @@ const Menu = (props) => {
...
+
+  if (children.length > 4) {
```
JBWatenbergScality: I would extract "4" in a variable `ITEMS_PER_SCROLL_WINDOW` and derive `initialOffset`, `optionHeight / 4` and `height={optionHeight * 4.5}` from this variable.

JBWatenbergScality (approved):

theocerutti:
@bert-e approve

PR #345, opened 2021-06-24
Scrollable Tabs

### LINKED ISSUE #239: Scrollable Tabs
**Component**: Tabs

**Why this is needed**:
We start to address the responsiveness issue of the UI. 
When we come to the small screen, we need to scroll to see all the tabs which are not always obvious for the user and cause a double scrollbar.
Not needed for 1024 width screen but we add a new tab we may face the issue.

**What should be done**:
As suggested by @JBWatenbergScality and agreed by @Cuervino, we will implement the scrollable tabs exactly the same as below:
image

https://material-ui.com/components/tabs/#scrollable-tabs

**Test plan**:
Please test in `MetalK8s UI - NodePage` Tabs to make sure it works well.

theocerutti:
**Component**: Scrollable Tabs

**Design**:

image
image

Most of the code comes from: https://github.com/mui-org/material-ui/blob/next/packages/material-ui/src/Tabs/Tabs.js

Closes: #239


Cuervino (changes requested):
1. The cursor should be also the "action" cursor when hovering on the scrollable button.
image

2. When pressing the "Next" button, the next cycle should be displayed, starting with the tab that was truncated in the previous cycle (especially if it's a "full" next cycle.
In the example, we see a little bit of tab6, it should not display it.
image

3. It has an impact on the last cycle, as the last cycle doesn't have to fill the full width of the tab line.
image

4. When arriving on the last cycle, the Next button should not be displayed, as there is no additional cycle. (maybe it's because it was few pixels from ending point, but the feeling is not nice in this case).
image

If these behaviours (2 and 3) don't feel right to you, we may discuss about them.

Otherwise, nicely looking!!

src/lib/components/tabsv2/animate.js
```diff
@@ -0,0 +1,51 @@
...
+
+export function animate(
+  property: string,
+  element: any,
+  to: number,
+  options: { ease: (any) => any, duration: number } = {},
+): () => void {
+  const {
+    ease = easeInOutSin,
+    duration = 300, // standard
```
Cuervino: This duration is way too small, especially when there is a large shift. 
800 or 1000 had a much better effect IMO.

Cuervino (changes requested):
Everything looks rather nice!
I would just lower the animation speed, for the user to "follow" the tab's shift.

Cuervino (approved):
LGTM

src/lib/components/tabsv2/ScrollButton.jsx
```diff
@@ -0,0 +1,44 @@
...
+    border-radius: ${defaultTheme.spacing.sp2};
+    box-shadow: none;
+    border: ${defaultTheme.spacing.sp1} solid
+      ${getThemePropSelector('infoPrimary')};
+  }
+`;
+
+type Props = {
+  direction: 'left' | 'right',
+  show: boolean,
```
JBWatenbergScality: `show` props could be replaced by a rendering condition in the parent. The condition being already computed, the implementation should be {show && <ScrollButton direction={} onClick={} />}

src/lib/components/tabsv2/StyledTabs.js
```diff
@@ -0,0 +1,109 @@
+import styled from 'styled-components';
+import { spacing } from '../../style/theme';
+import { getTheme, getThemePropSelector } from '../../utils';
+
+export const TabBar = styled.div`
+  display: flex;
+  height: 2.85714286rem; // 40px
```
JBWatenbergScality: This can be added to spacing.sp40

src/lib/components/tabsv2/Tabsv2.component.js
```diff
@@ -0,0 +1,402 @@
...
+} from './StyledTabs';
+import { useHistory, useLocation } from 'react-router-dom';
+import { matchPath, Route, Switch } from 'react-router';
+import { SecondaryText, BasicText, EmphaseText } from '../text/Text.component';
+import { animate } from './animate';
+import ScrollButton from './ScrollButton';
+import Tab from './Tab';
+import type { TabProps } from './Tab';
+
+const nextItem = (list, item) => {
```
JBWatenbergScality: it should be typed

src/lib/components/tabsv2/Tabsv2.component.js
```diff
@@ -0,0 +1,402 @@
...
+  if (list === item) {
+    return list.firstChild;
+  }
+  if (item && item.nextElementSibling) {
+    return item.nextElementSibling;
+  }
+  return list.firstChild;
+};
+
+const previousItem = (list, item) => {
```
JBWatenbergScality: it should be typed

src/lib/components/tabsv2/Tabsv2.component.js
```diff
@@ -0,0 +1,402 @@
...
+  if (list === item) {
+    return list.lastChild;
+  }
+  if (item && item.previousElementSibling) {
+    return item.previousElementSibling;
+  }
+  return list.lastChild;
+};
+
+const moveFocus = (list, currentFocus, traversalFunction) => {
```
JBWatenbergScality: it should be typed

src/lib/components/tabsv2/Tabsv2.component.js
```diff
@@ -0,0 +1,402 @@
...
+    start: false,
+    end: false,
+  });
+
+  const scrollButtonEndRef = useRef<?HTMLDivElement>(null);
+  const scrollButtonStartRef = useRef<?HTMLDivElement>(null);
+  const tabsListRef = useRef<any>(null);
+  const tabsRef = useRef<any>(null);
+
+  const scroll = (scrollValue: number, { animation = true } = {}) => {
```
JBWatenbergScality: This function is never called with `animation = false` so I assume the parameter and corresponding branch can be removed as they are never used nor tested.

src/lib/components/tabsv2/Tabsv2.component.js
```diff
@@ -0,0 +1,402 @@
...
+        moveFocus(list, null, nextItem);
+        break;
+      case 'End':
+        event.preventDefault();
+        moveFocus(list, null, previousItem);
+        break;
+      default:
+        break;
+    }
+  };
```
JBWatenbergScality: In order to simplify the component maintainability I would move this logic to a custom hook : 
`const {scrollButtonEndRef, scrollButtonStartRef, tabsListRef, tabsRef} = useScrollingTabs()`

JBWatenbergScality: Also as this code is be copied from https://github.com/mui-org/material-ui/blob/next/packages/material-ui/src/Tabs/Tabs.js I think it would be good to add a comment explaining this and copying the license of material-ui before the implementation of the hook

src/lib/components/tabsv2/Tabsv2.component.js
```diff
@@ -0,0 +1,402 @@
...
+        strict: child.props.strict,
+        sensitive: child.props.sensitive,
+      });
+      if (isSelected) {
+        setSelectedTabIndex(index);
+        hasSelectedTab = true;
+      }
+    });
+    if (!hasSelectedTab) setSelectedTabIndex(null);
+  }, [location.pathname, filteredTabsChildren]);
```
JBWatenbergScality, line 69: To me this not a side effect of the component so should not be in an effect. Considering this as an effect means that on first render the wrong tab will be displayed. Whereas I think the selectedTabIndex should be computed thanks to `useMemo` as a derived value of the current location.

src/lib/components/tabsv2/Tabsv2.component.js
JBWatenbergScality: To simplify the custom hook can be mutualised with the animation functions in the same file.

src/lib/components/tabsv2/Tabsv2.component.js
JBWatenbergScality: We need to wait a complete render so that the size of tabs is computed and we can scroll to selected tab. Hence we can keep an effect here

src/lib/components/tabsv2/Tabsv2.component.js
```diff
@@ -0,0 +1,190 @@
...
+              role="tablist"
+            >
+              {tabItems}
+            </TabBar>
+          </TabsScroller>
+          {displayScroll.end && (
+            <ScrollButton
+              ref={scrollButtonEndRef}
+              direction="right"
+              show={displayScroll.end}
```
JBWatenbergScality: this can be removed

JBWatenbergScality (approved):

theocerutti:
@bert-e approve

PR #347, opened 2021-07-07
Theme: update light theme

alexis-ld:
**Component**: Theme

**Description**: This updates the light theme following the new specifications.

**Breaking Changes**: none

---

Closes: #1618


Cuervino (commented):
The UI should not use the variables from the "old" theme.
I would advise to get rid of 
Alert
base
primary
critical
background
backgroundBluer
borderLight
info

@ChengYanJin @JBWatenbergScality WDYT?

alexis-ld:
Removed the deprecated colors.
Let me know if we need those back :)

ChengYanJin:
Hello @Cuervino, 
I would suggest the following steps to progress on the light theme topic before merging this PR.
- Check all the UIs(core-ui, MetalK8s UI and other micro apps) to make sure we are not using any **deprecated colors**. Then we can safely remove them.
- Check how all the components in core ui behave with the light mode and make sure they display correctly.

PR #351, opened 2021-07-30
Fix/1633 remove deprecated colors

alexis-ld:
**Component**: All

**Description**: Remove deprecated colors from most components.


ChengYanJin (changes requested):
Please remove the snapshots test and rename the branch to `bugfix/*`. 

I suggest while we are working on removing the deprecated color in the various UIs, we use this branch and eventually bump the `core-ui` version within the same PR.

PR #357, opened 2021-08-16
TabsV2: Add query props

theocerutti:


src/lib/components/tabsv2/Tabsv2.component.js
```diff
@@ -46,27 +46,53 @@ function Tabs({
...
+      for (const key of Object.keys(query)) {
+        if (!(queryURL.has(key) && queryURL.get(key) === query[key]))
+          return false;
+      }
+      return true;
+    },
+    [queryURL],
+  );
+
+  const serialize = (query?: Query): string => {
```
JBWatenbergScality, line 71: this is equivalent to `!query ? '' : '?'+ URLSearchParams.toString();`

JBWatenbergScality: query[key] may need to be url encoded (you can try query params containing &test= as a value to test)

JBWatenbergScality (approved):

theocerutti:
@bert-e approve

PR #359, opened 2021-08-20
global health: Remove radius on segments

JBWatenbergScality:
**Component**:

global health component

**Design**:

image


ChengYanJin (approved):
LGTM!

PR #364, opened 2021-09-06
Fix the linecolor for symmetrical chart

ChengYanJin:
**Component**: line temporal chart

**Description**:
- With this new change, for the symmetrical chart, we have to specify the color for each series. 
- Fix the bug when the values are all 0, we can't see the line because of the separation line.
- Fix the warning in the console when there is no key specify for legend. 

**Design**:

**Breaking Changes**:

- [] Breaking Changes

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/src/lib/components/linetemporalchart/LineTemporalChart.component.js
+++ b/src/lib/components/linetemporalchart/LineTemporalChart.component.js
@@ -351,6 +351,7 @@ function LineTemporalChart({
         orient: 'right',
         translate: -5, // translate both the x and y coordinates by 5 pixel
         tickOffset: 5, // shift back the y translate to make sure the tick align with the 0 seperation line
+        labelBaseline: "middle",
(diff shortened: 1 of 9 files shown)


src/lib/components/linetemporalchart/LineTemporalChart.component.js
```diff
@@ -593,7 +593,7 @@ function LineTemporalChart({
         <Legends>
           {legendLabels.map((legend, index) => {
             return (
-              <>
+              <div key={index}>
```
JBWatenbergScality: nit: You don't need to introduce an extra div to add a key here. You can use `<React.Fragment key={index}>`.

JBWatenbergScality (approved):

MonPote (approved):

JBWatenbergScality (approved):

PR #378, opened 2021-10-16
icon: Add the health icons

ChengYanJin:
**Component**: Add icons related to Health Status 

**Description**:
When I try to use the icon on the UI, I discovered the icon totally redener. 
hence I leave the PR as Draft, something to investigate.

**Design**:

**Breaking Changes**:

- [] Breaking Changes

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/src/lib/components/icon/Icon.component.js
+++ b/src/lib/components/icon/Icon.component.js
@@ -98,6 +98,7 @@ type Props = {
   name: $Keys<typeof iconTable>,
   size?: SizeProp,
   color?: $Keys<typeof brand>,
+  ariaLabel?: string,
 };
 
 function getLazyStyledIcon(iconInfo) {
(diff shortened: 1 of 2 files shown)


hug33k:
@ChengYanJin The re-render issue was caused by Metalk8s UI (cf. [PR#3576](https://github.com/scality/metalk8s/pull/3576)). I tester your branch on Metalk8s with the fix and everything works fine.

src/lib/components/icon/Icon.component.js
```diff
@@ -105,7 +108,7 @@ function getLazyStyledIcon(iconInfo) {
       const icon = await import(`@fortawesome/${fontAwesomeType}/${iconClass}.js`);
       return { default: ({ color, size, ...rest }) => <IconStyled color={ color } icon={ icon[iconClass] } size={ size } { ...rest } /> }
     } catch {
-      return { default: () => <Loader size="base" /> };
+      return { default: ({ ...rest }) => <Loader size="base" { ...rest }/> };
```
JBWatenbergScality: I'm not sure we should spread icon attributes to the loader icon (all icon attributes may not be compatible with the loader properties).
If the problem you are trying to solve is about aria label, then this label should probably be '{name of the icon}{additional aria label}' because we wouldn't care in this case if the icon is already loaded or not (anyway it can't be seen if you require aria label to be read to describe the ui 😕 ).
So to sum up I would restrain property spreading to only the aria properties and build up the label as described above

hug33k: @JBWatenbergScality
On metalk8s, label value looks like this: `status {status name}`.
Since this component can be used for something else, I think we should not include a naming template here and get the value through props directly.
Does this solution seems good to you?

hug33k: @JBWatenbergScality `aria-label` value is updated and now follows the schema you suggested

JBWatenbergScality (approved):

PR #385, opened 2021-11-10
Fix responsive for NavBar

hug33k:
**Component**:

NavBar

**Description**:

This PR replaces px values by rem values, to fix responsive issues on `NavBar`.

It also removes `TabLinkItem` inside `NavBar`, in order to replace this component by `<a>` element provided through `tabs.link` property.
Note: `TabLinkItem` styling has been migrated to `NavbarTabs` in order to apply it to `<a>` elements previously mentioned.

---

See: ARTESCA-2092

### DIFF
--- a/src/lib/components/navbar/Navbar.component.js
+++ b/src/lib/components/navbar/Navbar.component.js
@@ -1,7 +1,7 @@
 //@flow
 import React from 'react';
 import styled, { css } from 'styled-components';
-import type { Node } from 'react';
+import type { Node, Element } from 'react';
 import Logo from '../../icons/branding';
 import Dropdown from '../dropdown/Dropdown.component';
 import Button from '../button/Button.component';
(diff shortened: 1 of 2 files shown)


src/lib/components/navbar/Navbar.component.js
```diff
@@ -235,16 +216,14 @@ function NavBar({
...
-                role="tab"
-                aria-selected={selected}
-                key={`navbar_tab_link_item_${index}`}
-              >
-                {link}
-              </TabLinkItem>
+              React.cloneElement(link,
+                {
+                  className: (selected ? "selected" : ""),
+                  "aria-selected": selected
```
JBWatenbergScality: I think role="tab" and key props are missing

JBWatenbergScality (changes requested):

src/lib/components/navbar/Navbar.component.js
hug33k: Props `role` & `key` added

JBWatenbergScality (approved):

PR #390, opened 2021-11-30
refactor: add an option to hide ok health status

gaudiauj:
**Component**:

healthSelector

**Description**:
in order to avoid having an healthy option on alert. We need to add an option to hide the ok value.
**Design**:

**Breaking Changes**:

- [] Breaking Changes

---

Closes: #ISSUE_NUMBER


ChengYanJin (approved):
LGTM! Thanks for adding the test!!

PR #393, opened 2021-12-03
linetemporalchart: Normalize vega field string to avoid empty data in the tooltip

JBWatenbergScality:
**Component**:

linetemporalchart

**Description**:

If a serie tooltip label contains a dot ".", vega considers it as a child accessor.

**Design**:

We need to normalize serie tooltip labels to remove dots.

### DIFF
--- a/src/lib/components/linetemporalchart/ChartUtil.js
+++ b/src/lib/components/linetemporalchart/ChartUtil.js
@@ -9,6 +9,14 @@ export type VegaData = {
   isDashed: boolean,
 }[];
 
+//Given a field containing dots ".", vega is interpreting this as an accessor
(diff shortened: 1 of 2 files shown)


ChengYanJin:
After merging this PR, we might want to patch a release `v0.22.2`, since we embedded `v0.22.1` in MetalK8s 2.10.5.

JBWatenbergScality:
> After merging this PR, we might want to patch a release `v0.22.2`, since we embedded `v0.22.1` in MetalK8s 2.10.5.

Yes indeed else we would end up with navbar discrepancies between UIs (diff: https://github.com/scality/core-ui/compare/v0.22.1...development/1.0)

ChengYanJin (approved):

PR #398, opened 2021-12-14
Add 401 not authorized error page

ChengYanJin:
**Component**: 401 error page

**Description**:
- Generlize the ErrorPageStyle
- Update the `Return Home` Button with Buttonv2
- Pass onReturnHomeClick instead of `btnlink` and `history`

**Design**:
EN:
image
FR:
image

**Breaking Changes**:
We may have breaking changes if we pass btnLink and history as props to the error pages.
But after checking MetalK8s UI and XCore UI, there is no **Return Home** button on the current implementation.

### DIFF
--- a/src/lib/components/error-pages/ErrorPage401.component.js
+++ b/src/lib/components/error-pages/ErrorPage401.component.js
@@ -29,17 +29,15 @@ const translations = {
 };
 
 type Props = {
-  btnLink?: string,
   supportLink?: string,
   locale?: string,
-  history?: Object,
+  onReturnHomeClick?: (event: SyntheticMouseEvent<HTMLButtonElement>) => void,
 };
 
 function ErrorPage401({
-  btnLink = '/',
   supportLink = null,
   locale = 'en',
-  history,
+  onReturnHomeClick,
   ...rest
 }: Props) {
   if (!translations[locale]) locale = 'en';
@@ -69,11 +67,11 @@ function ErrorPage401({
         </Description>
       </Row>
 
-      {history && (
+      {onReturnHomeClick && (
         <Button
           label={translations[locale].return_home}
           variant="secondary"
-          onClick={() => history.push(btnLink)}
+          onClick={onReturnHomeClick}
         />
       )}
     </ErrorPageContainer>
--- a/src/lib/components/error-pages/ErrorPage401.component.js
+++ b/src/lib/components/error-pages/ErrorPage401.component.js
@@ -1,0 +1,83 @@
+//@flow
+import React from 'react';
+import {
+  ErrorPageContainer,
+  Row,
+  WarningIcon,
+  Title,
+  Description,
+  DescriptionContent,
+  Link,
+} from './ErrorPageStyle';
+import Button from '../buttonv2/Buttonv2.component';
+
+const translations = {
+  en: {
+    unexpected_error: 'Not authorized',
+    error_desc: `You don't have permission to view this page using the credentials that you supplied.`,
+    may_also_contact: 'You may contact ',
+    to_report_issue: ' to report this issue.',
+    return_home: 'Return Home',
+  },
+  fr: {
+    unexpected_error: `Vous n'êtes pas autorisé`,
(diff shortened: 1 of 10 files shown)


src/lib/components/error-pages/ErrorPage401.component.js
```diff
@@ -0,0 +1,83 @@
...
+    return_home: 'Return Home',
+  },
+  fr: {
+    unexpected_error: `Vous n'êtes pas autorisé`,
+    error_desc: `Vous n'êtes pas autorisé à afficher cette page en utilisant les informations d'identification que vous avez fournies.`,
+    may_also_contact: 'Vous pouvez contacter le ',
+    to_report_issue: ' pour signaler ce problème.',
+    return_home: "Retour à l'accueil",
+  },
+};
```
gaudiauj, line 29: In my opinion the translation should be dealt with in the app and not in the design-system.

src/lib/components/error-pages/ErrorPage401.component.js
```diff
@@ -0,0 +1,83 @@
...
+  supportLink?: string,
+  locale?: string,
+  history?: Object,
+};
+
+function ErrorPage401({
+  btnLink = '/',
+  supportLink = null,
+  locale = 'en',
+  history,
```
gaudiauj: i would rather have an `onReturnHome` function that let the app do whatever it want (history.push history.goBack window.reload) based on the context. rather than sending the history object and a link.
wdyt ?

src/lib/components/error-pages/ErrorPageStyle.jsx
```diff
@@ -0,0 +1,48 @@
...
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  height: 100vh;
+`;
+
+export const Title = styled.h2`
+  color: ${getThemePropSelector('textPrimary')};
+  margin: 0;
```
gaudiauj, line 16: is the `margin: 0` useful here ?

src/lib/components/error-pages/ErrorPage401.component.js
ChengYanJin: You are absolutely correct here! 
It's really cumbersome to pass the history object. 
I will just check we don't really use this `btnlink` for the moment, so with this change we won't introduce breaking change!

src/lib/components/error-pages/ErrorPage401.component.js
ChengYanJin: I agree that we should keep the translations within the UIs. 
However, for the error pages, we'd like to reuse the component as much as possible. 
Otherwise,  we will end up duplicating the same translations in all the UIs.

src/lib/components/error-pages/ErrorPageStyle.jsx
ChengYanJin: Yes :D 
I have a margin around the title because we styled `h2`. Let's keep the current style if it doesn't bother you too much?

gaudiauj (approved):
approved

src/lib/components/error-pages/ErrorPageStyle.jsx
```diff
@@ -0,0 +1,48 @@
...
+import { spacing, fontSize } from '../../style/theme';
+import { getThemePropSelector } from '../../utils';
+
+export const ErrorPageContainer = styled.div`
+  background-color: ${getThemePropSelector('backgroundLevel1')};
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  height: 100vh;
```
JBWatenbergScality: nit: I'm not super sure about this `height: 100vh;` I think in theory it should match the height of the parent. (e.g if we display this component below the navbar a scrollbar will be displayed, we already have this pb so this is nitpicking...)

JBWatenbergScality (approved):

src/lib/components/error-pages/ErrorPageStyle.jsx
ChengYanJin: You are right. height: 100% is enough.

PR #399, opened 2021-12-16
healthSelector: add the new health selector

gaudiauj:
**Component**:

healthSelector 

**Description**:

create a new healthSelector base on selectorV2

**Design**:

https://www.figma.com/file/N2jOAK44dTSiBclDjXzz6a/Next_Gen_Design?node-id=3915%3A58591

**Breaking Changes**:

none

- [] Breaking Changes

### DIFF
--- a/src/lib/components/healthselectorv2/HealthSelector.component.js
+++ b/src/lib/components/healthselectorv2/HealthSelector.component.js
@@ -14,16 +14,15 @@ import {
   icons,
 } from './HealthSelector.style';
 
-type optionType = 'all' | 'healthy' | 'warning' | 'critical' | 'unknown';
-type optionValue = {
+type OptionType = 'all' | 'healthy' | 'warning' | 'critical' | 'unknown';
+type OptionValue = {
   shortLabel?: string,
-  isHidden?: boolean,
 } & OptionProps;
 
-type OptionsConfiguration = { [key: optionType | string]: optionValue };
+type OptionsConfiguration = { [key: OptionType]: OptionValue };
 
 type Props = {
-  optionsConfiguration?: OptionsConfiguration,
+  options?: OptionValue[],
   label?: string,
 } & SelectProps;
 
@@ -33,53 +32,46 @@ export const optionsDefaultConfiguration: OptionsConfiguration = {
     label: 'All healthy statuses',
     shortLabel: 'All',
     value: 'all',
-    isHidden: false,
   },
   healthy: {
     icon: icons.healthy,
     label: 'Healthy only',
     shortLabel: 'Healthy',
     value: 'healthy',
-    isHidden: false,
   },
   critical: {
     icon: icons.critical,
     label: 'Critical only',
     shortLabel: 'Critical',
     value: 'critical',
-    isHidden: false,
   },
   warning: {
     icon: icons.warning,
     label: 'Warning only',
     shortLabel: 'Warning',
     value: 'warning',
-    isHidden: false,
   },
   unknown: {
     icon: icons.unknown,
     label: 'Unknown only',
     shortLabel: 'Unknown',
     value: 'unknown',
-    disabled: false,
   },
 };
 
+export const defaultOptions: OptionValue[] = Object.keys(
+  optionsDefaultConfiguration,
+).map((key) => {
+  return optionsDefaultConfiguration[key];
+});
+
 function HealthSelectorv2(props: Props) {
   const {
-    optionsConfiguration = optionsDefaultConfiguration,
+    options = defaultOptions,
     label = 'Health',
     onChange,
     ...rest
   } = props;
-
-  const options = [];
-
-  Object.keys(optionsConfiguration).forEach((key) => {
-    if (!optionsConfiguration[key].isHidden) {
-      options.push(optionsConfiguration[key]);
-    }
-  });
 
   const SingleValue = (props) => {
     return (
--- a/src/lib/components/healthselectorv2/HealthSelector.component.js
+++ b/src/lib/components/healthselectorv2/HealthSelector.component.js
@@ -1,0 +1,121 @@
+//@flow
+import * as React from 'react';
+import { components } from 'react-select';
+import type { OptionProps } from '../selectv2/Selectv2.component';
+import type { SelectProps } from '../selectv2/Selectv2.component';
+import {
+  SingleValueWrapper,
+  ValueIcon,
+  OptionIcon,
+  OptionWrapper,
+  OptionLabel,
+  SelectStyle,
+  ShortLabel,
+  icons,
+} from './HealthSelector.style';
+
+type optionType = 'all' | 'healthy' | 'warning' | 'critical' | 'unknown';
+type optionValue = {
+  shortLabel?: string,
+  isHidden?: boolean,
+} & OptionProps;
+
+type OptionsConfiguration = { [key: optionType | string]: optionValue };
+
+type Props = {
+  optionsConfiguration?: OptionsConfiguration,
+  label?: string,
+} & SelectProps;
+
+export const optionsDefaultConfiguration: OptionsConfiguration = {
+  all: {
+    icon: icons.all,
+    label: 'All healthy statuses',
+    shortLabel: 'All',
+    value: 'all',
+    isHidden: false,
+  },
+  healthy: {
+    icon: icons.healthy,
+    label: 'Healthy only',
+    shortLabel: 'Healthy',
+    value: 'healthy',
+    isHidden: false,
+  },
+  critical: {
+    icon: icons.critical,
+    label: 'Critical only',
+    shortLabel: 'Critical',
+    value: 'critical',
+    isHidden: false,
+  },
+  warning: {
+    icon: icons.warning,
+    label: 'Warning only',
+    shortLabel: 'Warning',
+    value: 'warning',
+    isHidden: false,
+  },
+  unknown: {
+    icon: icons.unknown,
+    label: 'Unknown only',
+    shortLabel: 'Unknown',
+    value: 'unknown',
+    disabled: false,
+  },
+};
+
+function HealthSelectorv2(props: Props) {
+  const {
+    optionsConfiguration = optionsDefaultConfiguration,
+    label = 'Health',
+    onChange,
+    ...rest
+  } = props;
+
+  const options = [];
+
+  Object.keys(optionsConfiguration).forEach((key) => {
+    if (!optionsConfiguration[key].isHidden) {
+      options.push(optionsConfiguration[key]);
+    }
+  });
+
+  const SingleValue = (props) => {
+    return (
+      <SingleValueWrapper>
+        <p data-testid="singleValueLabel">{label}</p>{' '}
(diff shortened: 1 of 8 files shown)


Cuervino (changes requested):
image

1. The component is not wide enough, the word "unknown" is collapsing with the small separator

2. No focus state when the keyboard is not used yet. 

3. Moreover, here there is a double focus state?

image

4. Better avoid such strange examples... It's a public repo.
Also, keep it english examples (no coucou, hello is ok).

src/lib/components/healthselectorv2/HealthSelector.component.js
```diff
@@ -0,0 +1,233 @@
...
+type optionType = 'all' | 'healthy' | 'warning' | 'critical' | 'unknown';
+type optionValue = {
+  icon?: React.Node,
+  label?: string,
+  shortLabel?: string,
+  value?: string,
+  disabled?: boolean,
+};
+
+type optionsConfiguration = { [optionType]: optionValue };
```
JBWatenbergScality: nit: We usually used capitalised CamelCase for custom type definition

src/lib/components/healthselectorv2/HealthSelector.component.js
```diff
@@ -0,0 +1,233 @@
...
+
+type Props = {
+  optionsConfiguration?: optionsConfiguration,
+  label?: String,
```
JBWatenbergScality:
```suggestion
  label?: string,
```
Primitive types in flow are not capitalised

src/lib/components/healthselectorv2/HealthSelector.component.js
```diff
@@ -0,0 +1,233 @@
...
+  }
+`;
+
+const icons = {
+  all: (
+    <ThreeCirclesWrapper>
+      <CSSCircle variant="statusHealthy" />
+      <CSSCircle variant="statusWarning" />
+      <CSSCircle variant="statusCritical" />
+    </ThreeCirclesWrapper>
```
JBWatenbergScality: nit: I would preferably use a svg here (In my mind declarative always win over imperative for these things) 
```suggestion
    <svg viewBox="0 0 200 100" xmlns="http://www.w3.org/2000/svg">
  <circle cx="50" cy="50" r="50" fill={theme.brand["statusHealthy"]}/>
  <circle cx="75" cy="50" r="50" fill={theme.brand["statusWarning"]}/>
  <circle cx="100" cy="50" r="50" fill={theme.brand["statusCritical"]}/>
</svg>
```
should do the trick I think but I haven't tested

src/lib/components/healthselectorv2/HealthSelector.component.js
```diff
@@ -0,0 +1,233 @@
...
+  ),
+  healthy: <Icon name={'Check-circle'} color={'statusHealthy'} size={'lg'} />,
+  warning: (
+    <Icon name={'Exclamation-circle'} color={'statusWarning'} size={'lg'} />
+  ),
+  critical: <Icon name={'Times-circle'} color={'statusCritical'} size={'lg'} />,
+  unknown: <Icon name={'Info'} color={'infoPrimary'} size={'lg'} />,
+};
+
+const optionsDefaultConfiguration: optionsConfiguration = {
```
JBWatenbergScality: If an application want to hide one of those configurations, it will need to redo this whole mapping which may be a bit cumbersome. 
In order to improve dev experience we could refactor this to individual configurations that are exported so that it is easy to compose with them from one of the app

src/lib/components/healthselectorv2/HealthSelector.component.js
```diff
@@ -0,0 +1,233 @@
...
+};
+
+function HealthSelectorv2(props: Props) {
+  const {
+    optionsConfiguration,
+    label = 'Health',
+    onChange,
+    defaultValue = 'all',
+    ...rest
+  } = props;
```
JBWatenbergScality: I don't see any way aside of the defaultValue from the props point of view to control the component. Eg. if the value is controlled from a parent component how do we pass it to this one ? (for e.g how to sync it with the url param without a page refresh/unmount-remount with the current implementation ?)

src/lib/components/healthselectorv2/HealthSelector.component.js
```diff
@@ -0,0 +1,233 @@
...
+
+  const options = [];
+
+  Object.keys({
+    ...optionsDefaultConfiguration,
+    ...optionsConfiguration,
```
JBWatenbergScality: instead of merging these config I would just pass a bunch of optionsConfig as a prop and use that prop directly here to ease composability. By default if no options is passed we use the whole default ones.

src/lib/components/healthselectorv2/HealthSelector.component.js
```diff
@@ -0,0 +1,233 @@
...
+
+  return (
+    <SelectStyle
+      components={{ SingleValue }}
+      onChange={onChange}
+      defaultValue={defaultValue}
+      {...rest}
+    >
+      {options.map(
+        (optionValue, i) =>
```
JBWatenbergScality:
```suggestion
        (optionValue) =>
```

src/lib/components/healthselectorv2/HealthSelector.component.js
```diff
@@ -0,0 +1,233 @@
...
+          !optionValue.isHidden && (
+            <Option
+              key={optionValue.value}
```
JBWatenbergScality: If we display 2 health selector (which is one of our use cases !) on the same page react will complain about duplicated keys I think. We may need to prefix this key somehow, to be checked

src/lib/components/healthselectorv2/HealthSelector.component.test.js
```diff
@@ -0,0 +1,51 @@
+import HealthSelector from './HealthSelector.component';
+import React from 'react';
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+describe('HealthSelector', () => {
+  test('it should display correctly without any props and select first option', () => {
```
JBWatenbergScality: nit: you can use `it` directly to define a test case, so that you can skip the 'it' in the test case definition
```suggestion
  it('should display correctly without any props and select first option', () => {
```

stories/healthselectorv2.stories.js
```diff
@@ -0,0 +1,70 @@
...
+            },
+            all: {
+              label: 'patrick',
+              shortLabel: 'late',
+            },
+            warning: {
+              label: 'attention',
+              shortLabel: 'ouch',
+              icon: <Icon name="Tape" />,
+              value: 'myValue',
```
JBWatenbergScality: Please avoid french and strange values in your test cases. Ideally having business oriented test cases is always better

JBWatenbergScality (changes requested):

Cuervino (changes requested):
1. I think that the icon should be responsive, or at least have the same behavior

image

2. There is a strange "moving"/"resizing" effect when the input is clicked.
The additional border should not make the content move, or the separator to change position/size.

https://user-images.githubusercontent.com/28654802/147126114-314901d9-4958-4275-99b7-41c1560f24a5.mp4

Cuervino (approved):
LGTM

src/lib/components/healthselectorv2/HealthSelector.component.js
```diff
@@ -0,0 +1,121 @@
...
+  OptionIcon,
+  OptionWrapper,
+  OptionLabel,
+  SelectStyle,
+  ShortLabel,
+  icons,
+} from './HealthSelector.style';
+
+type optionType = 'all' | 'healthy' | 'warning' | 'critical' | 'unknown';
+type optionValue = {
```
JBWatenbergScality: nit: Case of typings
```suggestion
type OptionType = 'all' | 'healthy' | 'warning' | 'critical' | 'unknown';
type OptionValue = {
```

src/lib/components/healthselectorv2/HealthSelector.style.js
```diff
@@ -0,0 +1,130 @@
...
+  width: 34px;
+  justify-content: center;
+  ${CSSCircle}:first-child {
+    margin-right: -20%;
+  }
+  ${CSSCircle}:last-child {
+    margin-left: -20%;
+  }
+  height: 16px;
+`;
```
JBWatenbergScality: I think this is not used anymore

stories/healthselectorv2.stories.js
```diff
@@ -0,0 +1,79 @@
...
+          onChange={(value) => {
+            action(value);
+          }}
+          optionsConfiguration={{
+            ...optionsDefaultConfiguration,
+            healthy: {
+              ...optionsDefaultConfiguration.healthy,
+              isHidden: true,
+            },
+          }}
```
JBWatenbergScality: Instead of having a `isHidden` attribute here I would just do 
```suggestion
          optionsConfiguration={[
              optionsDefaultConfiguration.all,
              optionsDefaultConfiguration.critical,
              optionsDefaultConfiguration.warning,
              optionsDefaultConfiguration.unknown,
            ]}
```

src/lib/components/healthselectorv2/HealthSelector.component.js
```diff
@@ -0,0 +1,121 @@
...
+    ...rest
+  } = props;
+
+  const options = [];
+
+  Object.keys(optionsConfiguration).forEach((key) => {
+    if (!optionsConfiguration[key].isHidden) {
+      options.push(optionsConfiguration[key]);
+    }
+  });
```
JBWatenbergScality: To me optionsConfiguration can directly be the target array

JBWatenbergScality (changes requested):

src/lib/components/healthselectorv2/HealthSelector.style.js
gaudiauj: indeed

src/lib/components/healthselectorv2/HealthSelector.style.js
```diff
@@ -0,0 +1,117 @@
...
+
+export const CSSCircle = styled.span`
+  background-color: ${(props) => defaultTheme.brand[props.variant]};
+  border-radius: 50%;
+  display: flex;
+  width: 16px;
+  height: 16px;
+  align-items: center;
+  justify-content: center;
+`;
```
JBWatenbergScality:
```suggestion

```

src/lib/components/healthselectorv2/HealthSelector.style.js
```diff
@@ -0,0 +1,107 @@
+import React from 'react';
+import styled from 'styled-components';
+import Select from '../selectv2/Selectv2.component';
+import Icon from '../icon/Icon.component';
+import * as defaultTheme from '../../style/theme';
```
JBWatenbergScality, line 5: we should likely used currently selected theme instead of the default one. 

To do so in styled component you can : ${props.theme[colorName]}.

Spacing can be imported directly through as it is not changing over selected theme

JBWatenbergScality (approved):

PR #400, opened 2021-12-16
Improvement/table v2 with textconstrained

gaudiauj:
**Component**:

TableV2

**Description**:

use textConstrained component to add ellipsis when no customCell is used
Capture d’écran du 2021-12-16 17-06-00

It also fix the issue with table Head not being the same width as the rest of the table when the scroll bar is displayed

**Design**:
no changes / better allignement when the scroll bar is displayed

**Breaking Changes**:

Ellipsis on default cell

### DIFF
--- a/src/lib/components/tablev2/SingleSelectableContent.js
+++ b/src/lib/components/tablev2/SingleSelectableContent.js
@@ -16,6 +16,7 @@ import { useTableContext } from './Tablev2.component';
 import { convertRemToPixels } from './TableUtil';
 import { Row } from 'react-table'; // may not import the type correctly
 import Tooltip from '../tooltip/Tooltip.component.js';
+import ConstrainedText from '../constrainedtext/Constrainedtext.component';
 
 export const tableRowHeight = {
   // in rem unit
@@ -117,7 +118,18 @@ export default function SingleSelectableContent({
 
             return (
               <div {...cellProps} className="td">
-                {cell.render('Cell')}
+                {
+                  // if cell.column.cell is defined that means we have a custom cell renderer
+                  // otherwise it should be just a string and we can use the constrainted text component
+                  !cell.column.cell && typeof cell.value === 'string' ? (
+                    <ConstrainedText
+                      text={cell.value}
+                      tooltipPlacement={index === 0 ? 'bottom' : 'top'}
+                    />
+                  ) : (
+                    cell.render('Cell')
+                  )
+                }
               </div>
             );
           })}
--- a/src/lib/components/tablev2/SingleSelectableContent.js
+++ b/src/lib/components/tablev2/SingleSelectableContent.js
@@ -1,5 +1,5 @@
 //@flow
-import React from 'react';
+import React, { useMemo } from 'react';
 import AutoSizer from 'react-virtualized-auto-sizer';
 import { FixedSizeList as List } from 'react-window';
 import {
@@ -49,6 +49,8 @@ export default function SingleSelectableContent({
   selectedId,
   onRowSelected,
 }: SingleSelectableContentProps) {
+  const [hasScrollbar, setHasScrollbar] = React.useState(false);
+
   if (!['h32', 'h40', 'h48', 'h64'].includes(rowHeight)) {
     console.error(
       `Invalid rowHeight props, expected h32, h40, h48, or h64 but received ${rowHeight}`,
@@ -100,6 +102,7 @@ export default function SingleSelectableContent({
                 display: 'flex',
                 flexDirection: 'column',
                 justifyContent: 'center',
+                'flex-grow': 1,
               },
               role: 'gridcell',
             });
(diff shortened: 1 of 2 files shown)


src/lib/components/tablev2/SingleSelectableContent.js
```diff
@@ -117,7 +122,18 @@ export default function SingleSelectableContent({
 
             return (
               <div {...cellProps} className="td">
-                {cell.render('Cell')}
+                {
+                  // if cell.column.cell is defined that means we have a custom cell renderer
+                  // otherwise it should be just a string and we can use the constrainted text component
+                  !cell.column.cell && typeof cell.value === 'string' ? (
+                    <ConstrainedText
```
ChengYanJin, line 128: 1. IMO, the tooltip is there **only** when we don't have enough space to display all the text in the cell.
   I don't think if it makes sense to have this tooltip in the following case:


2. I think we should have some padding between the ConstrainedText and the next cell.

image
@Cuervino WDYT?

src/lib/components/tablev2/Tablestyle.js
```diff
@@ -26,7 +26,10 @@ export const TableHeader = styled.div`
 
 export const HeadRow = styled.div`
   height: 2.286rem;
-  width: 100% !important;
+  width: ${(props) =>
+    props.hasScrollBar
+      ? `calc(100% - ${props.scrollBarWidth}px - 4px )!important;` // -4px for border
```
ChengYanJin: Ideally, we should extract the 4px to a prop.
If we hardcode the number here, one day when we update the border we are going to easily forget it here.

src/lib/components/tablev2/SingleSelectableContent.js
```diff
@@ -117,7 +122,18 @@ export default function SingleSelectableContent({
...
+                      text={cell.value}
+                      tooltipPlacement={index === 0 ? 'bottom' : 'top'}
```
ChengYanJin, line 130: Probably an issue with the tooltip here too. 
image

src/lib/components/tablev2/SingleSelectableContent.js
```diff
@@ -173,10 +202,20 @@ export default function SingleSelectableContent({
...
               width={width}
+              onItemsRendered={({
+                visibleStartIndex,
+                visibleStopIndex,
+                overscanStopIndex,
+              }) => {
+                setHasScrollbar(
+                  visibleStopIndex - visibleStartIndex < overscanStopIndex,
+                );
+              }}
```
JBWatenbergScality: I think this can be dropped.  If outerRef.offsetWidth == innerRef.offsetWidth then in TabelStyle:

```
width: calc(100% - ${props.scrollBarWidth> 0?props.scrollBarWidth-4:0}px )!important;// -4px for border 
```

src/lib/components/tablev2/SingleSelectableContent.js
```diff
@@ -48,6 +49,9 @@ export default function SingleSelectableContent({
   selectedId,
   onRowSelected,
 }: SingleSelectableContentProps) {
+  const [hasScrollbar, setHasScrollbar] = React.useState(false);
```
JBWatenbergScality, line 52: This state can be dropped as it can be derived from `scrollWidth`

src/lib/components/tablev2/SingleSelectableContent.js
```diff
@@ -134,11 +150,24 @@ export default function SingleSelectableContent({
     ],
   );
 
+  React.useEffect(() => {
+    if (listRef.current) {
+      const outerRef = listRef.current._outerRef;
```
JBWatenbergScality: I think the approach is good I just wonder why they prefixed outerRef with an underscore. I makes me feel that the lib want to keep this field private to be able to evolve it (to be checked). I would better not use it if they really want to keep it private. To workaround this we can put a ref on the parent and another one on one of the child

gaudiauj: well it's definitly a 'private' props, as it is not in the doc. But we need to access to this particular dom element since it's where is the scrollBar.
We could get a ref from the parent element and get the element with `.firstchild` but the same issue arise, what if the library change, and add more elements between the scroll container and the parent ?

src/lib/components/tablev2/Tablestyle.js
```diff
@@ -1,7 +1,10 @@
 import styled, { css } from 'styled-components';
+import AutoSizer from 'react-virtualized-auto-sizer';
```
JBWatenbergScality: It seems to be an unused import

src/lib/components/tablev2/Tablestyle.js
```diff
@@ -26,7 +29,10 @@ export const TableHeader = styled.div`
 
 export const HeadRow = styled.div`
   height: 2.286rem;
-  width: 100% !important;
+  width: ${(props) =>
+    props.hasScrollBar
+      ? `calc(100% - ${props.scrollBarWidth}px - ${borderSize} )!important;` // -4px for border
+      : '100% !important;'}
```
JBWatenbergScality, line 34: nit: I don't think that the hasScrollBar props is necessary, we can probably do :
```suggestion
  width: ${(props) =>
    props.scrollBarWidth
      ? `calc(100% - ${props.scrollBarWidth}px - ${borderSize} )!important;` // -4px for border
      : '100% !important;'} 
```

Then in src/lib/components/tablev2/SingleSelectableContent.js the state could be removed

src/lib/components/tablev2/Tablestyle.js
gaudiauj: don't know how it got here thanks !

src/lib/components/tablev2/Tablestyle.js
gaudiauj: we can not do that :) 

as we discussed, the scrollbarWidth is calculated by creating a new div and getting the scroll bar size : https://github.com/scality/core-ui/pull/400/files#diff-c49bb2bd3fc72cb8413901b8bf5b5f6ce38a75db48670f24165ee4e41bc18e26R152-R163

so we always have scrollbarWidth even if the scrollBar is not displayed. So we need to know if the scrollBar is displayed or not with this props.

JBWatenbergScality: Oh yes that's right 👍

JBWatenbergScality (approved):
I haven't tested it yet but the implementation looks good.

PR #402, opened 2021-12-22
Improvement/artesca 2986 tooltip position

gaudiauj:
**Component**:

Tooltip

**Description**:

use [floating-ui](https://floating-ui.com/) for tooltip position. Allow to shift the floating element in view to prevent overflow.

**Design**:
same

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,16 +78,16 @@
     "react-table": "^7.7.0",
     "react-test-renderer": "^16.9.0",
     "react-virtualized": "^9.21.1",
+    "react-virtualized-auto-sizer": "^1.0.5",
     "react-window": "^1.8.6",
     "regenerator-runtime": "^0.13.7",
-    "react-virtualized-auto-sizer": "^1.0.5",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -78,16 +78,16 @@
     "react-table": "^7.7.0",
     "react-test-renderer": "^16.9.0",
     "react-virtualized": "^9.21.1",
+    "react-virtualized-auto-sizer": "^1.0.5",
```
JBWatenbergScality, line 81: This change doesn't seems to relate to the scope of this pr, is it really necessary ?

gaudiauj: there is no change, npm decided to change the position on his own when i did the `npm i` (alphabetical order)

Cf: https://github.com/scality/core-ui/pull/402/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L83 same number version different position

JBWatenbergScality (approved):

PR #403, opened 2022-01-05
Improvement/artesca 3037 table search

gaudiauj:
**Component**:

Table V2

**Description**:

Add the search bar to the table

Capture d’écran du 2022-01-05 10-32-22
Capture d’écran du 2022-01-05 10-32-41


Cuervino (changes requested):
### General table comments:
1. Not enough space at the beginning of the table (should be rem equivalent of 8px)
2. Selectable table doesn't have selectable rows
3. The hover row blue color is not going until the end at the right
4. The column "age" is numerical, thus should be right-aligned

image

### Search-related comments
5. When result is 0 or 1 entity, the plural should be removed
6. Remark: IMO, when the search is "ON", it should be more obvious that there is a filter. 
	Could be by keeping the search input in a specific status (focused for instance), or to make the search a "tag".
7. Remark: the search is dynamic, and doesn't need to be triggered. It works well indeed.
8. Remark: the search implemented is all over the table, at any content? Currently, in the search, we are displaying for instance "Search by Entity name", but if the search is global, we could just put "Search" as placeholder, without additional info.


image

image

src/lib/components/tablev2/Tablev2.test.js
```diff
@@ -0,0 +1,150 @@
...
+                  <span {...cell.getCellProps()}>{cell.render('Cell')}</span>
+                );
+              })}
+            </span>
+          );
+        })}
+      </div>
+    </>
+  );
+}
```
JBWatenbergScality: instead of mocking the Table child, which is partly the component that we want to test I would only mock the component that is causing problems hence `autoSizer` from react-virtualized-auto-sizer. A simple way of doing so would be to mock react-virtualized module and override the AutoSizer component to be a dumb div rendering it's children I guess.

JBWatenbergScality: You can get some intel about how to do that in this gist I think https://gist.github.com/elevine/69a27e067032c53d64d66f4f56e978cc

src/lib/components/tablev2/Tablev2.component.js
```diff
@@ -119,4 +145,6 @@ function Table({
 }
 
 Table.SingleSelectableContent = SingleSelectableContent;
+Table.Search = Search;
+Table.SearcWithQueryParams = SearchWithQueryParams;
```
JBWatenbergScality: I think there is a typo here:
```suggestion
Table.SearchWithQueryParams = SearchWithQueryParams;
```

src/lib/components/tablev2/Tablev2.component.js
```diff
@@ -97,12 +116,19 @@ function Table({
     useRowSelect,
   );
 
+  useEffect(() => {
+    if (globalFilter != undefined) {
```
ChengYanJin, line 120:
```suggestion
    if (globalFilter !== undefined) {
```

ChengYanJin:
Hello @Cuervino 
image
do we have the design for the table when there is no result appears following the search?

Cuervino:
> Hello @Cuervino
> image
> do we have the design for the table when there is no result appears following the search?

Not yet, but here is my proposal:
image
With statusWarning color.

JBWatenbergScality:
> > Hello @Cuervino
> > image
> > do we have the design for the table when there is no result appears following the search?
> 
> Not yet, but here is my proposal: image With statusWarning color.

Shouldn't we add an empty placeholder instead of displaying nothing below the header ? Like a search engine behavior:
image

src/lib/components/tablev2/Tablev2.component.js
gaudiauj: it's actually to avoid to write `globalFilter !== undefined && globalFilter !== null &&  globalFilter !== ''`

JBWatenbergScality: Then just `!globalFilter` would do the trick

Cuervino:
> Shouldn't we add an empty placeholder instead of displaying nothing below the header ? Like a search engine behavior: image

So it could be looking like that?

image

I think something should be still written with the "Total", for keeping consistency and Search bar position.

gaudiauj:
replaced by : https://github.com/scality/core-ui/pull/404

PR #404, opened 2022-01-11
Improvement/artesca 3037 table search and fixes

gaudiauj:
**Component**:

Table V2

**Description**:

Add the search bar to the table

Capture d’écran du 2022-01-05 10-32-22
Capture d’écran du 2022-01-05 10-32-41

come from https://github.com/scality/core-ui/pull/403


src/lib/components/tablev2/SingleSelectableContent.js
```diff
@@ -40,6 +41,16 @@ export type SingleSelectableContentProps = {
...
   onRowSelected?: (row: Row) => void,
+  locale?: 'en' | 'fr',
+};
+
+const translations = {
+  en: {
+    noResult: 'No results found',
+  },
+  fr: {
+    noResult: `aucun résultat`,
```
gaudiauj:
```suggestion
    noResult: `Aucun résultat`,
```

src/lib/components/tablev2/Search.js
```diff
@@ -0,0 +1,92 @@
...
+
+  const totalDispayedRows = rows.length;
+
+  React.useEffect(() => {
+    setGlobalFilter(value);
+  }, [value, setGlobalFilter]);
+
+  return (
+    <SearchContainer>
+      {displayedName && displayTotalOf && (
```
JBWatenbergScality: As `displayedName` and `displayTotalOf` are 2 dissociated props I believe that we should be able to use them in a separated way. So providing only `displayTotalOf` would display `Total: XX`, providing an additional `displayedName` would result in `Total: XX {displayedName}`.
If we take the opposite approach where we only want to display the total when the `displayedName` is provided then I think a single props should be exposed in the component definition : something like `displayTotalOf={{plural, singular}}` would be enough

src/lib/components/tablev2/SearchWithQueryParams.js
```diff
@@ -0,0 +1,38 @@
...
+} & SearchProps;
+
+export default function SearchWithQueryParams(props) {
+  const { queryParams = 'search', onChange, ...rest } = props;
+  const match = useRouteMatch();
+  const { search } = useLocation();
+  const history = useHistory();
+  const params = new URLSearchParams(search);
+  const initialValue = params.get(queryParams);
+  const [value, setValue] = React.useState(initialValue);
```
JBWatenbergScality: nit: In order to benefit from tree shaking of React dependency (possible in React 17+) I usually prefer to import only useState and not the global React object

src/lib/components/tablev2/SearchWithQueryParams.js
```diff
@@ -0,0 +1,38 @@
...
+      onChange(value);
+    }
+  }
+
+  return (
+    <Search
+      onChange={handleOnChange}
+      value={value}
+      onReset={() => {
+        setValue('');
```
JBWatenbergScality: I think we likely also want to reset the query param here

src/lib/components/tablev2/Search.js
```diff
@@ -0,0 +1,92 @@
...
+  singular: string,
+};
+
+export type SearchProps = {
+  onChange: (string) => void,
+  value?: string,
+  displayTotalOf?: boolean,
+  displayedName?: DisplayedName,
+  locale?: 'en' | 'fr',
+};
```
JBWatenbergScality: It should extends the SearchInput props I guess. 
```suggestion
export type SearchProps = {
  onChange: (string) => void,
  displayTotalOf?: boolean,
  displayedName?: DisplayedName,
  locale?: 'en' | 'fr',
} & SearchInputProps;
```

src/lib/components/tablev2/SingleSelectableContent.js
```diff
@@ -121,18 +134,15 @@ export default function SingleSelectableContent({
...
-                    />
-                  ) : (
-                    cell.render('Cell')
-                  )
-                }
+                {cell.column.Cell.name === 'defaultRenderer' &&
+                typeof cell.value === 'string' ? (
+                  <ConstrainedText
+                    text={cell.value}
+                    tooltipPlacement={index === 0 ? 'bottom' : 'top'}
```
JBWatenbergScality: do we still need this with the new tooltip ?

src/lib/components/tablev2/SingleSelectableContent.js
```diff
@@ -150,17 +160,21 @@ export default function SingleSelectableContent({
...
+      );
+      // $FlowFixMe
+      outerRef.current.appendChild(scrollDiv);
+      const scrollbarWidth = scrollDiv.offsetWidth - scrollDiv.clientWidth;
+      // $FlowFixMe
+      outerRef.current.removeChild(scrollDiv);
+      return scrollbarWidth;
+    }
+    return 0;
+  }, [outerRef.current]);
```
JBWatenbergScality: Can you check this warning ? Maybe just mentioning outerRef in the dependency array would solve it ?

> React Hook useMemo has an unnecessary dependency: 'outerRef.current'. Either exclude it or remove the dependency array. Mutable values like 'outerRef.current' aren't valid dependencies because mutating them doesn't re-render the component

src/lib/components/tablev2/Tablev2.component.js
```diff
@@ -97,12 +116,19 @@ function Table({
     useRowSelect,
   );
 
+  useEffect(() => {
+    if (globalFilter != undefined) {
```
JBWatenbergScality:
```suggestion
    if (globalFilter) {
```

JBWatenbergScality (changes requested):

Cuervino (changes requested):
1. No need of a space before colons (so "Total: 2 entities" instead of "Total : 2 entities)
2. I would put the noResult: 'No results found' in normal weight (and not font-weight: bold)
3. I would prefer the Searchbar not to change place depending on the result (the Searchbar location jumps from one place to another if it's "Total: 0" or "Total: 4 peoples"
4. "Peoples" is not really English, or at least not in this context. Maybe you could use "person" instead?

src/lib/components/tablev2/SingleSelectableContent.js
gaudiauj: it should be able to move by itself yes

src/lib/components/tablev2/Tablev2.component.js
gaudiauj: if `globalFilter === ''` it should get into the if,
so i can leave it like this or for more expicit version : 
`globalFilter !== undefined && globalFilter !== null`

WDYT @JBWatenbergScality  ?

Cuervino (approved):
LGTM!

src/lib/components/tablev2/SingleSelectableContent.js
```diff
@@ -121,18 +131,12 @@ export default function SingleSelectableContent({
...
-                  ) : (
-                    cell.render('Cell')
-                  )
-                }
+                {cell.column.Cell.name === 'defaultRenderer' &&
+                typeof cell.value === 'string' ? (
+                  <ConstrainedText text={cell.value} />
+                ) : (
+                  cell.render('Cell')
+                )}
```
ChengYanJin: we need to change the constraintedText behavior .. if there is enough place, we should not have the tooltip displayed

src/lib/components/tablev2/Tablev2.component.js
JBWatenbergScality: Let's go for the explicit version `globalFilter !== undefined && globalFilter !== null `

src/lib/components/tablev2/SingleSelectableContent.js
gaudiauj: testé et pas approuvé, ça ne marche pas sans le `.current`

JBWatenbergScality: hmm probably because this should be an effect as we need to wait for a first render to get the ref initialised 🤔

JBWatenbergScality (approved):
LGTM !

PR #413, opened 2022-02-20
table: Make the table composable

ChengYanJin:
**Component**:
Add loading status to the table in order to display the `loading...` text in the table inside of `not found`.

**Description**:

**Design**:

### DIFF
--- a/src/lib/components/tablev2/SingleSelectableContent.js
+++ b/src/lib/components/tablev2/SingleSelectableContent.js
@@ -1,5 +1,5 @@
 //@flow
-import React, { useCallback, memo } from 'react';
+import React, { useCallback, memo, type Node } from 'react';
 import AutoSizer from 'react-virtualized-auto-sizer';
 import { FixedSizeList as List, areEqual } from 'react-window';
 import {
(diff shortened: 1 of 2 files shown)


JBWatenbergScality (commented):
I'm not very confident with growing table options with status information, because in the current status we will be missing error cases. I think composition here would be a better approach

JBWatenbergScality:
Sorry I think I wasn't clear enough

In the design we made some time ago the table v2 api was:

```js

<Table.SingleSelectableContent>
<Table.Row row={{key: 'value', ...}}></Table.Row>
</Table.SingleSelectableContent>

```
This design was composable ready because we could render a loading or error state instead of Rows within Table.SingleSelectableContent. However the implementation diverged and we end up with a data prop where composition is not really possible anymore.

My proposal to get back to a composable UI would be:
```js

<Table.SingleSelectableContent>
<Table.Rows data={[{key: 'value', ...}, ...]}></Table.Row>
</Table.SingleSelectableContent>

```

And then UIs would be able to render whatever state they would like within the table, being loading or error states

src/lib/components/tablev2/SingleSelectableContent.js
```diff
@@ -215,40 +253,12 @@ export function SingleSelectableContent({
         ))}
       </div>
       <TableBody role="rowgroup" className="tbody" ref={handleScrollbarWidth}>
-        {rows.length === 0 ? (
-          <NoResult>{translations[locale].noResult}</NoResult>
+        {typeof children === 'function' ? (
+          children(Rows)
+        ) : rows.length ? (
+          <>{Rows()}</>
```
JBWatenbergScality:
```suggestion
          <Rows />
```

JBWatenbergScality (approved):

PR #416, opened 2022-02-25
input: add disabled state

JBWatenbergScality:
image

Cursor : not-allowed is used in combination with a 0.3 opacity


Cuervino (changes requested):
The "Search" icon should also be included in the opacity.
The opacity that displays a disability in Figma is set to 0.3, not 0.5.

Cuervino (approved):
Top!

PR #422, opened 2022-04-14
Fix selected tab behavior

AsmaaELM:
Issue: https://scality.atlassian.net/browse/ARTESCA-4233


ChengYanJin (approved):

JBWatenbergScality (changes requested):
The focus ring shouldn't be displayed on click on the element but only when using keyboard navigation

PR #435, opened 2022-05-24
select: fix the reset

ChengYanJin:
**Component**: SelectV2

**Description**:
To fix the issue we can't set the value to null

add test for this scenario

### DIFF
--- a/src/lib/components/selectv2/Selectv2.component.tsx
+++ b/src/lib/components/selectv2/Selectv2.component.tsx
@@ -235,6 +235,11 @@ function SelectBox({
   className,
   ...rest
 }: SelectProps) {
+  if (defaultValue && value) {
+    console.error(
+      'The `defaultValue` will be overridden by the `value` if they are set at the same time.',
+    );
+  }
   const [keyboardFocusEnabled, setKeyboardFocusEnabled] = useState(false);
   const [searchSelection, setSearchSelection] = useState('');
   const [searchValue, setSearchValue] = useState('');
@@ -247,8 +252,13 @@ function SelectBox({
       return React.Children.toArray(children)
         .filter((child) => child.type === Option)
         .map((child) => {
-          const { value, children, disabled, icon, ...rest }: OptionProps =
-            child.props;
+          const {
+            value,
+            children,
+            disabled,
+            icon,
+            ...rest
+          }: OptionProps = child.props;
           return {
             value: value,
             label: children || '',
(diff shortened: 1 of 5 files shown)


src/lib/components/selectv2/Selectv2.component.tsx
```diff
@@ -235,6 +235,10 @@ function SelectBox({
   className,
   ...rest
 }: SelectProps) {
+  if (defaultValue & value)
```
JBWatenbergScality:
```suggestion
  if (defaultValue && value)
```

src/lib/components/selectv2/Selectv2.component.tsx
```diff
@@ -235,6 +235,10 @@ function SelectBox({
...
+    console.error(
+      'The `defaultValue` will be overridden by the `value` if they are set at the same time.',
+    );
```
JBWatenbergScality: nit: prefer using a block with condition when using it on multiline
```suggestion
  if (defaultValue && value) {
    console.error(
      'The `defaultValue` will be overridden by the `value` if they are set at the same time.',
    );
  }
```

src/lib/components/selectv2/Selectv2.component.tsx
```diff
@@ -291,6 +300,26 @@ function SelectBox({
     }
   };
 
+  const isEmtryValueInOptions = options.find(
```
JBWatenbergScality:
```suggestion
  const isEmptyValueInOptions = options.find(
```

src/lib/components/selectv2/Selectv2.component.tsx
```diff
@@ -291,6 +300,26 @@ function SelectBox({
...
+    if (
+      !defaultValue &&
+      !isEmtryValueInOptions &&
+      (value === '' || value === null || value === undefined) &&
+      selectRef.current &&
+      selectRef.current.select
+    ) {
+      selectRef.current.select.clearValue();
+    }
+  }, [value, selectRef]);
```
JBWatenbergScality:
```suggestion
  }, [value, selectRef, isEmptyValueInOptions]);
```

src/lib/components/selectv2/Selectv2.component.tsx
```diff
@@ -291,6 +300,26 @@ function SelectBox({
...
+      option.value === null ||
+      option.value === undefined,
+  );
+
+  // Force to reset the value
+  useEffect(() => {
+    if (
+      !defaultValue &&
+      !isEmtryValueInOptions &&
+      (value === '' || value === null || value === undefined) &&
```
JBWatenbergScality:
```suggestion
      (value === '') &&
```

src/lib/components/selectv2/Selectv2.component.tsx
```diff
@@ -291,6 +300,26 @@ function SelectBox({
     }
   };
 
+  const isEmtryValueInOptions = options.find(
+    (option) =>
+      option.value === '' ||
+      option.value === null ||
+      option.value === undefined,
```
JBWatenbergScality:
```suggestion
      option.value === null ||
      option.value === undefined,
```

JBWatenbergScality (approved):

PR #436, opened 2022-05-25
test multi select

MonPote:
**Component**:

**Description**:

**Design**:

**Breaking Changes**:

- [] Breaking Changes

---

Closes: #ISSUE_NUMBER


src/lib/components/tablev2/MultiSelectableContent.tsx
```diff
@@ -0,0 +1,281 @@
...
+
+type MultiSelectableContentProps = {
+  onMultiSelectionChanged: (rows: Row<object>[]) => void;
+  rowHeight?: TableHeightKeyType;
+  separationLineVariant?: TableVariantType;
+  backgroundVariant?: TableVariantType;
+  locale?: TableLocalType;
+  customItemKey?: (index: Number, data: any) => string;
+  children?: (rows: JSX.Element) => JSX.Element;
+};
```
JBWatenbergScality, line 47: I would try to make this type generic so that the row object can actually be typed
```suggestion
type MultiSelectableContentProps<ENTRY extends Record<string, any>> = {
  onMultiSelectionChanged: (rows: Row<ENTRY>[]) => void;
  rowHeight?: TableHeightKeyType;
  separationLineVariant?: TableVariantType;
  backgroundVariant?: TableVariantType;
  locale?: TableLocalType;
  customItemKey?: (index: Number, data: ENTRY) => string;
  children?: (rows: JSX.Element) => JSX.Element;
};
```

JBWatenbergScality: I think locale is a required props when the children props is not passed so the type could be reviewed in the following way:

```suggestion
type MultiSelectableContentProps<ENTRY extends Record<string, any>> = {
  onMultiSelectionChanged: (rows: Row<ENTRY>[]) => void;
  rowHeight?: TableHeightKeyType;
  separationLineVariant?: TableVariantType;
  backgroundVariant?: TableVariantType;
  customItemKey?: (index: Number, data: ENTRY) => string;
} & ({
  locale: TableLocalType;
} | {
  children: (rows: JSX.Element) => JSX.Element;
  });
```

src/lib/components/tablev2/MultiSelectableContent.tsx
```diff
@@ -0,0 +1,281 @@
...
+    setHiddenColumns,
+    selectedRowIds,
+    onBottom,
+    onBottomOffset,
+    isAllRowsSelected,
+  } = useTableContext();
+
+  useEffect(() => {
+    setHiddenColumns([]);
+  }, [setHiddenColumns]);
```
JBWatenbergScality: For now this is ok but maybe later on we will have use cases with hidden columns, So I wonder wether we could not jsut remove the checkbox column from the hidden columns instead of forcing to an empty array here.
Something like this maybe:

```suggestion
  useEffect(() => {
    setHiddenColumns((oldHiddenColumns) => oldHiddenColumns.filter(column => column.id !== 'selection'));
  }, [setHiddenColumns]);
```

src/lib/components/tablev2/MultiSelectableContent.tsx
```diff
@@ -0,0 +1,281 @@
...
+      scrollDiv.setAttribute(
+        'style',
+        'width: 100px; height: 100px; overflow: scroll; position:absolute; top:-9999px;',
+      );
+      node.appendChild(scrollDiv);
+      const scrollbarWidth = scrollDiv.offsetWidth - scrollDiv.clientWidth;
+      node.removeChild(scrollDiv);
+      setScrollBarWidth(scrollbarWidth);
+    }
+  }, []);
```
JBWatenbergScality: I think we could try to mutualize this with the singleSelectableTable. Maybe a custom hook or something

src/lib/components/tablev2/MultiSelectableContent.tsx
```diff
@@ -0,0 +1,281 @@
...
+
+  const RenderRow = memo(({ index, style }: RenderRowType) => {
```
JBWatenbergScality, line 105: Similarly part of this, or maybe even all could be mutualized with the singleselectable

src/lib/components/tablev2/MultiSelectableContent.tsx
```diff
@@ -0,0 +1,281 @@
...
+                          ) : (
+                            <i className="fas fa-sort-up" />
+                          )
+                        ) : (
+                          <SortIncentive>
+                            <i className="fas fa-sort" />
+                          </SortIncentive>
+                        )}
+                      </SortCaretWrapper>
+                    ) : null}
```
JBWatenbergScality: This is also duplicated with singleselectable, we can extract it in a component shared between both tables

src/lib/components/tablev2/Tablev2.component.tsx
```diff
@@ -9,37 +8,64 @@ import {
...
+  selectedFlatRows: Row<object>[];
+  preGlobalFilteredRows: Row<object>[];
+  setGlobalFilter: (filterValue: any) => void;
+  globalFilter: any;
+  setFilter: (columnId: string, updater: any) => void;
+  onBottom?: (rowLength: number) => void;
+  onBottomOffset?: number;
+  setHiddenColumns: (param: string[]) => void;
+  isAllRowsSelected?: boolean;
+};
```
JBWatenbergScality: Same as above try to ban types such as object and any, and use generics !

src/lib/components/tablev2/react-table-config.d.ts
```diff
@@ -0,0 +1,136 @@
...
+  UseSortByColumnOptions,
+  UseSortByColumnProps,
+  UseSortByHooks,
+  UseSortByInstanceProps,
+  UseSortByOptions,
+  UseSortByState,
+} from 'react-table';
+
+// From: https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react-table
+declare module 'react-table' {
```
JBWatenbergScality, line 52: I'm not very sure why we need to re-add this type declaration in core-ui

src/lib/components/tablev2/useCheckbox.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const resolvedRef = ref || defaultRef;
+
+  useEffect(() => {
+    (resolvedRef as MutableRefObject<HTMLInputElement>).current.indeterminate =
+      indeterminate;
+  }, [resolvedRef, indeterminate]);
+
+  return (
+    <>
+      <input type="checkbox" ref={resolvedRef} {...rest} />
```
JBWatenbergScality, line 27: @Cuervino  maybe we need to style a bit this IndeterminateCheckbox

JBWatenbergScality (commented):
A few comments and modifications to be done I think, but for the most part it is good for me. We should implement a test through to ensure that when selecting a row and all the rows the expected callbacks are correctly called, in theory it should not be very difficult if the checkboxes are accessible and I already implemented virtualised table tests in Zenko UI to mock the measurement API, so you can get some inspiration from there.

src/lib/components/tablev2/MultiSelectableContent.tsx
MonPote: For now I will keep it separated. There are some difference. We might be able to factorise it but I don't want to put too much effort on it for now.

src/lib/components/tablev2/MultiSelectableContent.tsx
MonPote: I've encounter some issues tackling this, so I put a comment and created a ticket for that. I will resume this work in a future PR ASAP

src/lib/components/tablev2/useCheckbox.tsx
Cuervino: Indeed, however it should be possible to do it in another iteration. For the moment let's focus on the select.

ChengYanJin (approved):

PR #439, opened 2022-06-10
Bugfix/fix search input

JBWatenbergScality:
**Component**:

SearchInput

**Description**:

Add a forward ref so that the component can be used safely when combined with another one (eg typeahead..)


ChengYanJin (approved):
LGTM!

PR #445, opened 2022-06-24
Added hasScrollbar prop for table v2 and fixed the 'flying' sorting arrow

ognyan-kostadinov:
**Component**: Table V2

**Description**:
Added a `hasScrollbar` prop in table V2 which can be used either to enable or disable the scrolling
Also fixed the "flying" sorting arrow (it was set as position absolute, but didn't have a parent with position relative)
---

Closes: #ISSUE_NUMBER

### DIFF
--- a/src/lib/components/tablev2/SingleSelectableContent.tsx
+++ b/src/lib/components/tablev2/SingleSelectableContent.tsx
@@ -31,6 +31,7 @@ export type SingleSelectableContentProps = {
   locale?: TableLocalType;
   customItemKey?: (index: Number, data: any) => string;
   isLoading?: boolean;
+  hasScrollbar?: boolean;
   children?: (rows: JSX.Element) => JSX.Element;
 };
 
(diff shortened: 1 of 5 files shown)


src/lib/components/tablev2/SingleSelectableContent.tsx
```diff
@@ -56,6 +57,7 @@ export function SingleSelectableContent({
   selectedId,
   onRowSelected,
   customItemKey,
+  hasScrollbar: tableHasScrollbar,
```
JBWatenbergScality, line 60: Maybe we also need the props for MultiSelectableTable as well for a sake of consistency

JBWatenbergScality (commented):
Additionally to my comment below I would advise bumping Core UI version in the same PR to avoid having to open yet another one before releasing it.

```
npm version --no-git-tag 0.39.1
git add package*
git commit
```

JBWatenbergScality (dismissed):

JBWatenbergScality (approved):

PR #446, opened 2022-06-26
table: Fix column alignment

JBWatenbergScality:
**Component**:

table

**Description**:

Scrollbar computation method was incorrect beforehands. This PR fix this and deduce the scroll size from the row width


ChengYanJin (approved):
LGTM!

PR #452, opened 2022-07-08
Fix the tooltip for the disabled button

ChengYanJin:
**Component**: Tooltip

**Description**:

**Disabled elements don't fire mouse events**
So the `onMouseLeave` didn't trigger for the disabled button.
We replaced `onMouseLeave` with `onPointerLeave` event.

---

Closes: #ISSUE_NUMBER


JBWatenbergScality (approved):
GO GO GO

PR #473, opened 2022-08-24
table: fix the wrong url when search in the table

ChengYanJin:
**Component**: table

**Description**:
We shouldn't use useRouteMatch() to retrieve the current pathname.

**Design**:

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/src/lib/components/tablev2/SearchWithQueryParams.tsx
+++ b/src/lib/components/tablev2/SearchWithQueryParams.tsx
@@ -1,5 +1,5 @@
 import { useState } from 'react';
-import { useLocation, useHistory, useRouteMatch } from 'react-router-dom';
+import { useLocation, useHistory } from 'react-router-dom';
 import { TableSearch as Search, SearchProps } from './Search';
 
 export type SearchWithQueryParamsProps = {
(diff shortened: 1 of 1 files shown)


src/lib/components/tablev2/SearchWithQueryParams.tsx
```diff
@@ -18,7 +17,7 @@ export function SearchWithQueryParams(props) {
   function handleOnChange(value) {
     const { onChange } = props;
     params.set(queryParams, value);
-    history.replace(`${match.url}?${params.toString()}`);
+    history.replace(`${pathname}?${params.toString()}`);
```
ognyan-kostadinov, line 20: All looks good, but I'm not familiar with the situation - do we want to add new history entry (thus enabling going back using the browser's back button) or replace it?

ognyan-kostadinov (approved):

src/lib/components/tablev2/SearchWithQueryParams.tsx
ChengYanJin: It's actually a bug. because useRouteMatch( ) attempt to match the current URL in the same way that a <Route> would. So in some case we may not get the entire path.

PR #477, opened 2022-08-30
Fix children JSX.Element -> React.ReactNode in Select

ghivert:
**Component**:

Select

**Description**:

Select was not allowing proper ReactNode, but was requiring JSX.Element. This complicated the code, since it did not allow String, for example.


JBWatenbergScality (approved):
love-stich-gif

PR #480, opened 2022-08-31
Small improvements to Table V2: filtering by individual columns, new event handlers for multiselect table and introducing data-testids for easier testing

ognyan-kostadinov:
**Component**:

**Description**:

**Design**:

**Breaking Changes**:

- [] Breaking Changes

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/src/lib/components/tablev2/MultiSelectableContent.tsx
+++ b/src/lib/components/tablev2/MultiSelectableContent.tsx
@@ -1,6 +1,6 @@
-import { useEffect, memo, CSSProperties, useRef } from 'react';
+import { useEffect, memo, CSSProperties } from 'react';
 import { Row } from 'react-table';
-import { areEqual, FixedSizeList } from 'react-window';
+import { areEqual } from 'react-window';
 
 import { ConstrainedText } from '../constrainedtext/Constrainedtext.component';
 import { Tooltip } from '../tooltip/Tooltip.component';
@@ -41,6 +41,8 @@ type MultiSelectableContentProps<
   DATA_ROW extends Record<string, unknown> = Record<string, unknown>,
 > = {
   onMultiSelectionChanged: (rows: Row<DATA_ROW>[]) => void;
+  onSingleRowSelected?: (row: Row<DATA_ROW>) => void;
+  onToggleAll?: (selected: boolean) => void;
   rowHeight?: TableHeightKeyType;
   separationLineVariant?: TableVariantType;
   backgroundVariant?: TableVariantType;
@@ -70,6 +72,8 @@ export const MultiSelectableContent = <
   DATA_ROW extends Record<string, unknown> = Record<string, unknown>,
 >({
   onMultiSelectionChanged,
+  onSingleRowSelected,
+  onToggleAll,
   rowHeight = 'h40',
   separationLineVariant = 'backgroundLevel3',
   backgroundVariant = 'backgroundLevel1',
@@ -145,6 +149,10 @@ export const MultiSelectableContent = <
             onMultiSelectionChanged(selectedRows);
           }
           row.toggleRowSelected(!row.isSelected);
+
+          if (onSingleRowSelected) {
+            onSingleRowSelected(row);
+          }
         },
       },
     };
@@ -227,7 +235,11 @@ export const MultiSelectableContent = <
                   <div>
                     {column.id === 'selection' ? (
                       <div
-                        onClick={() => {
+                        onClick={(event) => {
+                          const eventTarget = event.target as HTMLInputElement;
+                          if (onToggleAll) {
+                            onToggleAll(eventTarget.checked);
+                          }
                           if (onMultiSelectionChanged) {
                             if (isAllRowsSelected) {
                               onMultiSelectionChanged([]);
--- a/src/lib/components/tablev2/SingleSelectableContent.tsx
+++ b/src/lib/components/tablev2/SingleSelectableContent.tsx
@@ -96,6 +96,7 @@ export function SingleSelectableContent<
       <TableRow
         {...rowProps}
         isSelected={selectedId === row.id}
+        aria-selected={selectedId === row.id ? 'true' : 'false'}
         separationLineVariant={separationLineVariant}
         backgroundVariant={backgroundVariant}
         selectedId={selectedId}
--- a/src/lib/components/tablev2/Tablev2.component.tsx
+++ b/src/lib/components/tablev2/Tablev2.component.tsx
@@ -1,28 +1,29 @@
 import * as React from 'react';
 import { useEffect } from 'react';
 import {
-  useTable,
-  useSortBy,
+  Column,
+  CoreUIColumn,
+  HeaderGroup,
+  Row,
+  SortByFn,
+  TableBodyPropGetter,
+  TableBodyProps,
   useBlockLayout,
-  useRowSelect,
   useExpanded,
   useFilters,
   useGlobalFilter,
-  SortByFn,
-  Row,
-  HeaderGroup,
-  TableBodyPropGetter,
-  TableBodyProps,
-  CoreUIColumn,
-  Column,
+  useRowSelect,
+  useSortBy,
+  useTable,
 } from 'react-table';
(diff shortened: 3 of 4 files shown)


src/lib/components/tablev2/MultiSelectableContent.tsx
```diff
@@ -41,6 +41,8 @@ type MultiSelectableContentProps<
   DATA_ROW extends Record<string, unknown> = Record<string, unknown>,
 > = {
   onMultiSelectionChanged: (rows: Row<DATA_ROW>[]) => void;
+  onSingleRowSelected: (row: Row<DATA_ROW>) => void;
+  onToggleAll: (selected: boolean) => void;
```
JBWatenbergScality: I guess those callbacks should be optional
```suggestion
  onSingleRowSelected?: (row: Row<DATA_ROW>) => void;
  onToggleAll?: (selected: boolean) => void;
```

src/lib/components/tablev2/MultiSelectableContent.tsx
```diff
@@ -145,6 +149,7 @@ export const MultiSelectableContent = <
             onMultiSelectionChanged(selectedRows);
           }
           row.toggleRowSelected(!row.isSelected);
+          onSingleRowSelected(row);
```
JBWatenbergScality:
```suggestion
          if (onSingleRowSelected) {
            onSingleRowSelected(row);
          }
```

src/lib/components/tablev2/MultiSelectableContent.tsx
```diff
@@ -227,7 +232,9 @@ export const MultiSelectableContent = <
                   <div>
                     {column.id === 'selection' ? (
                       <div
-                        onClick={() => {
+                        onClick={(event) => {
+                          const eventTarget = event.target as HTMLInputElement;
+                          onToggleAll(eventTarget.checked);
```
JBWatenbergScality:
```suggestion
                     if (onToggleAll) {
                          onToggleAll(eventTarget.checked);
                     }
```

src/lib/components/tablev2/SingleSelectableContent.tsx
```diff
@@ -154,7 +155,7 @@ export function SingleSelectableContent<
   }
   return (
     <>
-      <div className="thead" role="rowgroup">
+      <div className="thead" role="rowgroup" data-testid="table-head">
```
JBWatenbergScality: Why is it necessary ?

src/lib/components/tablev2/SingleSelectableContent.tsx
```diff
@@ -96,6 +96,7 @@ export function SingleSelectableContent<
       <TableRow
         {...rowProps}
         isSelected={selectedId === row.id}
+        data-testid={selectedId === row.id ? 'row-selected' : undefined}
```
JBWatenbergScality: Why is it necessary ?

src/lib/components/tablev2/Tablev2.component.tsx
```diff
@@ -42,6 +42,7 @@ export type TableProps<
   globalFilter?: string;
   onBottom?: (rowLength: number) => void;
   onBottomOffset?: number;
+  allFilters?: { id: string; value: string }[];
```
JBWatenbergScality, line 46: Can you remind me the context where this is necessary ? Thinking wether we could come up with a better abstraction

src/lib/components/tablev2/Tablev2.component.tsx
```diff
@@ -163,6 +165,14 @@ function Table<
     }
   }, [globalFilter, setGlobalFilter, data]);
 
+  useEffect(() => {
+    if (!allFilters) {
+      setAllFilters([]);
+    } else {
+      setAllFilters(allFilters);
+    }
+  }, [allFilters, setAllFilters]);
```
JBWatenbergScality: We need to warn that allFilters props need to be stable across re-renders, or ensure it's stability within the context of this component

JBWatenbergScality (changes requested):

src/lib/components/tablev2/SingleSelectableContent.tsx
ognyan-kostadinov: I'll check, I think it's not, probably just put it for consistency. Will remove it if it's not used of course

src/lib/components/tablev2/Tablev2.component.tsx
ognyan-kostadinov: The context is filtering by a single (or in that matter multiple) column. Global filter filters by occurrence of a string in all columns, in the case of buckets and endpoints tables, we need to filter by a single column. The prop name and usage is one to one with react-table, my idea was to make it as familiar as possible to people who have used it. And to be able to easily see documentation about its usage

src/lib/components/tablev2/SingleSelectableContent.tsx
ognyan-kostadinov: The previous test about checking for selected rows was checking props. Although it's possible (https://stackoverflow.com/questions/58623666/how-to-test-if-a-component-is-rendered-with-the-right-props-when-using-react-tes  - check out the subcomments of the "approved" answer), it's very roundabout approach, and it defeats the idea of the library - to test the DOM, not props. Selected row has uglified class name, so I needed something to use as selector

JBWatenbergScality: I think the right way for selecting a row would be `getByRole('row')`

ognyan-kostadinov: And how would you know which row is the currently selected(clicked and active) by the user? :)
It's used by the following BucketList tests:

`it('should select row if the bucket name specified in the parameter matches one of the bucket names listed')`
`it('should select no row if there is no bucket name specified in the parameter')`

If you feel they're unnecessary, I can remove them.
I guess another option would be to get the style of the row and check if the background color is different from the other element

ognyan-kostadinov: Although we care if the row is selected and active, not how it's styled

JBWatenbergScality: I think the right way would be to use `aria-selected` https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-selected

ognyan-kostadinov: Thanks, looks perfect for the use-case, wasn't aware of its existence

src/lib/components/tablev2/SingleSelectableContent.tsx
```diff
@@ -182,7 +183,12 @@ export function SingleSelectableContent<
           </HeadRow>
         ))}
       </div>
-      <TableBody role="rowgroup" className="tbody" ref={handleScrollbarWidth}>
+      <TableBody
+        role="rowgroup"
+        className="tbody"
+        ref={handleScrollbarWidth}
+        data-testid="table-body"
```
JBWatenbergScality: I think this can be dropped

src/lib/components/tablev2/Tablev2.component.tsx
```diff
@@ -163,6 +166,22 @@ function Table<
...
+    allFilters,
+    (previous, next) => JSON.stringify(previous) === JSON.stringify(next),
+  );
+
+  useEffect(() => {
+    if (!filters) {
+      setAllFilters([]);
+    } else {
+      const validFilters = filters.filter(
+        (filter) => filter.id && filter.value,
```
JBWatenbergScality: I'm unsure about this `&& filter.value` check. Maybe there are some use cases were we would want to filter on falsy values ?

src/lib/components/tablev2/SingleSelectableContent.tsx
ognyan-kostadinov: The table header has a `div[role="row"]` as well, so I use this to target only table body's rows

JBWatenbergScality (approved):

PR #481, opened 2022-08-31
Remove RegExp in SelectV2

ghivert:
**Component**:
SelectV2

**Description**:
Remove RegExp because it was causing a bug in the match. Replaced with an interspersing.

### DIFF
--- a/src/lib/components/selectv2/Selectv2.component.tsx
+++ b/src/lib/components/selectv2/Selectv2.component.tsx
@@ -1,4 +1,3 @@
-// @ts-nocheck
 import React, {
   createContext,
   useContext,
@@ -163,6 +162,7 @@ const MenuList = (props) => {
     selectedIndex * optionHeight - (ITEMS_PER_SCROLL_WINDOW - 1) * optionHeight;
   useEffect(() => {
     if (listRef && listRef.current) {
+      // @ts-ignore
       listRef.current.scrollTo(
         getScrollOffset(
           listRef.current,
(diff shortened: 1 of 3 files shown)


src/lib/components/selectv2/Selectv2.component.tsx
```diff
@@ -60,7 +60,10 @@ const InternalOption = (props) => {
   const formatOptionLabel = () => {
     const label = props.data.label;
     const inputValue = props.selectProps.inputValue;
-    const parts = label.split(new RegExp(`(${inputValue})`, 'gi'));
+    const parts = label.split(inputValue).flatMap((item, index) => {
+      if (index % 2 === 1) return [inputValue, item]
+      return [item]
+    });
```
JBWatenbergScality: I don't think that this is correct.
For the following input: 'test-bla-test-bla', searching for test 
We expect `['', 'test', '-bla-', 'test', '-bla']` whereas with your implementation we get: `['', '-bla-', 'test', '-bla']`

Maybe simply do a `.flatMap((item) => [item, inputValue])` and remove the last entry

ghivert: I think it's correct, but I can do something simpler. :)

ghivert: Updated. :)

src/lib/components/selectv2/Selectv2.component.tsx
```diff
@@ -212,6 +216,8 @@ export type SelectProps = {
   children?: React.ReactNode;
   defaultValue?: string;
   value?: string;
+  onFocus: () => void;
+  onBlur: () => void;
```
JBWatenbergScality:
```suggestion
  onFocus?: (e: TOBEDEFINEDEVENT) => void;
  onBlur?: (e: TOBEDEFINEDEVENT) => void;
```

JBWatenbergScality (approved):

PR #492, opened 2022-09-13
Implemented optional pre-selected rows

ognyan-kostadinov:
**Component**:

**Description**:

**Design**:

**Breaking Changes**:

- [] Breaking Changes

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/src/lib/components/tablev2/Tablev2.component.tsx
+++ b/src/lib/components/tablev2/Tablev2.component.tsx
@@ -4,6 +4,7 @@ import {
   Column,
   CoreUIColumn,
   HeaderGroup,
+  IdType,
   Row,
   SortByFn,
   TableBodyPropGetter,
@@ -44,6 +45,7 @@ export type TableProps<
   onBottom?: (rowLength: number) => void;
   onBottomOffset?: number;
   allFilters?: { id: string; value: string }[];
+  initiallySelectedRowsIds?: Set<string>;
 };
 
 type setHiddenColumnFuncType = (oldHidden: string[]) => string[];
@@ -97,6 +99,7 @@ function Table<
   allFilters,
   onBottom,
   onBottomOffset = 10,
+  initiallySelectedRowsIds,
 }: TableProps<DATA_ROW>) {
   sortTypes = {
     health: (row1, row2) => {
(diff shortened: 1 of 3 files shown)


src/lib/components/tablev2/Tablev2.component.tsx
```diff
@@ -44,6 +45,7 @@ export type TableProps<
   onBottom?: (rowLength: number) => void;
   onBottomOffset?: number;
   allFilters?: { id: string; value: string }[];
+  initiallySelectedRows?: Set<string>;
```
JBWatenbergScality:
```suggestion
  initiallySelectedRowsIds?: Set<string>;
```

src/lib/components/tablev2/Tablev2.component.tsx
```diff
@@ -117,6 +120,18 @@ function Table<
...
+    if (initiallySelectedRows) {
+      return Array.from(initiallySelectedRows).reduce(
+        (accumulatedValue, currentValue) => (
+          (accumulatedValue[currentValue] = true), accumulatedValue
+        ),
+        {},
+      );
+    }
+    return {};
+  }, [initiallySelectedRows]) as Record<IdType<DATA_ROW>, boolean>;
```
JBWatenbergScality: initiallySelectedRows being an object this  useMemo will get executed on every render.

JBWatenbergScality: I think this type cast could be removed if the reduce function implementation was reviewed as below:

```
Array.from(initiallySelectedRows).reduce(
        (accumulatedValue, currentValue) => (
          ({...accumulatedValue, [currentValue]: true}), accumulatedValue
        ),
        {},
      );
```

ognyan-kostadinov: Unfortunately it's not the case, it still considers it `{}` if not explicitly typecasted

src/lib/components/tablev2/Tablev2.component.tsx
```diff
@@ -117,6 +120,19 @@ function Table<
...
+    if (initiallySelectedRowsIds) {
+      return Array.from(initiallySelectedRowsIds).reduce(
+        (accumulatedValue, currentValue) => ({
+          ...accumulatedValue,
+          [currentValue]: true,
+        }),
+        {},
+      );
+    }
+    return {};
```
JBWatenbergScality, line 133:
```suggestion
  const formattedInitiallySelectedRows = React.useMemo(() => {
    
      return Array.from(initiallySelectedRowsIds || []).reduce(
        (accumulatedValue, currentValue) => ({
          ...accumulatedValue,
          [currentValue]: true,
        }),
        {},
      );
```

JBWatenbergScality (approved):

PR #525, opened 2022-10-14
Feature/layout templates

ognyan-kostadinov:
**Component**:

**Description**:

**Design**:

**Breaking Changes**:

- [] Breaking Changes

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/src/lib/components/layout/v2/AppContainer.tsx
+++ b/src/lib/components/layout/v2/AppContainer.tsx
@@ -1,5 +1,6 @@
-import React, { ReactElement } from 'react';
+import { HTMLProps, ReactElement, ReactNode } from 'react';
 import styled from 'styled-components';
+import { ThemeColors } from '../../../style/theme';
 import { getTheme } from '../../../utils';
 
 const Container = styled.div`
@@ -13,16 +14,24 @@ const FillAvailableFlexBox = styled.div`
 `;
 
 const ContextWrapper = styled.div<{
-  background?: string;
+  background?: ThemeColors;
 }>`
-  background: ${(props) => props.background || 'initial'};
+  background: ${(props) =>
+    props.background ? getTheme(props)[props.background] : 'initial'};
   height: 3rem;
   box-sizing: border-box;
   display: flex;
   align-items: center;
 `;
 
-const ContextContainer = ({ children, background, ...rest }) => (
+const ContextContainer = ({
+  children,
+  background,
+  ...rest
+}: { background: ThemeColors } & Omit<
+  HTMLProps<HTMLDivElement>,
+  'ref' | 'as'
+>) => (
   <ContextWrapper background={background}>
     <FillAvailableFlexBox {...rest}>{children}</FillAvailableFlexBox>
   </ContextWrapper>
@@ -43,18 +52,37 @@ const OverallSummary = ({ children, ...rest }) => (
   </OverallSummaryContainer>
 );
 
-const MainContent = styled.div<{
+const MainContentContainer = styled.div<{
   hasPadding?: boolean;
-  background?: string;
+  background?: ThemeColors;
 }>`
   display: flex;
   flex: 1;
   padding: ${(props) => (props.hasPadding ? '1rem' : 'initial')};
   background: ${(props) =>
-    props.background || getTheme(props)['backgroundLevel3']};
+    getTheme(props)[props.background || 'backgroundLevel3']};
 `;
 
-const MainContentContainer = styled.div<{
+const MainContent = ({
+  children,
+  hasPadding,
+  background,
+  ...rest
+}: {
+  children: ReactNode;
+  hasPadding?: boolean;
+  background?: ThemeColors;
+}) => (
+  <MainContentContainer hasPadding={hasPadding} background={background}>
+    <FillAvailableFlexBox
+      {...rest}
+      style={{ display: 'flex', flexDirection: 'column' }}
+    >
+      {children}
+    </FillAvailableFlexBox>
+  </MainContentContainer>
+);
+const AppChildrenContainer = styled.div<{
   hasPadding?: boolean;
 }>`
   flex: 1;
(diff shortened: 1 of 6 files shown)


src/lib/components/layout/v2/AppContainer.tsx
```diff
@@ -0,0 +1,34 @@
+import { ReactElement } from 'react';
+import styled from 'styled-components';
+
+const Container = styled.div<{
+  hasPadding?: boolean;
+}>`
+  display: flex;
+  flex-direction: row;
+  flex: 1;
+  padding: ${(props) => (props.hasPadding ? '0 1rem' : 'initial')};
```
JBWatenbergScality: I'm not sure the padding should be applied here because the sidebar should stick to the left without padding.
The padding option should likely better be on the MainContentContainer

src/lib/components/layout/v2/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+const Navigation = styled.div`
+  height: ${defaultTheme.navbarHeight};
+`;
+
+export const ContextContainer = styled.div<{
+  background?: string;
+}>`
+  background: ${(props) =>
+    props.background || getTheme(props)['backgroundLevel2']};
```
JBWatenbergScality: Context is not supposed to have any background by default I think. It should take the one from the LayoutContainer

src/lib/components/layout/v2/index.tsx
```diff
@@ -0,0 +1,59 @@
+import { ReactElement } from 'react';
+import styled from 'styled-components';
+import * as defaultTheme from '../../../style/theme';
+import { getTheme } from '../../../utils';
+
+const LayoutContainer = styled.div`
```
JBWatenbergScality, line 6: I think it is missing a background (bgLevel1)

src/lib/components/layout/v2/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+`;
+
+export const OverallSummary = styled.div`
+  background: ${(props) => getTheme(props)['backgroundLevel2']};
+  height: 6rem;
+  box-sizing: border-box;
+  display: flex;
+`;
+
+export const MainContent = styled.div<{
```
JBWatenbergScality: I think it is missing a background level3

src/lib/components/layout/v2/panels.tsx
```diff
@@ -0,0 +1,73 @@
...
+`;
+
+const RightPanel = styled.div<{ hasPadding?: boolean; flex?: number }>`
+  flex: ${(props) => props.flex || '0 auto'};
+  display: flex;
+  padding: ${(props) => (props.hasPadding ? '1rem' : 'initial')};
+`;
+
+export const TwoPanelLayout = ({
+  panelsRatio,
```
JBWatenbergScality, line 70: I would better opt for a constrained `ratio` props that takes those values: `30-70` `50-50` `70-30`. it will be better to strongly enforce the ratios across the application.

src/lib/components/layout/v2/panels.tsx
```diff
@@ -0,0 +1,73 @@
...
+  FiftyFifty,
+};
+
+const PanelsContainer = styled.div<{
+  hasPadding?: boolean;
+}>`
+  display: flex;
+  flexdirection: row;
+  flex: 1;
+  padding: ${(props) => (props.hasPadding ? '1rem' : 'initial')};
```
JBWatenbergScality: I don't think that we want to provide any padding configurability on the PanelsContainer. It should not have any padding

src/lib/index.css
```diff
@@ -1,7 +1,6 @@
 html,
 body {
   font-family: 'Lato';
-  font-size: max(14px, 0.972vw);
```
JBWatenbergScality: Let's keep this line for now (it can have unexpected layout impacts to remove it now)

src/lib/components/layout/v2/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+  display: flex;
+  align-items: center;
+`;
+
+export const OverallSummary = styled.div`
+  background: ${(props) => getTheme(props)['backgroundLevel2']};
+  height: 6rem;
+  box-sizing: border-box;
+  display: flex;
+`;
```
JBWatenbergScality: We should probably vertically center the items

src/lib/components/layout/v2/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+}>`
+  background: ${(props) => props.background || 'initial'};
+  height: 3rem;
+  padding: 1rem;
+  box-sizing: border-box;
+  display: flex;
+  align-items: center;
+`;
+
+export const OverallSummary = styled.div`
```
ChengYanJin: I think we should have a padding on this OverallSummary component.

src/lib/index.ts
```diff
@@ -45,6 +45,14 @@ export { ErrorPage500 } from './components/error-pages/ErrorPage500.component';
...
 export { TextBadge } from './components/textbadge/TextBadge.component';
 export { SpacedBox } from './components/spacedbox/SpacedBox';
+export {
+  Layout as Layout2,
+  ContextContainer,
+  OverallSummary,
+  MainContent,
+} from './components/layout/v2';
+export { TwoPanelLayout } from './components/layout/v2/panels';
+export { AppContainer } from './components/layout/v2/AppContainer';
```
ChengYanJin: In terms of code architect/folder, I suggest we could put all those components in one folder, so when we import from other project, all the component will come from one place.

src/lib/components/layout/v2/index.tsx
JBWatenbergScality: To be precise only left-right padding enforced to 1rem

src/lib/index.ts
ognyan-kostadinov: But they are in one folder currently. What folder structure do you have in mind?

ChengYanJin (approved):
LET'S GO !

PR #532, opened 2022-10-21
Improve dropzone and tooltip

ChengYanJin:
**Component**: Tooltip, dropzone

**Description**:

**Design**:

### DIFF
--- a/src/lib/components/dropzone/Dropzone.tsx
+++ b/src/lib/components/dropzone/Dropzone.tsx
@@ -59,8 +59,16 @@ export function Dropzone({
     useDropzone({
       ...dropzoneOptions,
       onDrop: (acceptedFiles, rejectedFiles, e) => {
+        const { multiple } = dropzoneOptions;
+
         setAcceptedFiles((alreadyAcceptedFiles) => {
-          const newAcceptedFiles = [...alreadyAcceptedFiles, ...acceptedFiles];
+          const singleFileAlreadyHasFile =
+            !multiple && alreadyAcceptedFiles.length === 1;
+
+          const newAcceptedFiles = singleFileAlreadyHasFile
+            ? alreadyAcceptedFiles
+            : [...alreadyAcceptedFiles, ...acceptedFiles];
+
           if (onChange) {
             onChange(newAcceptedFiles);
           }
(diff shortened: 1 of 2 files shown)


src/lib/components/dropzone/Dropzone.tsx
```diff
@@ -59,9 +59,17 @@ export function Dropzone({
...
+
         setAcceptedFiles((alreadyAcceptedFiles) => {
-          const newAcceptedFiles = [...alreadyAcceptedFiles, ...acceptedFiles];
-          if (onChange) {
+          const singleFileAlreadyHasFile =
+            !multiple && alreadyAcceptedFiles.length === 1;
+
+          const newAcceptedFiles = singleFileAlreadyHasFile
+            ? alreadyAcceptedFiles
+            : [...alreadyAcceptedFiles, ...acceptedFiles];
```
JBWatenbergScality, line 70: I'm not sure we that we should change the signature of onChange when using a non multiple dropzone. I think it is fine to keep `onChange(files[])` even if the files array is constrained to a single entry.
The reasoning behind that is multiple:
 - We may imagine an use case where a toggle change a single -selectable dropzone to a multiple one and if the signature of the onChange method is different for both it would complexify the integration of the dropzone
 -  Making this signature change complexifies the typing of the dropzone

src/lib/components/tooltip/Tooltip.component.tsx
```diff
@@ -70,6 +70,8 @@ const TooltipText = styled.div`
   ul,
   ol {
     padding-inline-start: ${spacing.r16};
+    margin-block-start: 0;
```
JBWatenbergScality, line 73: I think we also needed to change:

`padding: ${spacing.r4} ${spacing.r8};` to `padding: ${spacing.r8} ${spacing.r4};`

ChengYanJin: Base on Valamir latest design, the padding-top is 4px, padding-bottom is 8px. So I think it was correct to keep the original value.  We had the confusion of Top is bigger than the Bottom is because there is the `margin-block-start: 10px;`  so by setting 0 to `margin-block-start` should be enough.

JBWatenbergScality (approved):

PR #544, opened 2022-11-09
Disabled built-in HTML validation of forms

ognyan-kostadinov:
**Component**:

**Description**:

**Design**:

**Breaking Changes**:

- [] Breaking Changes

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/src/lib/components/form/Form.component.tsx
+++ b/src/lib/components/form/Form.component.tsx
@@ -264,7 +264,7 @@ const PageForm = forwardRef<HTMLFormElement, PageFormProps>(
     const requireMode = useContext(RequireModeContext);
     return (
       <ScrollbarWrapper>
-        <PageFormWrapper {...formProps} ref={ref}>
+        <PageFormWrapper {...formProps} noValidate ref={ref}>
           <FixedHeader layoutKind="page">
             <PaddedForHeaderAndFooterContent>
               <Wrap>
--- a/src/lib/components/form/Form.component.tsx
+++ b/src/lib/components/form/Form.component.tsx
@@ -22,7 +22,10 @@ import { Text } from '../text/Text.component';
 const DESCRIPTION_PREFIX = 'describe-';
 const maxWidthTooltip = { maxWidth: '20rem' };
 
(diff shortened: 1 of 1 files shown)


src/lib/components/form/Form.component.tsx
```diff
@@ -264,7 +264,7 @@ const PageForm = forwardRef<HTMLFormElement, PageFormProps>(
     const requireMode = useContext(RequireModeContext);
     return (
       <ScrollbarWrapper>
-        <PageFormWrapper {...formProps} ref={ref}>
+        <PageFormWrapper {...formProps} noValidate ref={ref}>
```
JBWatenbergScality, line 270: It should probably also be set on the TabForm. Also as we enforce it at this level, it should not be allowed by the Form props type (aka `Omit<FormProps, 'novalidate'>` is necessary)

ognyan-kostadinov: I'm confused, should I apply this omission to PageForm or PageFormWrapper?

JBWatenbergScality: From PageForm props

src/lib/components/form/Form.component.tsx
```diff
@@ -22,7 +22,10 @@ import { Text } from '../text/Text.component';
 const DESCRIPTION_PREFIX = 'describe-';
 const maxWidthTooltip = { maxWidth: '20rem' };
 
-type FormProps = Omit<HTMLProps<HTMLFormElement>, 'ref' | 'as'> & {
+type FormProps = Omit<
+  HTMLProps<HTMLFormElement>,
+  'ref' | 'as' | 'noValidate' | 'formNoValidate'
```
JBWatenbergScality, line 27: In practice `formNoValidate` is an input property not a form one. But ok for me this is really minor !

JBWatenbergScality (approved):

src/lib/components/form/Form.component.tsx
ognyan-kostadinov: Yes, I saw that, but the IDE was suggesting it on the Form component for some reason, so I've made the decision to remove it in order to not mislead people in the future

PR #547, opened 2022-11-18
table: Enhance default cell renderer with constrained text in column …

JBWatenbergScality:
…definition instead of content rendering. This enables to fix an issue because when the code is minified the defaultRenderer name is changed to an auto-generated tiny equivalent


MonPote (approved):

ChengYanJin (approved):
LGTM!

PR #558, opened 2022-12-13
Improvement/improvements to table

ognyan-kostadinov:
**Component**:

**Description**:

**Design**:

**Breaking Changes**:

- [] Breaking Changes

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/src/lib/components/constrainedtext/Constrainedtext.component.tsx
+++ b/src/lib/components/constrainedtext/Constrainedtext.component.tsx
@@ -25,6 +25,7 @@ const ConstrainedTextContainer = styled.div`
   display: -webkit-box;
   -webkit-line-clamp: ${props.lineClamp};
   -webkit-box-orient: vertical;
+  overflow-wrap: break-word;
   `
       : `word-wrap: break-word;
       white-space: nowrap;`}
@@ -39,7 +40,11 @@ const BlockTooltip = styled.div`
 `;
 
 function isEllipsisActive(element: HTMLDivElement) {
-  return element && element.offsetWidth < element.scrollWidth;
+  return (
+    element &&
+    (element.offsetWidth < element.scrollWidth ||
+      element.offsetHeight < element.scrollHeight)
+  );
 }
 
 function getConstrainedTextContainer(constrainedTextRef, lineClamp, text) {
--- a/src/lib/components/constrainedtext/Constrainedtext.component.tsx
+++ b/src/lib/components/constrainedtext/Constrainedtext.component.tsx
@@ -22,7 +22,6 @@ const ConstrainedTextContainer = styled.div`
   ${(props) =>
     props.lineClamp > 1
       ? `
-  display: -webkit-box;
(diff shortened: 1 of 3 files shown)


src/lib/components/tablev2/SingleSelectableContent.tsx
```diff
@@ -203,6 +207,16 @@ export function SingleSelectableContent<
...
+      {isLoadingMoreItems && (
+        <Box
+          display="flex"
+          justifyContent="center"
+          marginTop={spacing.r16}
+          marginBottom={spacing.r16}
+        >
+          <Loader size="large" />
+        </Box>
+      )}
```
JBWatenbergScality, line 219: This change should also be done in the Multiselectable table

src/lib/components/constrainedtext/Constrainedtext.component.tsx
```diff
@@ -22,7 +22,6 @@ const ConstrainedTextContainer = styled.div`
   ${(props) =>
     props.lineClamp > 1
       ? `
-  display: -webkit-box;
```
JBWatenbergScality, line 25: Can you describe the use case where the ellipsis was not displayed ? According to mdn docs we can't use webkit-line-clamp without display: -webkit-box or -webkit-inline-box (https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp):

> The -webkit-line-clamp CSS property allows limiting of the contents of a [block](https://developer.mozilla.org/en-US/docs/Glossary/Block) to the specified number of lines.

> It only works in combination with the [display](https://developer.mozilla.org/en-US/docs/Web/CSS/display) property set to -webkit-box or -webkit-inline-box and the [-webkit-box-orient](https://developer.mozilla.org/en-US/docs/Web/CSS/box-orient) property set to vertical.

So bottomline I'm afraid that this change breaks the line-clamp "feature" of the component.

ognyan-kostadinov: The ellipsis was not displayed on any column of Groups or Users table (checked both on Firefox and Chrome). I saw the comment and the documentation, but I think it's outdated

ognyan-kostadinov: Here are some "before and after" screenshots: 
before-core-ui-ellipsis-fix

after-core-ui-ellipsis-fix

ognyan-kostadinov: Okay, I understood the nature of the problem. When `display: -webkit-box;` and `-webkit-line-clamp` are used together, ellipsis are placed only when the content exceeds the given number of lines (in our case 2). When the content is less than the given lines, the behavior is similar to `overflow: hidden` (first screenshot)

ognyan-kostadinov: I'll revert this change. It's less than ideal for shorter text, but we avoid breaking the table for longer one

JBWatenbergScality (approved):

PR #601, opened 2023-06-19
add support of width, height and variant for textarea

MonPote:
**Component**:

**Description**:

It add the support for width (and height) vs cols, rows.
It also add variant to handle not “code” style.

**Design**:

**Breaking Changes**:

- [] Breaking Changes

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/src/lib/components/textarea/TextArea.component.tsx
+++ b/src/lib/components/textarea/TextArea.component.tsx
@@ -2,22 +2,42 @@ import React, { forwardRef, HTMLProps } from 'react';
 import styled, { css } from 'styled-components';
 import { spacing } from '../../spacing';
 import { getTheme } from '../../utils';
-type Props = Omit<HTMLProps<HTMLTextAreaElement>, 'as'>;
+
+type TextAreaVariant = 'code' | 'text';
+type Props =
+  | Omit<HTMLProps<HTMLTextAreaElement>, 'as'> & { variant?: TextAreaVariant };
 type RefType = HTMLTextAreaElement | null;
 
-const TextAreaContainer = styled.textarea`
+const TextAreaContainer = styled.textarea<{
+  variant: TextAreaVariant;
+  width?: string | number;
+  height?: string | number;
+}>`
   padding: ${spacing.r12} ${spacing.r8};
   border-radius: 4px;
   resize: vertical;
-  font-family: Courier New;
+  font-family: ${(props) =>
+    props.variant === 'code' ? 'Courier New' : 'Lato'};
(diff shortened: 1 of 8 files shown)


src/lib/components/textarea/TextArea.component.tsx
```diff
@@ -2,22 +2,42 @@ import React, { forwardRef, HTMLProps } from 'react';
...
+
+type TextAreaVariant = 'code' | 'text';
+type Props =
+  | Omit<HTMLProps<HTMLTextAreaElement>, 'as'> & { variant?: TextAreaVariant };
 type RefType = HTMLTextAreaElement | null;
 
-const TextAreaContainer = styled.textarea`
+const TextAreaContainer = styled.textarea<{
+  variant: TextAreaVariant;
+  width?: string | number;
```
ChengYanJin: Just thinking maybe we can use `CSSProperties` which has already typed for us ? 
```suggestion
  width?: CSSProperties['width'];
```

src/lib/components/textarea/TextArea.component.tsx
```diff
@@ -2,22 +2,42 @@ import React, { forwardRef, HTMLProps } from 'react';
...
+  height?: string | number;
```
ChengYanJin:
```suggestion
  height?: CSSProperties['height'];
```

src/lib/components/textarea/TextArea.component.tsx
```diff
@@ -2,22 +2,42 @@ import React, { forwardRef, HTMLProps } from 'react';
 import styled, { css } from 'styled-components';
 import { spacing } from '../../spacing';
 import { getTheme } from '../../utils';
-type Props = Omit<HTMLProps<HTMLTextAreaElement>, 'as'>;
+
+type TextAreaVariant = 'code' | 'text';
+type Props =
+  | Omit<HTMLProps<HTMLTextAreaElement>, 'as'> & { variant?: TextAreaVariant };
```
JBWatenbergScality: Does this type include the width/height props already ?  If yes does their types complies with `CSSProperties['width'];` ?

MonPote: Yep it does and it comply with `CSSProperties['width']`.

src/lib/components/textarea/TextArea.component.tsx
MonPote: Indeed thanks for pointing this, much better !

JBWatenbergScality (approved):

PR #622, opened 2023-08-09
add Binary Prefix option to PrettyBytes

hervedombya:
**Component**:

**Description**:

**Design**:

**Breaking Changes**:

- [] Breaking Changes

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/src/lib/components/prettybytes/PrettyBytes.component.tsx
+++ b/src/lib/components/prettybytes/PrettyBytes.component.tsx
@@ -2,9 +2,10 @@ import prettyBytes from 'pretty-bytes';
 type Props = {
   bytes?: number | null;
   decimals?: number | null;
+  unit?: 'iB' | 'B';
 };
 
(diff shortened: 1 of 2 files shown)


src/lib/components/prettybytes/PrettyBytes.component.tsx
```diff
@@ -2,9 +2,14 @@ import prettyBytes from 'pretty-bytes';
 type Props = {
   bytes?: number | null;
   decimals?: number | null;
+  activeBinaryPrefix?: boolean;
```
JBWatenbergScality: I would prefer using a more developer friendly name, maybe a format prop with 2 possible values (iB - default and B) or base prop with 2 possible values (1000, 1024 - default). I feel like saying binary here expose too much of the choices of the underlying library.

JBWatenbergScality (approved):
LGTM !!

PR #628, opened 2023-09-11
feature: Toast component

hervedombya:
**Component**:
'Toast'

**Description**:
Toast inform users of a process that an app has performed or will perform. They appear temporarily, towards the top of the screen. They shouldn't interrupt the user experience, and they don't require user input to disappear.

**Design**:
Enregistrement de l’écran 2023-09-18 à 16 36 07

**Breaking Changes**:

- [] Breaking Changes

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/src/lib/components/toast/Toast.component.tsx
+++ b/src/lib/components/toast/Toast.component.tsx
@@ -1,0 +1,180 @@
+import { motion } from 'framer-motion';
+import React, { ReactNode, useRef } from 'react';
+import { useTheme } from 'styled-components';
+import { Box } from '../box/Box';
+import { Button } from '../buttonv2/Buttonv2.component';
+import { Icon } from '../icon/Icon.component';
+import { BasicText } from '../text/Text.component';
+import { ToastPosition, positionOutput } from './ToastPositionHelpers';
+import { DurationBasedProgressBar } from './DurationBasedProgressBar';
+import { useToastParameters } from './useToastParameters';
+import styled from 'styled-components';
+
+export type ToastStatus = 'success' | 'error' | 'warning' | 'info';
+
+export type ToastProps = {
+  open: boolean;
+  message: ReactNode;
+  onClose: () => void;
+  status?: ToastStatus;
+  position?: ToastPosition;
+  autoDismiss?: boolean;
+  duration?: number;
+  icon?: React.ReactNode;
+  width?: React.CSSProperties['width'];
+  withProgressBar?: boolean;
+  style?: React.CSSProperties;
+};
+
+export const useGetBackgroundColor = (status: string) => {
+  const theme = useTheme();
+  switch (status) {
+    case 'success':
+      return theme.statusHealthy;
+    case 'error':
+      return theme.statusCritical;
+    case 'warning':
+      return theme.statusWarning;
+    default:
+      return theme.infoPrimary;
+  }
+};
+
+const useGetRgbBackgroundColor = (status: string) => {
+  const theme = useTheme();
+  switch (status) {
+    case 'success':
+      return `rgba(${theme.statusHealthyRGB}, 0.4)`;
+    case 'error':
+      return `rgba(${theme.statusCriticalRGB}, 0.4)`;
+    case 'warning':
+      return `rgba(${theme.statusWarningRGB}, 0.4)`;
+    default:
+      return theme.infoSecondary;
+  }
+};
+
+const defaultIconName = (status: string) => {
+  switch (status) {
+    case 'success':
+      return 'Check-circle';
+    case 'error':
+      return 'Times-circle';
+    case 'warning':
+      return 'Exclamation-circle';
+    default:
+      return 'Info-circle';
+  }
+};
+
+const DefaultIcon = ({ status }: { status: string }) => {
+  const color = useGetBackgroundColor(status);
+  const iconName = defaultIconName(status);
+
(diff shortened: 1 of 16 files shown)


src/lib/components/toast/Toast.component.tsx
```diff
@@ -0,0 +1,180 @@
...
+    default:
+      return 'Info-circle';
+  }
+};
+
+const DefaultIcon = ({ status }: { status: string }) => {
+  const color = useGetBackgroundColor(status);
+  const iconName = defaultIconName(status);
+
+  return <Icon name={iconName} color={color as IconColor} />;
```
JBWatenbergScality: Why do we need to cast the IconColor type here ? Is this type not accurate ? maybe we can update it if so 🤔

src/lib/components/toast/Toast.component.tsx
```diff
@@ -0,0 +1,180 @@
...
+    default:
+      return theme.infoPrimary;
+  }
+};
+
+const useGetRgbBackgroundColor = (status: string) => {
+  const theme = useTheme();
+  switch (status) {
+    case 'success':
+      return 'rgba(10, 173, 166, 0.4)';
```
JBWatenbergScality: nit: I wonder if those rgba shouldn't either be added to the theme either be derived from one of the theme variable

src/lib/components/toast/ToastPositionHelpers.ts
```diff
@@ -0,0 +1,36 @@
...
+  },
+  'top-right': {
+    top: '3rem',
+    right: '1rem',
+  },
+  'top-center': {
+    top: '1rem',
+    left: '50%',
+    transform: 'translateX(-50%)',
+  },
```
JBWatenbergScality, line 24: Does those default 'top' position already take the navbar in account ?

src/lib/components/toast/ToastPositionHelpers.ts
```diff
@@ -0,0 +1,36 @@
+export type ToastPosition =
+  | 'top-left'
+  | 'top-right'
+  | 'top-center'
+  | 'bottom-left'
+  | 'bottom-right'
+  | 'bottom-center';
+
+export const positionOutput: Record<ToastPosition, {}> = {
```
JBWatenbergScality:
```suggestion
export const positionOutput: Record<ToastPosition, CSSProperties> = {
```

src/lib/components/toast/ToastProgressBar.tsx
```diff
@@ -0,0 +1,45 @@
+import { darken } from 'polished';
+import { useEffect, useState } from 'react';
+import { ProgressBar } from '../progressbar/ProgressBar.component';
+
+export function ToastProgressBar({
```
JBWatenbergScality: nit: I wonder wether we shouldn't enlarge the name of this component. It's more like a 'fake' progress bar based on a duration than a Toast specific progress bar
```suggestion
export function DurationBasedProgressBar({
```

src/lib/components/toast/useMutationsHandler.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+      mutation: {
+        isLoading: false,
+        isSuccess: false,
+        isError: false,
+        isIdle: true,
+      },
+      name: 'mutation3',
+      isPrimary: false,
+    },
+  ] as MutationConfig<unknown, unknown>[];
```
JBWatenbergScality: I think the generics should remain on the mutation level, we can have several mutation in this array and each of them with their own variable and return type

src/lib/components/toast/Toast.component.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+  if (!open) {
+    return null;
+  }
+
+  return (
+    <div
+      ref={ref}
+      role="alert"
+      aria-live="polite"
```
JBWatenbergScality: 👍 does the accessibility spec on toast mention the fact that we need to add a signaler for errors ? something like aria-invalid, but probably not that one as it is reserved for fields iirc

src/lib/components/toast/ToastPositionHelpers.ts
hervedombya: yes

src/lib/components/toast/Toast.component.tsx
hervedombya: `IconColor` type is `keyof typeof theme` but actually Typescript considers a variable `theme.statusHealthy` as a string

src/lib/components/toast/Toast.component.tsx
hervedombya: opacity can vary from case to case in my opinion, but the rgb value doesn't, so maybe just add the RGB value?

src/lib/components/toast/Toast.component.tsx
hervedombya: https://docs.google.com/document/d/12HoCYmkDvDMATOOJWdryoSwghHFoisA9OQdiPqCqgvY/edit#heading=h.yx1g2yktvxqs not in documentation, you know where is mentionned ?

src/lib/components/toast/Toast.component.tsx
JBWatenbergScality: Agree !

JBWatenbergScality (approved):

PR #640, opened 2023-11-13
Attach stepperContext to window context

ChengYanJin:
**Component**: Stepper

**Description**:

Attach the Stepper context to window context in this way, so that even though the core-ui bundle is loaded again, we can reuse the existing context.

**Design**:


JBWatenbergScality (approved):
LGTM

PR #654, opened 2023-11-23
Fix type inlining issue for useStepper

JBWatenbergScality:
Those inline types for recursive types seems to get **Mad** with loop of T1_Head something. To avoid this issue I move the types to a reference .d.ts file handled in the source code.

**Component**:

**Description**:

**Design**:

**Breaking Changes**:

- [] Breaking Changes

---

Closes: #ISSUE_NUMBER


hervedombya (approved):
LGTM!

PR #697, opened 2024-02-13
Separator component

hervedombya:
This pull request includes a change in the `src/lib/spacing.tsx` file. The `HSeparator` styled component's background color was updated to use the theme's border color instead of the backgroundLevel1 color.

This change can potentially affect the visual appearance of any UI components that use the `HSeparator` component. It's important to verify that this change doesn't unintentionally alter the look and feel of these components.


JBWatenbergScality (approved):
LGTM !

PR #783, opened 2024-10-09
Feature/accordion component

JeanMarcMilletScality:
**Component**:

**Description**:

**Design**:

**Breaking Changes**:

- [] Breaking Changes

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/src/lib/components/accordion/accordion.component.tsx
+++ b/src/lib/components/accordion/accordion.component.tsx
@@ -12,6 +12,7 @@ export type AccordionProps = {
   title: string;
   id: string;
   children: React.ReactNode;
+  style?: React.CSSProperties;
 };
 
 const AccordionHeader = styled.button`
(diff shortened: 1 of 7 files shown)


src/lib/components/accordion/accordion.component.tsx
```diff
@@ -0,0 +1,96 @@
...
+      const height = containerRef.current.scrollHeight;
+      return height + 'px';
+    } else return '0px';
+  };
+
+  return (
+    <Box style={{ width: '100%' }}>
+      <h3 style={{ margin: 0 }}>
+        <AccordionHeader
+          id={`Accordion header ${id}`}
```
JBWatenbergScality: nit: semantically it's better to not have spaced in HTML IDs.

src/lib/components/accordion/accordion.component.tsx
```diff
@@ -0,0 +1,96 @@
...
+              }}
+            />
+            <Text isEmphazed>{title}</Text>
+          </Stack>
+        </AccordionHeader>
+      </h3>
+
+      <AccordionContainer
+        ref={containerRef}
+        computedHeight={calcContentHeight()}
```
JBWatenbergScality: I wonder if we can simplify the code similarly to this way:
`ref={(element) => element.style.height = element.scrollHeight + 'px'}`

JBWatenbergScality (approved):

PR #790, opened 2024-10-31
Fix bug of Accordion inside Form

JeanMarcMilletScality:
Accordion is not working inside form, it's impossible to use keyboard navigation as enter also trigger form submit


stories/Accordion/accordion.stories.tsx
```diff
@@ -59,7 +61,29 @@ export const Stacked: AccordionStory = {
...
+      <Accordion {...args} />
     </Stack>
   ),
 };
+
+export const WithCustomStyle: AccordionStory = {
+  render: (args) => {
+    const { title } = args;
+    const theme = useTheme();
+    console.log(theme);
```
JBWatenbergScality: nit: this console.log

JBWatenbergScality (approved):

PR #800, opened 2024-11-20
Fix/dropdown component and navbar bottom line

momodaadaa99:
The changes include adding new dependencies and refactoring the dropdown component to use `@floating-ui` so that it fixes the dropdown problem and add automatic position of the dropdown in case of edges.

### Dependencies:
* Added `@floating-ui/react` to the dependencies in `package.json`.

### Dropdown Component Enhancements:
* Refactored `Dropdown.component.tsx` to use `@floating-ui` for positioning and interactions.

### Additional Changes:
* Removed the border-bottom style from `Navbar.component.tsx` to fix the difference in height in navbar when there is a dropdown.

### DIFF
--- a/src/lib/components/dropdown/Dropdown.component.tsx
+++ b/src/lib/components/dropdown/Dropdown.component.tsx
@@ -1,6 +1,5 @@
 // @ts-nocheck
-import { useState, useRef, useCallback, useEffect } from 'react';
-import styled, { css } from 'styled-components';
+import styled from 'styled-components';
 import {
   ButtonStyled,
   ButtonIcon,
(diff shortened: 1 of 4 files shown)


src/lib/components/dropdown/Dropdown.component.tsx
```diff
@@ -109,19 +113,30 @@ function Dropdown({
     items,
     itemToString: (item) => item?.label || '',
   });
+
+  const { refs, floatingStyles } = useFloating({
+    // open: isOpen,
```
JBWatenbergScality: can be deleted

JBWatenbergScality (approved):

PR #809, opened 2024-12-17
add mail icon

JeanMarcMilletScality:
Add new icon to `iconTable` : `Mail` icon with the class `fas faEnvelope`


JeanMarcMilletScality:
Do we still want to add a flag icon?
image

MonPote (approved):

PR #820, opened 2025-03-16
Implement ref API for Select component

ZiyangLinScality:
# Component: Select V2

## Description
Enhanced the `Select V2` component to support ref forwarding, enabling programmatic control of the component. This allows parent components to directly interact with the Select, implementing features like focus management, menu control, and value manipulation.

## Design
- Used React's `forwardRef` and `useImperativeHandle` to expose a controlled API
- Added a strongly typed `SelectRef` interface with the following methods:
  - `focus()`: Sets focus to the select component
  - `blur()`: Removes focus from the select component
  - `openMenu()`: Programmatically opens the dropdown menu
  - `closeMenu()`: Programmatically closes the dropdown menu
  - `setValue(value: string)`: Sets a specific option value
  - `clearValue()`: Clears the currently selected value
  - `select`: Access to the underlying SelectInstance for advanced usage

- Implemented comprehensive test coverage for all exposed methods

## Breaking Changes

- [x] No Breaking Changes

The implementation maintains full backward compatibility with existing usages. Components that don't use refs will continue working exactly as before, while new code can optionally leverage the ref-based API.

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -71,6 +71,7 @@
         "@types/react-dom": "^18.3.1",
         "@types/react-router": "^5.1.20",
         "@types/react-router-dom": "^5.3.3",
+        "@types/react-select": "^4.0.18",
         "@types/react-table": "^7.7.11",
         "@types/react-virtualized-auto-sizer": "^1.0.1",
         "@types/react-window": "^1.8.5",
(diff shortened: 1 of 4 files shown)


JBWatenbergScality (approved):
LGTM

PR #833, opened 2025-07-08
Add time range handling:

JeanMarcMilletScality:
Add bars for every day in time range with value prefilled with 0 
Update tests

Next Step : Add unit Range

### DIFF
--- a/src/lib/components/barchartv2/utils.ts
+++ b/src/lib/components/barchartv2/utils.ts
@@ -1,5 +1,5 @@
-import { BarchartProps } from './Barchart.component';
-import { DAY_MONTH_FORMATER } from '../date/FormattedDateTime';
+import { BarchartProps, TimeType } from './Barchart.component';
+import { DAY_MONTH_FORMATER, TIME_FORMATER } from '../date/FormattedDateTime';
 
 export const getRoundReferenceValue = (value: number): number => {
   if (value <= 0) return 10; // Default for zero or negative values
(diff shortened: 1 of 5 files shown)


src/lib/components/barchartv2/utils.ts
```diff
@@ -36,6 +37,35 @@ export const getMaxValue = (data: { [key: string]: string | number }[]) => {
...
   return Math.max(...values);
 };
+
+/**
+ * Generates all days between start and end timestamps (inclusive)
+ * @param startTimestamp - Start timestamp in milliseconds
+ * @param endTimestamp - End timestamp in milliseconds
+ * @returns Array of timestamps for each day at midnight
+ */
+const generateDayTimestamps = (
```
ChengYanJin: Would be really great if we have unit test to convert this case:

it('should render when the starting data point is missing in the time range', async () => {})

src/lib/components/barchartv2/utils.ts
```diff
@@ -36,6 +37,35 @@ export const getMaxValue = (data: { [key: string]: string | number }[]) => {
...
+): number[] => {
+  const dayTimestamps: number[] = [];
+  const startDate = new Date(startTimestamp);
+  const endDate = new Date(endTimestamp);
+
+  // Set to midnight to ensure consistent day boundaries
+  startDate.setHours(0, 0, 0, 0);
+  endDate.setHours(0, 0, 0, 0);
+
+  const currentDate = new Date(startDate);
```
ChengYanJin: Nit: just currentDate is bit confusing for me.
```suggestion
  const iteratingDate = new Date(startDate);
```

src/lib/components/barchartv2/utils.ts
```diff
@@ -39,61 +39,92 @@ export const getMaxValue = (data: { [key: string]: string | number }[]) => {
...
+      TIME_FORMATER.format(date)
+    );
+  } else if (interval === 24 * 60 * 60 * 1000) {
+    // Daily or longer intervals - use day format
+    return DAY_MONTH_FORMATER.format(date).replace(/[ ,]/g, '');
+  } else if (interval >= 60 * 60 * 1000) {
+    // Hourly intervals - use hour format
+    const hours = TIME_FORMATER.format(date);
+    const day = DAY_MONTH_FORMATER.format(date).replace(/[ ,]/g, '');
+    return `${day} ${hours}`;
```
ChengYanJin: For hourly intervals and minute intervals, maybe we could display the label only with hours. 
otherwise I feel too many repetition for day .

ChengYanJin (approved):

PR #835, opened 2025-07-10
ARTESCA-15250 Handle stacked bars

JeanMarcMilletScality:
Add logic to sum values for stacked bars to get maxValue 
Add test and story

Next Step : Add a sorting callback function

### DIFF
--- a/src/lib/components/barchartv2/utils.ts
+++ b/src/lib/components/barchartv2/utils.ts
@@ -228,26 +228,8 @@ export const formatPrometheusDataToChartData = (
   // Convert map to array (order is preserved for time ranges)
   const data = Array.from(categoryMap.values());
 
-  // Sort bars by their average values in descending order when stacked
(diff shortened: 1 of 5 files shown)


src/lib/components/barchartv2/utils.ts
```diff
@@ -211,6 +228,27 @@ export const formatPrometheusDataToChartData = (
...
+      const values = data
+        .map((item) => Number(item[bar.dataKey]) || 0)
+        .filter((value) => !isNaN(value));
+      const average =
+        values.length > 0
+          ? values.reduce((a, b) => a + b, 0) / values.length
+          : 0;
+      return { ...bar, average };
+    });
+
```
ChengYanJin: I think this could be extracted to a helper/ util function.  ?

ChengYanJin (approved):

PR #836, opened 2025-07-17
Add sorting function and typing

JeanMarcMilletScality:
Add sorting callback to Barchart

### DIFF
--- a/src/lib/components/barchartv2/Barchart.component.tsx
+++ b/src/lib/components/barchartv2/Barchart.component.tsx
@@ -25,24 +25,28 @@ export type TimeType = {
     interval: number;
   };
 };
-type Point = {
+export type Point = {
   key: string | number;
   values: { label: string; value: number }[];
 };
 
-export type BarchartProps = {
+export type BarchartProps<
+  T extends readonly {
+    readonly label: string;
+    readonly data: readonly (readonly [number | string, number | string])[];
+    readonly color: string;
+  }[],
+> = {
   type: 'category' | TimeType;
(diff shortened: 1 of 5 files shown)


src/lib/components/barchartv2/Barchart.component.tsx
```diff
@@ -25,24 +25,28 @@ export type TimeType = {
...
   values: { label: string; value: number }[];
 };
 
-export type BarchartProps = {
+export type BarchartProps<
+  T extends readonly {
+    readonly label: string;
+    readonly data: readonly (readonly [number | string, number | string])[];
+    readonly color: string;
+  }[],
```
JBWatenbergScality: I'm unsure about this typing. I think the array needs to be converted to a tuple, but maybe readonly does that automatically now, do you have the sort typing as expected ?

src/lib/components/barchartv2/utils.ts
```diff
@@ -135,21 +136,23 @@ const findRangeForTimestamp = (
...
  */
-export const formatPrometheusDataToChartData = (
-  bars: BarchartProps['bars'],
-  type: BarchartProps['type'],
+export const formatPrometheusDataToChartData = <
+  T extends readonly {
+    readonly label: string;
+    readonly data: readonly (readonly [number | string, number | string])[];
+    readonly color: string;
+  }[],
```
JBWatenbergScality: the type shouldn't be copied/pasted, export it from the other file

JBWatenbergScality (approved):

PR #839, opened 2025-07-21
Feature/artesca 15250 title props

JeanMarcMilletScality:
Add loading
Add title props
Add colorSet props

### DIFF
--- a/src/lib/components/barchartv2/Barchart.component.tsx
+++ b/src/lib/components/barchartv2/Barchart.component.tsx
@@ -12,17 +12,20 @@ import {
 import styled, { useTheme } from 'styled-components';
 import { spacing, Stack, Wrap } from '../../spacing';
 import { ConstrainedText } from '../constrainedtext/Constrainedtext.component';
-import {
-  computeUnitLabelAndRoundReferenceValue,
-  formatPrometheusDataToChartData,
-  getMaxBarValue,
-  renderTooltipContent,
-  UnitRange,
-} from './utils';
+import { renderTooltipContent, useChartData, UnitRange } from './utils';
 import { Text } from '../text/Text.component';
 import { IconHelp } from '../iconhelper/IconHelper';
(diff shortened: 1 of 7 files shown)


JeanMarcMilletScality:
I will do another PR  separating color adding in its own fn and refactoring the main formatting fn as it is becoming too big and not easily readable

src/lib/components/barchartv2/Barchart.component.tsx
```diff
@@ -204,11 +205,10 @@ export const Barchart = <T extends BarchartBars>(props: BarchartProps<T>) => {
     helpTooltip,
     rightTitle,
     isLoading,
-    colorSet,
   } = props;
 
   const { rechartsBars, unitLabel, roundReferenceValue, rechartsData } =
-    useChartData(bars, type, stacked, defaultSort, unitRange, colorSet);
+    useChartData(bars, type, colorSet, stacked, defaultSort, unitRange);
```
ChengYanJin, line 211: oops,

src/lib/style/theme.ts
```diff
@@ -162,6 +171,20 @@ export const lineColor5 = '#E3FF73';
 export const lineColor6 = '#BE2543';
 export const lineColor7 = '#FD8144';
 export const lineColor8 = '#F6B187';
+
+export type ChartColors = keyof typeof chartColors;
+
+export const chartColors = {
+  lineColor1: '#A14FBF',
```
ChengYanJin: Nit: we can directly export const chartColors = { lineColor1, lineColor2, ....} since we have already define it before.

src/lib/components/barchartv2/Barchart.component.tsx
ChengYanJin: We need add `colorSet` to all the stories to avoid the type error.

ChengYanJin (approved):
LGTM!!!

PR #840, opened 2025-07-23
Split formatPrometheusDataToRechartsDataAndBars logic

JeanMarcMilletScality:
split logic for better readability and maintenance
Update tests and stories


ChengYanJin (approved):
LGTM !!

ChengYanJin:
Just a small detail of UX, do you think the background color can be added only in between the ReferenceLine and X-axis ?

PR #850, opened 2025-08-05
fix toast provider unnecessary re-renders

damiengillesscality:
**Component**:

Toast

**Description**:

Fix toast provider unnecessary re-renders.

**Design**:

Toast was missing a memoization on the Provider value. They should always be memoized. Without that every time the ToastProvider is re-rendered the context value reference changes (and this can happen a lot if it's parent is a QueryClientProvider for example). This means that the components calling useToast would re-render for completely unrelated reasons.

I did not provide the `setToastProps` dependency since react provides guaranties that it's ref is stable and the project don't enforce exhaustive deps.

**Breaking Changes**:

- None


ChengYanJin (approved):
LGTM!

PR #862, opened 2025-08-13
Refactor icon loading logic to remove react-query

damiengillesscality:
**Component**:

Icon

**Description**:

This is a code refactor of Icon component to remove it's dependency on react-query.

**Design**:

We all want to upgrade react query but it's a scary upgrade as it's a cornerstone of our applications. This change will make it a lot easier.

Few components actually use react query and they are themself used by very few projects, but Icon is the exception it's used a bit everywhere and oddly uses react query.

If we get rid of react-query here the our other project that does not rely on `InlineInput`, `useMutationHandler`, `AttachmentConfirmationModal` or `AttachmentTable` wont have to provide a react-queryV3 provider anymore and will be free to upgrade to whatever version of `@tanstack/react-query`.

By replacing react-queryV3 with this component implementation we are only loosing the react-query retry and cache policies eventually set in the provider of the client. I'd call that a bugfix, and say the user's browser cache policy should decide how long to cache the image. Also, the default options like refetchOnWindowFocus caused the recreation of the intermediate component `default` and engendered many redraw that will be fixed with this change.

Overall I made the minimal change to remove react-query and the related issues, it should render exactly the same but less often and faster.

**Breaking Changes**:

- Icon: `onClick?: (event: MouseEvent) => void` -> `onClick?: (event: React.MouseEvent) => void`
  - The good type was always React.MouseEvent but due to many implicit any ts did not spot it before. The types are very similar but if the client typed is handler as MouseEvent ts will cry and it will need to be changed to the appropriate type.

### DIFF
--- a/src/lib/components/icon/Icon.component.tsx
+++ b/src/lib/components/icon/Icon.component.tsx
@@ -7,11 +7,13 @@ import {
   useEffect,
   useState,
 } from 'react';
-import { useQuery } from 'react-query';
 import styled, { css } from 'styled-components';
 import { CoreUITheme } from '../../style/theme';
 import { Loader } from '../loader/Loader.component';
 import { RemoteGroup, RemoteUser } from './CustomsIcons';
+
+// Module-level cache for imported icons
+const iconCache: Record<string, any> = {};
 
 export const iconTable = {
   Account: 'fas faWallet',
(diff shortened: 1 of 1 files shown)


ChengYanJin (approved):

MonPote (approved):

src/lib/components/icon/Icon.component.tsx
```diff
@@ -226,71 +228,57 @@ export const IconWrapper = styled.div<{ size: SizeProp }>`
...
-          color={color}
-          size={size}
-          ariaLabel={ariaLabel}
-          title={title}
-          {...rest}
-        />
-      )}
-    </>
+    // Handle FontAwesome icons with dynamic import
+    import(`@fortawesome/${fontAwesomeType}/${iconClass}.js`)
```
JBWatenbergScality, line 256: I think that we are missing some unmount cleaning here that was handled by react query. If the component is unmounted we have the promise leaking and the state of an unpresent anymore component updated

PR #869, opened 2025-08-19
Improvement/barchart state improvements

JeanMarcMilletScality:
Add State handling for Barchart:
Loading state and Error state
Handle undefined bars
Update tests

Add export for ChartLegend

### DIFF
--- a/src/lib/components/barchartv2/utils.ts
+++ b/src/lib/components/barchartv2/utils.ts
@@ -28,6 +28,11 @@ export const getMaxBarValue = (
   data: { [key: string]: string | number }[],
   stacked?: boolean,
 ) => {
+  // Handle empty data gracefully
+  if (data.length === 0) {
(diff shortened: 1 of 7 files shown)


src/lib/components/barchartv2/utils.ts
```diff
@@ -18,16 +18,25 @@ export const getRoundReferenceValue = (value: number): number => {
...
 };
 
 export const getMaxBarValue = (
   data: { [key: string]: string | number }[],
   stacked?: boolean,
 ) => {
+  // Handle empty data gracefully
+  if (data.length === 0) {
+    return 0;
+  }
```
ChengYanJin: With the change of `Math.max(...values, 0)` its already handle this case. I think we can remove it.

ChengYanJin (approved):

PR #875, opened 2025-08-21
Format date in tick formatter to have timestamp to manipulate for too…

JeanMarcMilletScality:
date format for chart is not really convenient to handle so format tick in tick component


ChengYanJin (approved):
LGTM!

PR #877, opened 2025-08-25
Improvement/chart tooltip

JeanMarcMilletScality:
Main change:
- Add a tooltip to use in Barchart

Other Changes:
- Update ConstrainedText component to allow color def (avoid color mismatch between suspension point and text)
- Add format to FormattedDateTime component
- Update BarChart Component


ChengYanJin (approved):
LGTM!

PR #882, opened 2025-08-26
Removed motion used in toaster

damiengillesscality:
**Component**:

Toaster

**Description**:

Removed react motion from the implementation to have it compatible with react19.

**Design**:

Behaviour is unchanged. I did not implement ToastExit animation because it was not working in the first place with motion as the previous implementation forgot to use `AnimatePresence`.

**Breaking Changes**:

- None


hervedombya (approved):
LGTM

PR #884, opened 2025-08-27
Multi select issue in ChartLegend

JeanMarcMilletScality:
Previous behavior : when multiple legend items but not all are selected, clicking on a selected item select all

New behavior: when multiple legend items but not all are selected, clicking on a selected item select only this item

### DIFF
--- a/src/lib/components/chartlegend/ChartLegend.tsx
+++ b/src/lib/components/chartlegend/ChartLegend.tsx
@@ -73,8 +73,8 @@ export const ChartLegend = ({
     addSelectedResource,
     removeSelectedResource,
     selectAllResources,
-    getSelectedCount,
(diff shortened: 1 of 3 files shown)


src/lib/components/chartlegend/ChartLegend.tsx
```diff
@@ -97,14 +96,10 @@ export const ChartLegend = ({
         } else {
           addSelectedResource(resource);
         }
+      } else if (itemIsSelected && getSelectedCount() === 1) {
```
ChengYanJin: Nit: Maybe we can change `getSelectedCount()` to `isOnlyoneSelected ()` ?

ChengYanJin (approved):
LGTM!

PR #885, opened 2025-08-27
Support irregular interval dataset

ChengYanJin:
**Component**:

**Description**:

**Design**:

**Breaking Changes**:

- [] Breaking Changes

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/src/lib/components/linetemporalchart/ChartUtil.test.ts
+++ b/src/lib/components/linetemporalchart/ChartUtil.test.ts
@@ -165,7 +165,7 @@ it('returns the unit label GiB/Sec', () => {
   expect(valueBase).toEqual(1024 * 1024 * 1024);
 });
 // test for addMissingDataPoint function
-const originalValue: [number, string | null][] = [
+const originalValue: [number, number | string | null][] = [
   [0, '0'],
   [1, '1'],
   [2, '2'],
--- a/src/lib/components/linetemporalchart/ChartUtil.test.ts
+++ b/src/lib/components/linetemporalchart/ChartUtil.test.ts
@@ -165,33 +165,33 @@ it('returns the unit label GiB/Sec', () => {
   expect(valueBase).toEqual(1024 * 1024 * 1024);
 });
 // test for addMissingDataPoint function
-const originalValue = [
-  [0, 0],
-  [1, 1],
-  [2, 2],
-  [3, 3],
-  [4, 4],
-  [5, 5],
-  [6, 6],
-  [8, 8],
-  [9, 9],
-  [10, 10],
+const originalValue: [number, string | null][] = [
+  [0, '0'],
+  [1, '1'],
+  [2, '2'],
+  [3, '3'],
+  [4, '4'],
+  [5, '5'],
+  [6, '6'],
+  [8, '8'],
+  [9, '9'],
+  [10, '10'],
 ];
 const startingTimeStamp = 0;
 const sampleDuration = 10;
 const sampleFrequency = 1;
 const newValues = [
-  [0, 0],
-  [1, 1],
-  [2, 2],
-  [3, 3],
-  [4, 4],
-  [5, 5],
-  [6, 6],
+  [0, '0'],
+  [1, '1'],
+  [2, '2'],
+  [3, '3'],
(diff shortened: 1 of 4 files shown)


src/lib/components/linetemporalchart/ChartUtil.ts
```diff
@@ -134,27 +135,37 @@ export function addMissingDataPoint(
...
-  if (newValues.length === 0) return [];
-  let nextIndex = 0;
+  const newValues: [number, string | null][] = [];
+
+  for (let i = 0; i < orginalValues.length; i++) {
+    // Always add the current data point
+    newValues.push(orginalValues[i]);
+
+    // Check if this is not the last data point
+    if (i < orginalValues.length - 1) {
```
damiengillesscality: I'd recomment looping from 0 to orginalValues.length-1 to avoid this check, at after the loop you can simply push the last value.

src/lib/components/linetemporalchart/ChartUtil.ts
```diff
@@ -134,27 +135,37 @@ export function addMissingDataPoint(
...
+      const currentTimestamp = orginalValues[i][0];
+      const nextTimestamp = orginalValues[i + 1][0];
+      const gap = nextTimestamp - currentTimestamp;
+
+      // Only add missing data points if gap is bigger than 2 intervals
+      if (gap >= 2 * sampleFrequency) {
```
damiengillesscality: You do not need this check, it's repetitive with your for loop condition j <= missingIntervals

src/lib/components/linetemporalchart/ChartUtil.ts
```diff
@@ -134,27 +135,37 @@ export function addMissingDataPoint(
...
-      orginalValues[nextIndex] &&
-      newValues[i][0] === orginalValues[nextIndex][0]
-    ) {
-      newValues[i][1] = orginalValues[nextIndex][1];
-      nextIndex++;
+        // Add missing data points with NAN_STRING
+        for (let j = 1; j <= missingIntervals; j++) {
+          const missingTimestamp = currentTimestamp + j * sampleFrequency;
+          // Only add if the missing timestamp is before the next actual data point
+          if (missingTimestamp < nextTimestamp) {
```
damiengillesscality: That condition should never be falsy, if it is sometime we should fix the logic of the missing interval loop to do the good calculation

src/lib/components/linetemporalchart/ChartUtil.test.ts
```diff
@@ -256,7 +244,7 @@ it('should return an empty array when sample frequency is undefined', () => {
     originalValue,
     startingTimeStamp,
     sampleDuration,
-    undefined,
+    undefined as any,
```
damiengillesscality: We should not have to lie about the type, addMissingDataPoint sampleDuration should probably be optional mark it with a `?` in the function signature.

damiengillesscality:
Could we please rename sampleFrequency -> samplePeriod to beter reflect the expected unit (time and not 1/time)

damiengillesscality:
Why is the expected data type `[number, string | null][]` and not `[number, number | null][]` ?

ChengYanJin:
> Why is the expected data type `[number, string | null][]` and not `[number, number | null][]` ?

Hi, it is because the value returned by Prometheus API is string.

damiengillesscality:
> > Why is the expected data type `[number, string | null][]` and not `[number, number | null][]` ?
> 
> Hi, it is because the value returned by Prometheus API is string.

I see perharps we could accept both number and string ? It would be odd to receive numbers convert then on string for core ui and have the graph lib taking numbers ultimately ?

ChengYanJin:
> > > Why is the expected data type `[number, string | null][]` and not `[number, number | null][]` ?
> > 
> > 
> > Hi, it is because the value returned by Prometheus API is string.
> 
> I see perharps we could accept both number and string ? It would be odd to receive numbers convert then on string for core ui and have the graph lib taking numbers ultimately ?

Yes. Make perfectly sense ! I will make the change to support both string and number.

damiengillesscality (approved):

PR #880, opened 2025-08-26
Deprecate component

JeanMarcMilletScality:
Removal of deprecate components and their stories

Next Steps : 
- Refactor Dropdown component to remove Button v1
- GlobalHealthBar with Recharts

PR #886, opened 2025-08-28
exported CoreUITheme and updated readme

damiengillesscality:
**Component**:

- None

**Description**:

I wished to bring some updates to the readme to better reflect the dependencies you actually need to use core-ui and to clean a few formulations and syntax stuff.

Doing so I noticed the readme encourages to use `CoreUITheme` type but it is actually never exported by the lib. So I decided to export it and discussed with @ChengYanJin @JeanMarcMilletScality to n know if I should put it in index or next.

We concluded that we probably wants to put everything that is not deprecated in next. But reading the readme it says otherwise, I also noticed @JBWatenbergScality that your latest components are exported in index.

So should I export `CoreUITheme` in index or should I correct the readme ?

**Design**:

**Breaking Changes**:

- None

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,38 +11,21 @@ Core-UI is a component library containing all components, layouts, icons and the
 - Add `@scality/core-ui` in the `package.json`'s dependencies of your project.
 
 ```json
-    "@scality/core-ui": "0.115.0",
+    "@scality/core-ui": "^0.165.0",
 ```
 
 - `@scality/core-ui` requires the peerDependencies below. Make sure that you have them in the `package.json`'s dependencies.
 
 ```json
-    "@fortawesome/fontawesome-free": "^5.10.2",
-    "@fortawesome/fontawesome-svg-core": "^1.2.35",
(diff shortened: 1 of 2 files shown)


JBWatenbergScality:
I think you can export in both. The strategy would be to make a v1 of Core UI one day with a single entry point (being the index one) and drop all deprecated things and vega once for all 😄 .

JeanMarcMilletScality:
> I think you can export in both. The strategy would be to make a v1 of Core UI one day with a single entry point (being the index one) and drop all deprecated things and vega once for all 😄 .

Soon 🤞 
You can look at my PR [deprecate components](https://github.com/scality/core-ui/pull/880), most of deprecated components are not use anywhere. The biggest issue is Vega. But with the new charts the only missing step to remove it is to have GlobalHealthBar with Recharts

JBWatenbergScality (approved):

damiengillesscality:
Wrong branch name, closed in favor of https://github.com/scality/core-ui/pull/887

PR #903, opened 2025-09-03
improve line timeserie chart - xaxis tick label

ChengYanJin:
**Component**: LineTemporalChart

**Description**:
For the X-axis label, If time range is greater than 1 year, use YYYY-MM-DD format. Otherwise, use MM-DD format.

**Design**:

**Breaking Changes**:

- [] Breaking Changes

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/src/lib/components/linetimeseriechart/linetimeseriechart.component.tsx
+++ b/src/lib/components/linetimeseriechart/linetimeseriechart.component.tsx
@@ -8,6 +8,7 @@ import {
   YAxis,
   CartesianGrid,
 } from 'recharts';
+import type { Payload } from 'recharts/types/component/DefaultTooltipContent';
 import { useCallback, useMemo, useRef } from 'react';
 import { useTheme } from 'styled-components';
 import { addMissingDataPoint } from '../linetemporalchart/ChartUtil';
@@ -23,6 +24,12 @@ import { Tooltip as TooltipComponent } from '../tooltip/Tooltip.component';
 import { FormattedDateTime } from '../date/FormattedDateTime';
 import { Box } from '../box/Box';
 import { formatXAxisLabel } from './utils';
+
+type TooltipPayload = Payload<number, string> & {
+  value: number;
+  name: string;
+  color: string;
+};
 
 const LineTemporalChartWrapper = styled.div`
   display: flex;
@@ -150,12 +157,7 @@ const CustomTooltip = ({
   timeFormat,
 }: {
   active?: boolean;
-  payload?: Array<{
-    value: number;
-    name: string;
-    color: string;
-    dataKey: string;
-  }>;
+  payload?: Array<TooltipPayload>;
   label?: string;
   unitLabel?: string;
   timeFormat?: 'date-time' | 'date';
@@ -164,8 +166,8 @@ const CustomTooltip = ({
   // We can't use the default itemSorter method because it's a custom tooltip.
   // Sort the payload here instead
   const sortedPayload = [...payload].sort((a, b) => {
-    const aValue = Number(a.value);
-    const bValue = Number(b.value);
+    const aValue = a.value;
+    const bValue = b.value;
 
     if (aValue >= 0 && bValue >= 0) {
       return bValue - aValue; // Higher positive values first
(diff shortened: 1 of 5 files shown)


src/lib/components/linetimeseriechart/utils.ts
```diff
@@ -0,0 +1,48 @@
...
+  if (timeFormat === 'date-time') {
+    return DAY_MONTH_ABBREVIATED_HOUR_MINUTE.format(date).replace(',', '');
+  } else if (timeFormat === 'date') {
+    // Calculate the time range to determine format
+    if (chartData.length > 0) {
+      const timestamps = chartData.map((d) => d.timestamp);
+      const minTimestamp = Math.min(...timestamps);
+      const maxTimestamp = Math.max(...timestamps);
+      const timeRangeMilliseconds = maxTimestamp - minTimestamp;
+      const oneYearMilliseconds = 366 * 24 * 60 * 60 * 1000;
```
damiengillesscality: This should be on top of the file in UPPER_SNAKE_CASE

src/lib/components/linetimeseriechart/utils.ts
```diff
@@ -0,0 +1,48 @@
...
+  timeFormat: 'date-time' | 'date' = 'date-time',
+  chartData: ChartDataPoint[] = [],
+): string => {
+  const date = new Date(timestamp);
+
+  if (timeFormat === 'date-time') {
+    return DAY_MONTH_ABBREVIATED_HOUR_MINUTE.format(date).replace(',', '');
+  } else if (timeFormat === 'date') {
+    // Calculate the time range to determine format
+    if (chartData.length > 0) {
```
damiengillesscality: Please prefer early returns to avoid indentation

if (!chartData.length) {
  return ...
}

src/lib/components/linetimeseriechart/utils.ts
```diff
@@ -0,0 +1,48 @@
...
+export const formatXAxisLabel = (
+  timestamp: number,
+  timeFormat: 'date-time' | 'date' = 'date-time',
+  chartData: ChartDataPoint[] = [],
+): string => {
+  const date = new Date(timestamp);
+
+  if (timeFormat === 'date-time') {
+    return DAY_MONTH_ABBREVIATED_HOUR_MINUTE.format(date).replace(',', '');
+  } else if (timeFormat === 'date') {
```
damiengillesscality: unecessary `else if` statement it can never be false

src/lib/components/linetimeseriechart/linetimeseriechart.component.tsx
```diff
@@ -181,22 +183,22 @@ const CustomTooltip = ({
...
             <TooltipName>{entry.name}</TooltipName>
-            <TooltipInstanceValue>
-              {isNaN(Number(entry.value))
-                ? '-'
-                : `${Number(entry.value).toFixed(2)}${unitLabel}`}
-            </TooltipInstanceValue>
-          </TooltipContent>
+          </TooltipLeftGroup>
+          <TooltipInstanceValue>
+            {isNaN(Number(entry.value))
```
damiengillesscality: You should probably be using `Number.isFinite` instead it is truthy only when the param is of type number not NaN not infinite. The check is better and faster

src/lib/components/linetimeseriechart/linetimeseriechart.component.tsx
```diff
@@ -181,22 +183,22 @@ const CustomTooltip = ({
...
+              ? '-'
+              : `${Number(entry.value).toFixed(2)} ${unitLabel}`}
```
damiengillesscality: `entry.value` is already garanted to be number don't rewrap with `Number(` maybe you need better typing hints for `sortedPayload.map`

src/lib/components/linetimeseriechart/utils.ts
damiengillesscality: I mean not only the if is unecessary but the else to. We always return when the condition is true and execute everything after the if block when the condition is false.

damiengillesscality (approved):

PR #904, opened 2025-09-08
Global health bar with recharts

JeanMarcMilletScality:
Reworks Global health bar using Recharts lib
Tooltip:
- Use Portal for tooltip to avoid clipping issue when GlobalHealthBar is in a container

Label: 
- Add calculation and conditions for ticks labels to have consistent spacing between labels and have same behavior as previous chart
- Add margin to chart to avoid label clipping

Next Step: Add round edges to bars when bars is at start or end of chart (use mask?)


src/lib/components/globalhealthbar/components/AlertBar.tsx
```diff
@@ -0,0 +1,44 @@
...
+    onPointerLeave={onPointerLeave}
+    isAnimationActive={false}
+  />
+);
+
+export const createAlertBarRenderer = (
+  startTimestamp: number,
+  endTimestamp: number,
+) => {
+  return (props: any, key: string) => {
```
JBWatenbergScality:
```suggestion
  return (props: SHOULD_BE_RECT_PROPS, key: string) => {
```

src/lib/components/globalhealthbar/components/GlobalHealthBarTooltip.tsx
```diff
@@ -0,0 +1,135 @@
...
+      }
+
+      // Final safety check to ensure tooltip is never off-screen
+      if (top < MARGIN) {
+        top = MARGIN;
+      }
+
+      setTooltipPosition({ left, top });
+    }
+  }, [coordinate, chartContainerRef]);
```
JBWatenbergScality: nit(for later): I feel like the whole logic of the effect also needs to be executed on scroll and window resize 🤔

JBWatenbergScality: Probably some of this can be computed using utils from floating-ui instead of re-inventing the wheel 🤔

src/lib/components/globalhealthbar/components/HealthBarXAxis.tsx
```diff
@@ -0,0 +1,94 @@
...
+          {is7DaySpan || isDaySpan ? (
+            <FormattedDateTime
+              format="day-month-abbreviated-hour-minute"
+              value={new Date(payload.value)}
+            />
+          ) : (
+            <FormattedDateTime format="time" value={new Date(payload.value)} />
+          )}
+        </text>
+      </g>
```
JBWatenbergScality, line 59: Isn't this the same logic for linetemporal chart and temporal bar charts ?

src/lib/components/globalhealthbar/GlobalHealthBarRecharts.component.tsx
```diff
@@ -0,0 +1,142 @@
...
+  const handlePointerEnter = useCallback(
+    (key: string) => {
+      setTooltipData(data[0][`alert_${key}`]);
+    },
+    [data],
+  );
+
+  const handlePointerLeave = useCallback(() => {
+    setTooltipData(null);
+  }, []);
```
JBWatenbergScality, line 103: Shouldn't be only pointer, but focus changes so that keyboard navigation can trigger the tooltip

src/lib/components/globalhealthbar/components/AlertBar.tsx
```diff
@@ -0,0 +1,44 @@
+import { Bar, BarProps, Rectangle } from 'recharts';
+import { getRectangleProps } from '../healthBarUtils';
+interface AlertBarProps {
+  dataKey: string;
+  yAxisId: string;
+  fill: string;
+  shape: (props: BarProps) => JSX.Element;
```
JBWatenbergScality: pretty sure we can replace this props by alertstartTime and endTime and simplify usage of this alertBar component

src/lib/components/globalhealthbar/healthBarUtils.ts
```diff
@@ -0,0 +1,199 @@
...
+
+// =============================================================================
+// RECTANGLE PROPERTIES
+// =============================================================================
+
+/**
+ * Calculates rectangle properties for alert bars
+ */
+export const getRectangleProps = (
+  props: any,
```
JBWatenbergScality: would be nice to type this

src/lib/components/globalhealthbar/useHealthBarData.ts
```diff
@@ -0,0 +1,68 @@
...
+          (new Date(alert.endsAt).getTime() >= startTimestamp &&
+            new Date(alert.startsAt).getTime() <= endTimestamp) ||
+          (new Date(alert.startsAt).getTime() <= endTimestamp &&
+            new Date(alert.endsAt).getTime() >= startTimestamp),
+      ),
+    [alerts, startTimestamp, endTimestamp],
+  );
+
+  const data = useMemo(
+    () => [
```
JBWatenbergScality: I'm not sure to understand why data needs to be an array if there is always a single entry in there

stories/GlobalHealthBar/globalhealthbarRecharts.stories.tsx
```diff
@@ -0,0 +1,238 @@
...
+//       <>
+//         <HistoryAlertProvider>
+//           <GlobalHealthBar start={start2} end={end2} alerts={alerts} id="1" />
+//           <InputDate start={start2} end={end2} />
+//         </HistoryAlertProvider>
+//         <GlobalHealthBar start={start2} end={end2} alerts={alerts} id="2" />
+//       </>
+//     );
+//   },
+// };
```
JBWatenbergScality: Remove this commented code

stories/GlobalHealthBar/globalhealthbarRecharts.stories.tsx
```diff
@@ -0,0 +1,238 @@
...
+//         <GlobalHealthBar
+//           startDate={startDate}
+//           endDate={endDate}
+//           alerts={alerts}
+//           id="1"
+//         />
+//       </Stack>
+//     );
+//   },
+// };
```
JBWatenbergScality: Remove this commented code

stories/GlobalHealthBar/globalhealthbarRecharts.stories.tsx
```diff
@@ -0,0 +1,238 @@
...
+const start = '2021-01-30T23:00:00Z'; // UTC time
+const end = '2021-02-06T23:00:00Z';
+const start2 = '2021-01-29T14:00:00Z';
+const end2 = '2021-02-05T14:00:00Z';
+const startLast24h = '2021-02-01T00:00:00Z';
+const endLast24h = '2021-02-02T00:00:00Z';
+const startLastHour = '2021-02-01T00:00:00Z';
+const endLastHour = '2021-02-01T01:00:00Z';
+
+const alerts = [
```
JBWatenbergScality, line 44: Can you try with superposed alerts and see the behavior ?

JeanMarcMilletScality: @JBWatenbergScality When there is superposed alert, the critical alerts are displayed on top. On hover, the tooltip display only info for the critical one

src/lib/components/globalhealthbar/useHealthBarData.ts
JeanMarcMilletScality: Because Barchart expect an array for data. 
I compute everything in data so I propose to separate the data needs for the barchart from the data for tooltip and bars

src/lib/components/globalhealthbar/components/HealthBarXAxis.tsx
JeanMarcMilletScality: There is no such logic in Barchart but we could use it in Line chart

src/lib/components/globalhealthbar/components/AlertBar.tsx
JeanMarcMilletScality: @JBWatenbergScality Remove it as proper data for recharts make it useless

JBWatenbergScality (approved):

PR #908, opened 2025-09-09
Added transparent background Layout varient

damiengillesscality:
**Component**:

- Layoutv2

**Description**:

Add a `transparent` varient to the Layout. This will allow guardian to have a glow effect in the background and not the blend background1 color.

**Design**:

**Breaking Changes**:

- None

### DIFF
--- a/src/lib/components/layout/v2/index.tsx
+++ b/src/lib/components/layout/v2/index.tsx
@@ -2,12 +2,12 @@ import { ReactElement } from 'react';
 import styled from 'styled-components';
 import { navbarHeight } from '../../../style/theme';
 
-const LayoutContainer = styled.div`
+const LayoutContainer = styled.div<{ variant?: 'transparent' }>`
   display: flex;
   flex-direction: column;
   height: 100vh;
   box-sizing: border-box;
-  background: ${(props) => props.theme['backgroundLevel1']};
+  background: ${props => props.variant === 'transparent' ? 'transparent' : props.theme.backgroundLevel1};
 `;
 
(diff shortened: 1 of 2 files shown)


src/lib/components/layout/v2/index.tsx
```diff
@@ -2,12 +2,12 @@ import { ReactElement } from 'react';
...
 import { navbarHeight } from '../../../style/theme';
 
-const LayoutContainer = styled.div`
+const LayoutContainer = styled.div<{ variant?: 'transparent' }>`
   display: flex;
   flex-direction: column;
   height: 100vh;
   box-sizing: border-box;
-  background: ${(props) => props.theme['backgroundLevel1']};
+  background: ${props => props.variant === 'transparent' ? 'transparent' : props.theme.backgroundLevel1};
```
JBWatenbergScality, line 10: I wonder if the variant shouldn't be defined on the App container instead. The reason motivating this thinking is that the layout component in our apps is often defined within shell-ui, whereas it is up to the micro app to decide to have a glow background due to its responsibility.

damiengillesscality: The layout could be always transparent and the background colours could be decided by `children` and `headerNavigation` the parameters of Layoutv2 however I think that would be a breaking change to all our apps.

image

JBWatenbergScality: Go for merging as is, we can discuss this when the hypothetical problem I have in mind arise.

JBWatenbergScality (approved):

PR #910, opened 2025-09-15
Improvement/add missing data point fixes

JeanMarcMilletScality:
Add check for data point to avoid creating point outside range
Create data point between start time and firstDatapoint, and between last data point and span end: Avoid chart starting at first data point instead of Startime timestamp (same issue with end)
Add filtering from legend data


ChengYanJin (approved):
LGTM !

PR #916, opened 2025-09-22
Enhance ChartLegendWrapper functionality with dynamic setting of colorSet

JeanMarcMilletScality:
Requirements: 
Current State: ChartLegendWrapper accepts a static colorSet object prop
Goal: Support dynamic colorSet generation where multiple charts can register their series
Two Modes:
Static: colorSet is an object (current behavior)
Dynamic: colorSet is a function (seriesNames: string[]) => Record<string, color>
Dynamic Mode Flow:
Charts call register(chartId, seriesNames[]) using useChartId() and useChartLegend()
ChartLegendWrapper maintains registeredColorSets registry tracking which chart owns which series
When register() called → collect all unique series from registry → call colorSet function → update internal colorSet object
Registry prevents removing colors still used by other charts
Example:
Chart1 registers ["Serie1", "Serie2"], Chart2 registers ["Serie1"]
→ Function receives ["Serie1", "Serie2"] → Returns {Serie1: color1, Serie2: color2}

What this PR do:
- Add `uuid` package (version 13.0.0) and its type definitions (version 10.0.0) to package.json and package-lock.json.
- Refactor `ChartLegendWrapper` to include a new `useChartId` hook for generating unique chart IDs.
- Update the internal state management of `ChartLegendWrapper` to handle dynamic color sets based on registered series names.
- Add story to show example of dynamic setting.


src/lib/components/chartlegend/ChartLegendWrapper.tsx
```diff
@@ -18,23 +31,69 @@ export type ChartLegendState = {
...
+        setInternalColorSet(newColorSet);
+      }
+    }
+  }, [registeredColorSets, colorSet]);
+
+  useEffect(() => {
+    if (typeof colorSet !== 'function') {
+      setInternalColorSet(colorSet);
+    }
+  }, [colorSet]);
```
ChengYanJin: Just wondering why do we have two useEffect handle the same condition ?

ChengYanJin (approved):

PR #918, opened 2025-09-29
Improvement/aix 355 sparkline

damiengillesscality:
**Component**:

- Sparkline

**Description**:

This is a new component based on those specifications:
- [VitalsUI requierments](https://docs.google.com/document/d/1Z5Kkh5LQDxI23q3UqjB8VnlPMadBGAZZ5TJypUDJjAo/edit?tab=t.0)
- [figma](https://www.figma.com/design/N2jOAK44dTSiBclDjXzz6a/ARTESCA?node-id=24564-2151&p=f&t=MA0gQkjtf61qbhbp-0)
- [AIX-355](https://scality.atlassian.net/browse/AIX-355)

This components displays a simple single serie area chart without any interaction (no tooltips).

image

**Design**:

The colour of the charts defaults to `lineColor1` and can be customised via the `serie.color` property. I did not import the red or green logic depending on the value in this component as it would make reusability and extension way harder.

**Breaking Changes**:

- None

[AIX-355]: https://scality.atlassian.net/browse/AIX-355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/components/sparkline/sparkline.component.tsx
+++ b/src/lib/components/sparkline/sparkline.component.tsx
@@ -1,11 +1,11 @@
-import { memo } from "react";
+import { useMemo, useState } from "react";
 import { Area, AreaChart, CartesianGrid, ResponsiveContainer } from "recharts";
 import { chartColors } from "../../style/theme";
 
 type SparklineProps = {
   serie: {
     data: [number, number][],
-    color?: string,
+    color?: string, // exa color code like '#ff0000'
   }
 };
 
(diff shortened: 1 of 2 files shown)


damiengillesscality:
Updated style slightly based on Valamir feedbacks

Cuervino (approved):
LGTM

ChengYanJin (approved):
Just wondering if we want to keep the transform animation for this sparkline chart ? 
because we have removed it for barchart and linechart.

src/lib/components/sparkline/sparkline.component.tsx
```diff
@@ -0,0 +1,44 @@
+import { useMemo } from "react";
+import { Area, AreaChart, CartesianGrid, ResponsiveContainer } from "recharts";
+import { useTheme } from "styled-components";
+import { chartColors } from "../../style/theme";
+
+type SparklineProps = {
+  serie: {
+    data: [number, number][],
```
JBWatenbergScality: I think it should support missing timestamp similarly as other time series component

damiengillesscality: Allowed missing values and added tests to storybook

JBWatenbergScality:
IMHO we have to addMissingTimestamp the same way it is done in linetemporalchart. It should be the responsibility of our charts to ensure the continuity of timeintervals i think

PR #925, opened 2025-10-01
enhance SearchInput

hervedombya:
**Component**:

**Description**:

**Design**:

**Breaking Changes**:

- [] Breaking Changes

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/src/lib/components/inputv2/inputv2.tsx
+++ b/src/lib/components/inputv2/inputv2.tsx
@@ -3,6 +3,7 @@ import styled, { css } from 'styled-components';
 import { spacing } from '../../spacing';
 import { DESCRIPTION_PREFIX, useFieldContext } from '../form/Form.component';
 import { Icon, IconName } from '../icon/Icon.component';
+import { CoreUITheme } from '../../style/theme';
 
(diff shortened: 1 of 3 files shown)


src/lib/components/inputv2/inputv2.tsx
```diff
@@ -92,9 +93,9 @@ const InputBorder = styled.div<{
   }
 `;
 
-const SelfCenterredIcon = styled(Icon)`
+const SelfCenterredIcon = styled(Icon)<{ color: keyof CoreUITheme }>`
   align-self: center;
-  color: ${(props) => props.theme.textSecondary};
+  color: ${(props) => props.color};
```
damiengillesscality: Did it work ? I'm pretty sure you are writing `color: textSecondary` which is not valid css. You probably meant `props.theme[color]`.
Can you add an example in a story please.

hervedombya: It works because the `Icon` itself add `color: props.theme[color]` but your right we should should have this here.
Fixed in 220e9414ff0af6bcd73d66c2f10faffc59a7a9d5

damiengillesscality (approved):

PR #936, opened 2025-10-09
Use same date format for tooltip and X axis for line chart and barchart

JeanMarcMilletScality:
Use same format in tooltip and X axis label for line chart and barchart

### DIFF
--- a/src/lib/components/date/FormattedDateTime.tsx
+++ b/src/lib/components/date/FormattedDateTime.tsx
@@ -65,6 +65,27 @@ export const TIME_FORMATER = Intl.DateTimeFormat('en-GB', {
 });
 
 /**
+ * @description Day month abbreviated formatter. Used for describing long term date.
+ * @example 6 Oct
+ */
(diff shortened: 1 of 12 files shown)


src/lib/components/date/FormattedDateTime.tsx
```diff
@@ -64,6 +64,27 @@ export const TIME_FORMATER = Intl.DateTimeFormat('en-GB', {
...
+ * @description Day month abbreviated formatter. Used for describing long term date.
+ * @example 6 Oct 25
+ */
+export const DAY_MONTH_ABBREVIATED_YEAR = Intl.DateTimeFormat('en-GB', {
+  day: '2-digit',
+  month: 'short',
+  year: '2-digit',
+  hour12: false,
+});
+
```
ChengYanJin, line 87: Not sure we can use this format for long term metrics, I will discuss with @Cuervino

ChengYanJin (approved):

PR #938, opened 2025-10-09
Create portal for chart tooltip

JeanMarcMilletScality:
Create Portal tooltip to avoid clipping and tooltip being render under other charts
Remove focus style on the chart
add focus and hover style on bar in globalhealthbar

### DIFF
--- a/src/lib/components/charttooltip/ChartTooltip.tsx
+++ b/src/lib/components/charttooltip/ChartTooltip.tsx
@@ -21,7 +21,8 @@ export const ChartTooltipContainer = styled.div`
   font-size: ${fontSize.small};
   padding: ${spacing.r8};
   min-width: 10rem;
-  max-width: 250px;
+  max-width: 40rem;
+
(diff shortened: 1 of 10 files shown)


src/lib/components/charttooltip/ChartTooltip.tsx
```diff
@@ -121,3 +131,125 @@ export const TooltipHeader = ({
     </ChartTooltipHeader>
   );
 };
+
+export interface ChartTooltipPortalProps {
+  children: React.ReactNode;
+  coordinate?: { x: number; y: number };
+  chartContainerRef: React.RefObject<HTMLDivElement>;
+  isVisible?: boolean;
+  middleware?: any[];
```
ChengYanJin:
```suggestion
import type { Middleware } from '@floating-ui/react';
  middleware?: Middleware[];
```

ChengYanJin (approved):

PR #939, opened 2025-10-10
add sort() to legend wrapper

JeanMarcMilletScality:
Add sort to list resources to avoid issue with object.keys sorting when keys is a number


ChengYanJin (approved):
LGTM!

PR #941, opened 2025-10-13
Improve chart labels

JeanMarcMilletScality:
Various improvement on label for charts
Use getRoundreferecneValue for Line chart to add space between top chart value and top data value
Create fn to getTicks for charts to avoid issues with weird steps on Y axis (incosnistent steps and number of ticks, not showing 0 on symmetrical chart)
Put HH:MM on X axis label for span <= 7days

### DIFF
--- a/src/lib/components/barchartv2/utils.ts
+++ b/src/lib/components/barchartv2/utils.ts
@@ -21,9 +21,7 @@ export const getRoundReferenceValue = (value: number): number => {
   let result: number;
 
   if (normalized <= 1) result = magnitude;
-  else if (value > 10 && normalized <= 1.5) result = 1.5 * magnitude;
   else if (normalized <= 2) result = 2 * magnitude;
-  else if (value > 10 && normalized <= 3) result = 3 * magnitude;
   else if (value > 10 && normalized <= 4) result = 4 * magnitude;
   else if (normalized <= 5) result = 5 * magnitude;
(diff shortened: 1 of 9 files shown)


src/lib/components/barchartv2/utils.ts
```diff
@@ -4,23 +4,55 @@ import { chartColors, ChartColors } from '../../style/theme';
...
   else if (normalized <= 5) result = 5 * magnitude;
+  else if (value > 10 && normalized <= 7.5) result = 7.5 * magnitude;
   else result = 10 * magnitude;
 
-  // Ensure minimum value of 5 for better chart appearance
-  return Math.max(result, 5);
+  return result;
+};
+
+export const getTicks = (topValue: number, isSymmetrical: boolean) => {
```
ChengYanJin, line 33: Should we add unit test to check if we get the expected ticks ?

src/lib/components/linetimeseriechart/utils.ts
```diff
@@ -27,8 +27,9 @@ export const formatXAxisLabel = (
   const date = new Date(timestamp);
   if (duration <= 24 * 60 * 60) {
     return TIME_FORMATER.format(date);
+    //? At which point do we consider chart to be long term? 1 week? 2 weeks? 1 month?
```
ChengYanJin: I think longer than 7 days (> 7 days)

ChengYanJin (approved):
LGTM!

PR #945, opened 2025-10-15
Improvement/charts restructuration

JeanMarcMilletScality:
Restructure charts components
Edit utils fn to share logic between charts when possible

Remove Vega

⚠️ This PR remove Vega file and move a lot of files and could break imports


ChengYanJin (approved):
LGTM !!

PR #946, opened 2025-10-20
use second for all date formatting utils fn for charts

JeanMarcMilletScality:


src/lib/components/barchartv2/Barchart.component.tsx
```diff
@@ -160,7 +160,7 @@ export const CustomTick = ({
           <Text variant="Smaller">
             {type.type === 'time' ? (
               <FormattedDateTime
-                format={formatDate(duration)}
+                format={formatDate(duration / 1000)}
```
ChengYanJin: It is little bit confusing for me to see divided duration by 1000 ;)
Maybe when we compute the duration, we could have already divided by 1000 ?

```
  const duration =
    type.type === 'time'
      ? type.timeRange.endDate.getTime() - type.timeRange.startDate.getTime()
      : 0;
```

JeanMarcMilletScality: 👍

ChengYanJin (approved):

PR #950, opened 2025-10-21
add more possible top value for charts

JeanMarcMilletScality:
Enhance getRoundReferenceValue and getTicks functions for improved tick calculation and rounding logic in charts
Avoid too big space between 100 and 200 or 200 and 400, etc


ChengYanJin (approved):
Thanks !

PR #972, opened 2025-11-20
Add auto-scroll to selected row

hervedombya:
**Component**:

**Description**:

**Design**:

**Breaking Changes**:

- [] Breaking Changes

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/src/lib/components/tablev2/TableCommon.tsx
+++ b/src/lib/components/tablev2/TableCommon.tsx
@@ -65,8 +65,8 @@ export const VirtualizedRows = <
       return (
         <List
           height={height - 1}
-          itemCount={rows.length}
-          itemSize={convertRemToPixels(tableRowHeight[rowHeight])}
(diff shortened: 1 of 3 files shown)


src/lib/components/tablev2/TableCommon.tsx
```diff
@@ -122,14 +137,23 @@ type TableRowsProps<
   RenderRow: React.MemoExoticComponent<
     ({ index, style }: RenderRowType) => JSX.Element
   >;
+  listRef?: LegacyRef<FixedSizeList<Row<DATA_ROW>[]>>;
```
ChengYanJin: Just wondering should we get rid or LegacyRef ?

hervedombya: yes we should, fixed in 7506203c67a0480267edd4333094272a3ba99598

ChengYanJin (approved):
LGTM!

PR #991, opened 2026-01-21
Improvement/artesca 16534 chart axis issues

JeanMarcMilletScality:


src/lib/components/charts/linetimeseries/LineTimeSerieChart.tsx
```diff
@@ -558,27 +190,17 @@ export function LineTimeSerieChart({
...
-                  <Line
-                    key={`${title}-${resource}-${serieIndex}`}
-                    type="monotone"
-                    dataKey={label}
-                    stroke={colorMapping[resource]}
-                    dot={false}
-                    isAnimationActive={false}
-                    strokeDasharray={serie.isLineDashed ? '4 4' : undefined}
-                    onMouseEnter={() => setHoveredValue(label)}
-                    onMouseLeave={() => setHoveredValue(undefined)}
```
ZiyangLinScality, line 577: The Line components previously had onMouseEnter/onMouseLeave handlers that updated hoveredValue state, which was used to highlight the corresponding item in the     
  tooltip (via isHovered prop on ChartTooltipItem). This appears to have been removed. Was this intentional, or should the hover highlighting be preserved?

ZiyangLinScality (approved):

PR #993, opened 2026-01-23
Table selected row misalignment

JeanMarcMilletScality:
Border was added to row on selection causing its content to shift to the left, misaligning it with other rows.
Replace the border with box shadow: the visual effect is the same but does not cause shifting


src/lib/components/tablev2/Tablestyle.tsx
```diff
@@ -8,7 +8,7 @@ import { HeaderGroup } from 'react-table';
 import { Icon } from '../icon/Icon.component';
 import { FocusVisibleStyle } from '../buttonv2/Buttonv2.component';
 import { spacing } from '../../spacing';
-import { Box } from '../box/Box';
```
ZiyangLinScality, line 11: should format first, this line change nothing

ZiyangLinScality (approved):

PR #1000, opened 2026-02-10
Improvement/misc improvement placeholder line height

JeanMarcMilletScality:
### Placeholder

Placeholder can be too long/not useful. The prefix sometimes unnecessarily extends the placeholder (SearchInput case). Or the input is too small (compared to available place)

- Input has a default prefix for placeholder. We don't want it for SearchInput and possibly others cases but we want it most of the times. Add a props that remove it if needed
- Add size to InputList to be able to handle size. Sometimes placeholder is too long while there is a lot of room for input

### Line-Height

Line height is too big for constrained height container (Table, Select options, etc).
- Add a prop in text to allow the change of line-height. 
For now it is used for helper text but it could be used elsewhere. Create a Text wrapper specific to helper text 
- Change line-height for ConstrainedText on multiple lines

Ref: https://docs.google.com/document/d/1Mtv5SFpai_2FxN-jh00eZ6LShaUoUTZ4-rIDrX90Xnw/edit?tab=t.hvadtzdafoq5#heading=h.gfn68ibob54k

### DIFF
--- a/src/lib/components/banner/Banner.component.test.tsx
+++ b/src/lib/components/banner/Banner.component.test.tsx
@@ -1,0 +1,58 @@
+import React from 'react';
+import '@testing-library/jest-dom';
+import { render, screen, waitFor } from '@testing-library/react';
+import { Banner } from './Banner.component';
+import { Icon } from '../icon/Icon.component';
+import { getWrapper } from '../../testUtils';
+
+describe('Banner', () => {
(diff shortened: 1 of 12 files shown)


MonPote (approved):
LGTM, we probably want to make a PR for the formating too.
More than half of the change are formating noise 🥲

PR #1014, opened 2026-02-25
CUI-5: Prevent Select from triggering implicit form submission on Enter

JeanMarcMilletScality:
## TL;DR

Pressing **Enter** on a `Select` inside any `<form>` no longer submits the form. The Select widget now consumes Enter itself — matching the WAI-ARIA combobox pattern — so the fix works in every `<form>`, not just core-ui's own `<Form>`.

## Context

`CUI-5`: pressing Enter on a `Select` triggered the browser's implicit form submission. Most visible on a **searchable** Select (>8 options) — pressing Enter in its search input submitted the form whenever the query was empty or matched no option, instead of staying inside the combobox.

## Approach

**Root cause.** react-select renders a real `<input>`; the browser's [implicit submission](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#implicit-submission) treats it like any text field and submits the enclosing `<form>` on Enter whenever react-select doesn't consume Enter itself (menu closed, or search field with no highlighted option).

**The fix** — wrap react-select in a zero-layout `display: contents` div and swallow Enter's default action on the **bubble** phase, after react-select has already done its own work:

```tsx
<div
  style={{ display: 'contents' }}                        // ← no layout box: zero width/flex/margin impact
  onKeyDown={(event) => {
    if (event.key === 'Enter') event.preventDefault();   // ← kills implicit submit, not react-select's own handling
  }}
>
  <ReactSelect ... />
</div>
```

Belongs in the widget, not the form: a combobox owns its keyboard behaviour ([WAI-ARIA combobox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/)), and a form-level fix would only work inside core-ui's own `<Form>` — not when a consumer wraps the Select in their own `<form>`. See the alternatives below.

**Alternatives considered:**

| Approach | Why rejected |
| --- | --- |
| Form-level handler with a text-input whitelist | Only blocked Enter on readonly inputs, so a **searchable** Select still submitted; and did nothing inside a consumer's own `<form>`. Flagged in review ([discussion](https://github.com/scality/core-ui/pull/1014#discussion_r2857815199)). |
| `preventDefault()` inside react-select's `onKeyDown` prop | react-select [checks `event.defaultPrevented`](https://github.com/jedwatson/react-select/blob/v3.2.0/packages/react-select/src/Select.js#L1237-L1241) and skips *all* keyboard handling if set — breaks menu open / option select. |
| `stopPropagation()` inside react-select's `onKeyDown` prop | Implicit submission is a browser default action, not event bubbling — `stopPropagation()` cannot prevent it. |

## Review focus

- 🟡 **`selectv2/Selectv2.component.tsx › SelectBox`** — Enter is now consumed for *every* Select in every consuming app. This is spec-correct (a combobox shouldn't submit its form), but it is a behaviour change: confirm no flow relied on Enter-to-submit while a Select was focused. The bubble-phase `preventDefault()` leaves react-select's own selection intact (verified by the searchable test).
- ⚪ **`selectv2/selectv2.test.tsx`** — two regression tests (closed Select, searchable Select); the searchable one also asserts Enter still *selects* the highlighted option.
- ⚪ **`form/Form.test.tsx`** — guards the opposite contract: a plain text input inside `<Form>` still submits on Enter.

## How to test

1. Render a `Select` inside a `<form>` with a `type="submit"` button (Storybook: `npm run storybook`, then a Select story wrapped in a form).
2. **Non-searchable** (≤8 options): focus the Select, press **Enter** → menu opens, **form does not submit**.
3. **Searchable** (>8 options): open it, type a query (including an empty / no-match query), press **Enter** → highlighted option is selected (or nothing, if none), **form does not submit**.
4. **Regression:** a plain `<input type="text">` in the same form → **Enter still submits**.

## References

- **[CUI-5](https://scality.atlassian.net/browse/CUI-5)** — Bug: Enter on a Select's search input submits the form when the query is empty or matches no option.
- **[Review discussion r2857815199](https://github.com/scality/core-ui/pull/1014#discussion_r2857815199)** — @JBWatenbergScality flagged that the original form-level approach still let a searchable Select submit; drove the move into the widget.

[CUI-5]: https://scality.atlassian.net/browse/CUI-5?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/components/form/Form.component.tsx
+++ b/src/lib/components/form/Form.component.tsx
@@ -4,7 +4,6 @@ import {
   FormHTMLAttributes,
   forwardRef,
   isValidElement,
-  KeyboardEvent,
   ReactElement,
   ReactNode,
   useContext,
@@ -272,22 +271,6 @@ const FormSection = ({
   );
 };
 
-const TEXT_INPUT_TYPES = ['text', 'search', 'url', 'tel', 'password', 'email', 'number'];
-
-const preventImplicitSubmission = (
-  event: KeyboardEvent<HTMLFormElement>,
-) => {
-  if (event.key !== 'Enter') return;
-  const target = event.target as HTMLElement;
-  const isTextInput =
-    target instanceof HTMLInputElement &&
-    TEXT_INPUT_TYPES.includes(target.type) &&
-    !target.readOnly;
-  if (!isTextInput && !(target instanceof HTMLButtonElement)) {
-    event.preventDefault();
-  }
-};
-
 const PageForm = forwardRef<HTMLFormElement, PageFormProps>(
   (
     { layout, leftActions, rightActions, children, banner, ...formProps },
@@ -296,16 +279,7 @@ const PageForm = forwardRef<HTMLFormElement, PageFormProps>(
     const requireMode = useContext(RequireModeContext);
     return (
       <ScrollbarWrapper>
-        <StyledForm
-          {...formProps}
-          onKeyDown={(event) => {
-            preventImplicitSubmission(event);
-            formProps.onKeyDown?.(event);
-          }}
-          noValidate
-          ref={ref}
-          layout={layout}
-        >
+        <StyledForm {...formProps} noValidate ref={ref} layout={layout}>
           <FixedHeader layoutKind="page">
             <PaddedForHeaderAndFooterContent>
               <Wrap>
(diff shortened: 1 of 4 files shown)


src/lib/components/form/Form.component.tsx
```diff
@@ -271,6 +272,22 @@ const FormSection = ({
...
+  if (event.key !== 'Enter') return;
+  const target = event.target as HTMLElement;
+  const isTextInput =
+    target instanceof HTMLInputElement &&
+    TEXT_INPUT_TYPES.includes(target.type) &&
+    !target.readOnly;
+  if (!isTextInput && !(target instanceof HTMLButtonElement)) {
+    event.preventDefault();
+  }
+};
```
JBWatenbergScality: I'm really unsure about this proposal. What if we have a select with search field (because we have more than 8 items) ? Then pressing enter on the search field trigger the submission. 

Are we trying to solve a real problem in this PR ? What is described as a pb in here is a standard behavior I think.

MonPote (approved):
LGTM

PR #1015, opened 2026-02-26
Fix CopyButton 'Copied' tooltip not showing in Firefox

JeanMarcMilletScality:
## Context
Issue found on the Kasten ISV assistant summary page. The copy button did not show the "Copied !" tooltip in Firefox after the user copied content. The issue was reported on Firefox / Windows 11 (Chrome had no issue).
Ticket: https://scality.atlassian.net/browse/ZKUI-459

This issue applies to all copy button

## Problem
When the user clicked copy, the button was set to `disabled` on success. In Firefox, disabling the button causes it to lose focus and fire `onBlur` on the Tooltip wrapper. The Tooltip hides on blur, so the "Copied !" message disappeared immediately and was never shown.

## Changes
- **Do not set `disabled` on success** — `CopyButton` now only passes through `props.disabled`, so the button stays focusable and the tooltip remains visible.
- **Disabled-like behavior during success** — When `copyStatus === 'success'`, the button gets `cursor: not-allowed` and `opacity: 0.5` so it looks disabled.
- **No-op on click during success** — `onClick` only calls `copy()` when not in success state, so the button is not clickable in practice for the ~2s feedback period.
- **Promise handling** — `navigator.clipboard.writeText()` is now awaited with `.then()/.catch()` so success/unsupported state is set correctly (e.g. if clipboard fails in some environments).

### DIFF
--- a/src/lib/components/buttonv2/CopyButton.component.test.tsx
+++ b/src/lib/components/buttonv2/CopyButton.component.test.tsx
@@ -1,0 +1,252 @@
+import "@testing-library/jest-dom";
+import { act, render, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import React from "react";
+import { getWrapper } from "../../testUtils";
+import { CopyButton } from "./CopyButton.component";
+
+describe("CopyButton", () => {
+	const { Wrapper } = getWrapper();
+	const originalClipboard = navigator.clipboard;
+
+	beforeEach(() => {
+		Object.assign(navigator, {
+			clipboard: {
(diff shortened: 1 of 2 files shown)


ZiyangLinScality (commented):
Consider adding aria-disabled={isSuccess || props.disabled} to maintain accessibility semantics.

ZiyangLinScality:
Need some unit test at lease:
- When writeText returns a rejected Promise, it should enter the unsupported state.
- Clicking in the success state should not call copy again.

ZiyangLinScality (approved):

PR #1024, opened 2026-03-04
feature/valalint

damiengillesscality:
Resolves https://scality.atlassian.net/browse/CUI-2 by implementing an EsLint rule plugin that check for capitalisation issues.
It can be used easily by configuring projects like this:

```js
import valalint from '@scality/core-ui/eslint-plugin';

export default [
  {
    files: ['**/*.{js,jsx,ts,tsx}'],
    ignores: ['src/gen/**'],
    ...valalint.configs['flat/recommended'],
  },
];
```

The rule is also tested using the existing jest setup.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -73,8 +73,8 @@
         "@types/styled-components": "^5.1.25",
         "@types/styled-system": "^5.1.15",
         "@types/uuid": "^10.0.0",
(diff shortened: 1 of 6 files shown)


damiengillesscality:
Bad branch name, claused for #1025

PR #1026, opened 2026-03-09
fix: harden CI/CD workflows against supply-chain and token exposure risks

damiengillesscality:
## Security: harden CI/CD workflows

Three vulnerabilities identified during a security audit of the GitHub Actions workflows have been addressed.

### 1. Token exposed in git remote URL (`github-pages.yml`)
The `Deploy Storybook` workflow was constructing the git remote URL with `${{ secrets.GITHUB_TOKEN }}` interpolated directly into the shell command, risking token leakage in runner logs. The token is now passed to `actions/checkout` via its `token:` input, which handles credential storage safely and opaquely.

### 2. Non-reproducible npm install before publish (`post-release.yml`)
The `publish-npm` job was using `npm install`, which can silently resolve dependency versions that differ from the reviewed lockfile. Replaced with `npm ci` to guarantee the published package is built from an exact, reproducible dependency tree.

### 3. Third-party action with PAT access replaced (`post-release.yml`)
The `bump-version` job delegated branch creation and PR opening to the `peter-evans/create-pull-request@v3` third-party action, which was granted the `GIT_ACCESS_TOKEN` PAT. This introduced a supply-chain risk: a compromised or tag-moved action could exfiltrate the token or make arbitrary changes to the repository. The action has been replaced with a native shell step (for the git commit/push) and `actions/github-script` (already present in the same job) for PR creation via the GitHub REST API — eliminating the third-party dependency entirely.

PR #1027, opened 2026-03-12
fix: harden CI/CD workflows against supply-chain and token exposure r…

damiengillesscality:
see #1026

### DIFF
--- a/.github/workflows/github-pages.yml
+++ b/.github/workflows/github-pages.yml
@@ -6,10 +6,10 @@ on:
 jobs:
   deploy:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     steps:
       - uses: actions/checkout@v4
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
       - uses: actions/setup-node@v4
         with:
           node-version: 20
--- a/.github/workflows/github-pages.yml
+++ b/.github/workflows/github-pages.yml
@@ -8,12 +8,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
       - uses: actions/setup-node@v4
         with:
           node-version: 20
-      - run: |
-          git remote set-url origin https://${{ github.actor }}:${{ secrets.GITHUB_TOKEN }}@github.com/$GITHUB_REPOSITORY
-          npm ci
-          npm run storybook:deploy
+      - run: npm ci
+      - run: npm run storybook:deploy
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/github-pages.yml
```diff
@@ -8,12 +8,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
JBWatenbergScality: Not sure this is really required

.github/workflows/post-release.yml
```diff
@@ -9,21 +9,36 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.GIT_ACCESS_TOKEN }}
```
JBWatenbergScality: I think it can be dropped. Checkout action could be bumped

.github/workflows/post-release.yml
```diff
@@ -9,21 +9,36 @@ jobs:
...
+        uses: actions/github-script@v5
         id: cpr
         with:
-          token: ${{ secrets.GIT_ACCESS_TOKEN }}
-          branch: feature/bump-core-ui-version-to-${{ env.nextVersion }}
-          commit-message: Bump core-ui version to  ${{ env.nextVersion }}
-          title: Bump core-ui version to  ${{ env.nextVersion }}
-          body: ""
-          base: ${{ github.event.repository.default_branch }}
+          github-token: ${{ secrets.GIT_ACCESS_TOKEN }}
```
JBWatenbergScality: probably unrequired

.github/workflows/post-release.yml
damiengillesscality: Unless the default GITHUB_TOKEN is set to have the write to push we need it to push bellow line 26, it depends on the setup.

.github/workflows/post-release.yml
damiengillesscality: Same depends if the default GITHUB_TOKEN is set to have the `pull-requests: write`

.github/workflows/post-release.yml
```diff
@@ -7,27 +7,41 @@ on:
...
+      contents: write
+      pull-requests: write
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - name: Install node
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v4
         with:
           node-version: "16"
```
MonPote, line 18: nit: should we also take the opportunity to bump node-version to at least 22 ?

MonPote (approved):

PR #1032, opened 2026-03-16
feat: add Drawer component

hervedombya:
## Summary

- Adds a `Drawer` component: a slide-in panel that can open from any edge (left/right/top/bottom)
- Supports optional overlay backdrop, configurable size, footer slot, and Escape key to close
- Renders via portal (same pattern as Modal) with dedicated `drawer` zIndex (8200, between overlay and modal)

## Context

Needed by the Data Browser theme customization feature (DBR-4) as the container for the branding editor panel.

Ref: https://scality.atlassian.net/browse/CUI-10

### DIFF
--- a/src/lib/components/drawer/Drawer.component.test.tsx
+++ b/src/lib/components/drawer/Drawer.component.test.tsx
@@ -1,0 +1,108 @@
+import '@testing-library/jest-dom';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import React from 'react';
+import { getWrapper } from '../../testUtils';
+import { Drawer } from './Drawer.component';
+
+describe('Drawer', () => {
+  const { Wrapper } = getWrapper();
+
+  it('should not be visible when closed', () => {
+    render(
+      <Wrapper>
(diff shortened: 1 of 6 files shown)


Cuervino (commented):
1. I'm not sure about what bgColor, I would do the same as a Modal or a page for the title / content / footer.
2. Save button should be larger, maybe it's a button min-width.
3. Maybe a border on the side?
4. I like the shadow effect.

JeanMarcMilletScality (commented):
I found an issue with the close button. It is still focusable while the drawer is closed (using keyboard) 👇 
Image

hervedombya:
> 1. I'm not sure about what bgColor, I would do the same as a Modal or a page for the title / content / footer.
> 2. Save button should be larger, maybe it's a button min-width.
> 3. Maybe a border on the side?
> 4. I like the shadow effect.

1. Align background colors with Modal: header=`backgroundLevel3`, body=`backgroundLevel4`, footer=`backgroundLevel3`
2. Added `minWidth` for storybook stories
3. Added 1px solid border on the side facing the content
4. 👍

JeanMarcMilletScality (approved):

PR #1033, opened 2026-03-16
Add Modal component guideline

Cuervino:
## Description

  This PR introduces the first complete design guideline for the Modal component in Storybook, along with a rewrite of the existing stories to match the documented
  conventions.

  The `modal.mdx` file that previously existed was a placeholder (a single Canvas and two story references with no documentation). It has been replaced by
  `modal.guideline.mdx`, a full guideline covering all usage decisions for the component.

  The `modal.stories.tsx` file has been rewritten: the existing stories used `"Yes"` / `"No"` as button labels, which violates the `modal-button-forbidden-label` valalint
  rule. Footer alignment was also incorrect (`space-between` instead of `flex-end`). A new `DestructiveModal` story has been added to illustrate the `alertdialog` pattern.

  ### Change list

  | Change | File | Comment |
  |--------|------|---------|
  | Added | `stories/modal.guideline.mdx` | Full design guideline for the Modal component |
  | Deleted | `stories/modal.mdx` | Was a placeholder, replaced by the guideline |
  | Rewritten | `stories/modal.stories.tsx` | Correct button labels, footer alignment, new DestructiveModal story |
  | Updated | `.storybook/preview.js` | storySort updated to display Guideline before stories for all components |

  ## Design

  Guideline covers:

  - **Role** — `dialog` vs `alertdialog`: when to use each, ESC behavior, close button presence
  - **Size** — min-width 480px, max-width 90vw
  - **Scroll** — avoid when possible, vertical padding against viewport, visible scrollbar on overflow
  - **Title conventions** — mirrors the trigger action verb; question mark for destructive/irreversible actions
  - **Footer** — always required, right-aligned, 2 buttons standard / 3 acceptable, no links
  - **Destructive modals** — `alertdialog` role, `danger` variant, resource name mandatory in body if it exists
  - **Async** — `isLoading` under 3s, contextual message over 3s, no modal for long operations
  - **Forbidden patterns** — no modal on modal, no event-triggered modals at app level

  ## Breaking Changes
No


stories/modal.guideline.mdx
```diff
@@ -0,0 +1,103 @@
...
+- The **close button** (×) in the header is present by default via the `close` prop. Omit it only when using `role="alertdialog"` — the user must explicitly choose an action.
+- **Clicking the backdrop never closes the modal.** This prevents accidental data loss, especially in forms.
+- **ESC** closes `dialog`-role modals. It does not close `alertdialog`-role modals.
+
+## Title and subtitle
+
+- `title` is always required and always text. It mirrors the verb of the action that triggered the modal: the button "Delete node" opens a modal titled "Delete node?".
+- For standard modals, the title describes the action: "Edit settings", "Create bucket".
+- For destructive or irreversible actions, the title is phrased as a question: "Delete node?", "Revoke access?". The question mark signals that a decision is required.
+- `subTitle` is used for step indicators in multi-step modals only (e.g. `"Step 1/2"`).
```
JBWatenbergScality: This is opposed to:

> **❌ Do not use a modal for:**
> - Multi-step flows — use a full page or wizard instead

Are we missing some use case where we allow multi-steps modals ?

stories/modal.guideline.mdx
```diff
@@ -0,0 +1,103 @@
...
+| `dialog` (default) | Standard interaction — form, optional confirmation | Yes | Present |
+| `alertdialog` | Critical or destructive action requiring a forced decision | No | Omit |
+
+Use `alertdialog` when accidentally dismissing the modal would be dangerous or meaningless (e.g. a delete confirmation — ESC should not silently cancel the user's intent).
+
+## Close button and backdrop
+
+- The **close button** (×) in the header is present by default via the `close` prop. Omit it only when using `role="alertdialog"` — the user must explicitly choose an action.
+- **Clicking the backdrop never closes the modal.** This prevents accidental data loss, especially in forms.
+- **ESC** closes `dialog`-role modals. It does not close `alertdialog`-role modals.
```
JBWatenbergScality: I'm unsure about those esc key rules. I'm afraid it can restricts accessibility.

stories/modal.guideline.mdx
```diff
@@ -0,0 +1,103 @@
...
+| < 3s | Use `isLoading` on the action button. Keep the modal open until resolved. |
+| > 3s | Add a contextual message: "This may take a few moments." |
+| Minutes or more | Do not use a modal. Use a background job with status feedback elsewhere in the UI. |
+
+On error: stay in the modal and display the error inline. Do not close the modal silently. Do not rely on a toast as the only error signal.
+
+## Forbidden patterns
+
+- **No modal on top of another modal.** If a flow requires a second modal, redesign as a full page or wizard.
+- **No event-triggered modals** at the application level (login, WebSocket events, etc.) — race conditions between modals create an uncontrollable stacking experience.
```
JBWatenbergScality: We have the ISV exception case.

stories/modal.guideline.mdx
Cuervino: I believe the `subtitle` attribute is shared by both the `modal` and the full-page `form` component. 
However, I suspect the modal header might not actually use it for descriptive text (at least not in the standard sense). It’s possible it’s being only used for the '* indicates required fields' disclaimer. 

I'll remove this mention to the multi-step modal. The Download modal is the only kind of multistep modal I can think of, but we don't display steps per say.

stories/modal.guideline.mdx
Cuervino: Do you mean you are pro or against it?

stories/modal.guideline.mdx
Cuervino: Yeah, indeed, I'll say that we need to limit it, not to forbid it.

stories/modal.guideline.mdx
```diff
@@ -0,0 +1,103 @@
...
+| `role` | When to use | ESC closes it | Close button |
+|---|---|---|---|
+| `dialog` (default) | Standard interaction — form, optional confirmation | Yes | Present |
+| `alertdialog` | Critical or destructive action requiring a forced decision | No | Omit |
+
+Use `alertdialog` when accidentally dismissing the modal would be dangerous or meaningless (e.g. a delete confirmation — ESC should not silently cancel the user's intent).
+
+## Close button and backdrop
+
+- The **close button** (×) in the header is present by default via the `close` prop. Omit it only when using `role="alertdialog"` — the user must explicitly choose an action.
```
MonPote, line 41: So this is a new logic to be implemented right ?

Cuervino: You're right to flag this. After checking the component code:

- ESC behavior is not tied to the `role` prop
- It solely depends on whether the `close` prop is provided
- If `close` is passed, ESC closes the modal (and the close button is shown)
- If not, ESC does nothing

Since my goal is to document existing behavior rather than introduce new logic, I'll update the guideline accordingly and remove the `dialog` vs `alertdialog` distinction for ESC.

stories/modal.guideline.mdx
Cuervino: Done — I've removed the multi-step reference from `subTitle`. The description now reflects its actual role: it occupies the header slot when no close button is present, and can be used for contextual information such as a step indicator or a required-fields disclaimer.

stories/modal.guideline.mdx
Cuervino: I think I get the concern, but I don't think it's actually a trap -- WCAG 2.1.2 doesn't require ESC specifically, just that keyboard users can get out. And they can, via the footer buttons. The whole point of blocking ESC on an `alertdialog` is to force an explicit choice, not to trap anyone. Is that what you were talking about?

stories/modal.guideline.mdx
Cuervino: Good point, I changed 'No' to 'Avoid' -- less absolute, leaves room for edge cases without making them a recommendation in the guideline.

MonPote (approved):

JBWatenbergScality (approved):

PR #1042, opened 2026-03-23
Bugfix/module export

JeanMarcMilletScality:
Fix issue with type module in core-ui


src/lib/valalint/index.mjs
```diff
@@ -0,0 +1,49 @@
+import tsParser from "@typescript-eslint/parser";
+import modalButtonForbiddenLabel from "./rules/modal-button-forbidden-label.js";
+import noRawNumberInJsx from "./rules/no-raw-number-in-jsx.js";
+import technicalSentenceCase from "./rules/technical-sentence-case.js";
```
damiengillesscality: We may want mjs here too

JeanMarcMilletScality: 👍

damiengillesscality (approved):

PR #1052, opened 2026-03-26
Spec: Deployment name inline edit

Cuervino:
## Purpose

This PR is a **design spec**, not a production implementation. It contains a Storybook template story that illustrates the full interaction model for inline deployment name editing in the navbar. It is intended for the development team to understand the requirements before implementation.

Open the story at **Templates / Deployment Name Edit** in Storybook.

### Navbar — editable name

- The deployment name appears next to the logo in the navbar.
- On hover, a border appears around the name and a tooltip reads "Edit deployment name".
- On click, the name is replaced by an inline input, pre-filled with the current name and auto-focused.
- Pressing **Enter** or clicking outside the input triggers the confirmation modal (if the name changed and is non-empty).
- Pressing **Escape** cancels editing and restores the original name.

### Confirmation modal

- Title: "Rename deployment?"
- Body contains:
  - An info block explaining what the deployment name is and where it is used.
  - The confirmation question.
  - A key-value summary: current name / new name.
  - A note (italic, secondary text): "This change may take a few minutes to propagate across all services."
- Footer: Cancel (outline, left) + Rename (primary, right).

### Pending state (modal open)

- While the confirmation modal is open, the navbar already shows the new name in a muted/pending state (opacity reduced), not the original name.
- If the user cancels, the original name is restored.

### Propagation state (post-confirmation)

- After confirmation, the name updates immediately and a propagation simulation begins.
- During propagation: a small inline spinner appears next to the name, the name is muted and non-interactive.
- On hover during propagation, a tooltip reads "Cannot edit while propagating".
- After propagation completes, the name returns to its normal interactive state.

## Style proposals

A second story **Style Proposals** shows four visual treatments for the deployment name in the navbar, to help decide how to visually distinguish it from navigation tab items.

### DIFF
--- a/stories/Templates/deployment-name-edit.stories.tsx
+++ b/stories/Templates/deployment-name-edit.stories.tsx
@@ -1,0 +1,474 @@
+import React, { useEffect, useRef, useState } from 'react';
+import { Meta, StoryObj } from '@storybook/react';
+import styled, { createGlobalStyle } from 'styled-components';
+import { Navbar } from '../../src/lib/components/navbar/Navbar.component';
+import { Modal } from '../../src/lib/components/modal/Modal.component';
+import { Tooltip } from '../../src/lib/components/tooltip/Tooltip.component';
+import { Button } from '../../src/lib/components/buttonv2/Buttonv2.component';
+import { Icon } from '../../src/lib/components/icon/Icon.component';
+import { Loader } from '../../src/lib/components/loader/Loader.component';
+import { Stack, spacing } from '../../src/lib/spacing';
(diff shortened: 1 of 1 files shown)


Cuervino:
Probably the info message is too large, it should just be what is the deployment name, but not the use case of the deployment name since we don't have it everywhere.

Cuervino:
Closing this. This spec has been redone in the ux-exploration-portal, which is the right place for design specs and interaction explorations (core-ui stays generic, no product-specific content).

The equivalent story lives there as `deployment-name-edit`, with `@scality/core-ui` consumed as an npm dependency. No need to keep this duplicate open in the library repo.

PR #1073, opened 2026-04-15
fix(Tooltip): prevent long words from overflowing tooltip container

JeanMarcMilletScality:
Add rule to wrap a word if it will overflow in the Tooltip:
Example of issue:
image

Result:
image


MonPote (approved):
LGTM

PR #1074, opened 2026-04-17
Improvement/chart modern style

JeanMarcMilletScality:
## Context

New overview page design has a 'modern' look and feel that demand change in our chart style

### Add new props to LineTimesSeriesChart 

`displayOptions` and `preset` to allow the customization of the chart
`displayOptions` allow flexibility  and p`reset` is a way to easily and with a single prop enable the look wanted

### Add labelMap

Chart have a getTooltip to change the display name of the serie but this was not used by chart Legend

### Add color prop for legend

Allow the change of legend color text as it is possible with size

### Default style
image

### Modern style

image

### DIFF
--- a/src/lib/components/charts/linetimeseries/LineTimeSerieChart.tsx
+++ b/src/lib/components/charts/linetimeseries/LineTimeSerieChart.tsx
@@ -56,12 +56,12 @@ export function LineTimeSerieChart({
   yAxisTitle,
   helpText,
   rightTitle,
-  preset = 'default',
+  displayPreset = 'default',
   displayOptions,
   syncId,
   renderTooltip,
 }: LineChartProps) {
-  const presetOptions = CHART_PRESETS[preset];
+  const presetOptions = CHART_PRESETS[displayPreset];
   const resolvedNoBackground = displayOptions?.noBackground ?? presetOptions.noBackground;
   const resolvedShowHorizontalGridLines = displayOptions?.showHorizontalGridLines ?? presetOptions.showHorizontalGridLines;
   const resolvedNoHeader = displayOptions?.noHeader ?? presetOptions.noHeader;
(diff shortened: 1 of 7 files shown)


src/lib/components/charts/linetimeseries/LineTimeSerieChart.tsx
```diff
@@ -58,9 +55,18 @@ export function LineTimeSerieChart({
   yAxisType = 'default',
   yAxisTitle,
   helpText,
+  rightTitle,
+  preset = 'default',
+  displayOptions,
```
ChengYanJin: I think we could have one prop? because IIUC they are control the same thing. 

```  
export const CHART_DISPLAY_PRESETS: Record<string, Required<ChartDisplay>> = {
    default: { background: true,  horizontalGridLines: false, header: true,  yAxisLine: true  },
    modern:  { background: false, horizontalGridLines: true,  header: false, yAxisLine: false },
  };

  // Preset only
  <LineTimeSerieChart display={CHART_DISPLAY_PRESETS.modern} />

  // Preset with override
  <LineTimeSerieChart display={{ ...CHART_DISPLAY_PRESETS.modern, header: true }} />


```

JeanMarcMilletScality: For now the use cases are the default preset or the modern. From previous discussion with Valamir, I have kept the individual options (grouped into displayOptions) to have some flexibility if needed, but there is no use cases for custom display right now. We will either use preset='modern' or nothing (default display). So for me, we should keep the 'preset' prop, but maybe rename to something more explicit ? 
I have added Js doc to make the props clearer and easier to use
@ChengYanJin

ChengYanJin (approved):

PR #1082, opened 2026-04-21
CUI-20 -  Introduce Fira Code font

ChengYanJin:
**Component**:

**Description**:
Core-ui currently only bundles Lato (sans-serif). But we need a monospace font for rendering code (inline code, code blocks), but until now relied on font-family: 'Fira Code', monospace without actually bundling the font — users saw different system monospace fonts depending on their OS (Menlo on Mac, Consolas on Windows). This ensures a consistent monospace font across all users and platforms.

**Design**:
https://fonts.google.com/specimen/Fira+Code

---

Closes: #CUI-20


MonPote (approved):
LGTM

Cuervino:
I'm not really sure we need the bold version. If there is no cost, we can bring it, but I don't see yet the use case.

ChengYanJin:
> I'm not really sure we need the bold version. If there is no cost, we can bring it, but I don't see yet the use case.

OK I will remove it then. Because it will increase the bundle size for nothing. Thank you !

PR #1089, opened 2026-04-23
feat(Modal): structured actions prop, layout constraints & deprecations (CUI-12)

JeanMarcMilletScality:
## TL;DR

Makes the `Modal` component enforce Scality's modal guidelines through its own TypeScript API — a structured `actions` prop replaces free-form footers, the body pins to a 480 px default width, and `role="alertdialog"` can no longer take a `close` handler. Existing call sites keep compiling; `@deprecated` markers point the way to the new API.

## Context

CUI-12 asked us to fold the modal patterns explored in #1033 **into the modal definition itself**, rather than re-documenting them per consumer. The guidelines previously lived only in docs (and briefly in lint rules) — nothing stopped a caller from shipping a modal with no footer, a rich-node title that breaks screen readers, or an `alertdialog` that could be dismissed with a header ✕. This PR moves that enforcement into the type system, where it can't be ignored.

## Approach

The core move is **guideline → type-system constraint**. An earlier iteration added four `valalint` ESLint rules for these; they were dropped once the same rules could be expressed as types (hence the add-then-remove of `valalint/rules/modal-*` in the commit history — none survive in the final diff).

| Guideline | Enforced now by |
|---|---|
| Title is plain text (accessible name) | `title: string \| Exclude<ReactNode, string>` — only the non-string branch is `@deprecated` |
| Every modal has a footer | discriminated union — exactly one of `actions` / `footer` is required |
| Canonical footer layout | `actions` prop; the Modal owns each button's variant + order, callers can't diverge |
| `alertdialog` is not dismissable via ✕ | `close?: never` on the `alertdialog` branch |

The title rule is the subtle one. `string | ReactNode` collapses to just `ReactNode` (string is a subset), so a property-level `@deprecated` struck through *correct* string usage too. Splitting the union fixes the over-fire:

**Before** — deprecation fires on every title, including guideline-correct strings:
```ts
/** @deprecated Pass a plain string. */
title: ReactNode;   // ← string usage struck through too
```
**After** — deprecation fires only on the non-string branch:
```ts
title:
  | string                        // ← guideline-correct, not flagged
  | /** @deprecated Pass a plain string. */ Exclude<ReactNode, string>;
```

## Screenshots

Captured from Storybook → `Components/Feedback/Modal` (`WidthLayoutCases`, `WithExtraAction`):

Modal: 480 px default vs. wide body, and the structured actions footer

## 🔧 Usage

New code uses `actions` — the Modal renders the buttons and fixes their variants. Sourced from the `AttachmentConfirmationModal` migration in this PR:

**Before** — hand-built footer, caller owns layout and variants:
```tsx
import { Icon, Modal, Stack, Wrap } from '@scality/core-ui';

const modalFooter = () => (
  <Wrap>
    <p></p>
    <Stack>
      <Button variant="outline" onClick={handleClose} label="Cancel" />
      <Button icon={<Icon name="Arrow-right" />} variant="primary"
              onClick={attach} label="Confirm" disabled={isLoading} />
    </Stack>
  </Wrap>
);

<Modal close={handleClose} title="Attachment" footer={modalFooter()}>   {/* ← footer */}
  <AttachmentList />
</Modal>
```
**After** — structured `actions`, layout + variants owned by the Modal:
```tsx
import { Modal } from '@scality/core-ui';

<Modal
  close={handleClose}
  title="Attachment"
  actions={{                                                            {/* ← actions */}
    cancel: { label: 'Cancel', onClick: handleClose },
    confirm: { label: 'Confirm', icon: 'Arrow-right', onClick: attach, disabled: isLoading },
  }}
>
  <AttachmentList />
</Modal>
```

## Review focus

- 🔴 **`Modal.component.tsx` › `AlertDialogProps`** — `close?: never` on the `alertdialog` branch. This is the breaking change: any consumer passing `role="alertdialog"` **and** `close` stops compiling after the bump (see Breaking changes).
- 🟡 **`Modal.component.tsx` › `FooterProps`** — the `actions`/`footer` XOR union makes a footer mandatory; a modal passing *neither* now fails `tsc`. Confirm this is the intended floor (opt out with `footer={null}`, which still type-checks and renders no footer bar via the `(actions || footer)` guard).
- 🟡 **`Modal.component.tsx` › `ModalBody`** — body width is now `min(480px, 90vw)` by default. This is a visual change for **every** consumer modal, not just new ones (see Visual regression).

## ⚠️ Breaking changes

1. **`role="alertdialog"` + `close` no longer type-checks.** An alert dialog requires an explicit user choice, so a header-✕ dismissal contradicts the semantic. Fix a flagged call site by either dropping `close` (users dismiss via a footer action such as Cancel), or switching to the default `role="dialog"` if header dismissal was intentional.
2. **A modal must now provide `actions` or `footer`.** `footer` was previously optional; passing neither now fails `tsc`. Opt out explicitly with `footer={null}` (renders no footer bar).

Both are compile-time only — surfaced by `tsc` after bumping core-ui, not silent runtime changes.

## Visual regression (not breaking)

Otherwise-valid call sites keep compiling and running, but modals may need a small adjustment after the bump:

- The body is now pinned to **480 px**. Modals that previously sized to wider content (large tables, multi-column forms) will look cramped or clipped — add **`wide`** to those.
- Modals narrower than 480 px (short confirmations) will render slightly wider — usually an improvement, no change required.
- Footer layout is unchanged for code still on `footer`. Migrating to `actions` gets the canonical layout for free.

Watch your editor for `@deprecated` strike-throughs on `footer` and non-string `title` — those are the migration points, no failing build required.

## How to test

- `npm test` · `npm run lint` · `tsc --noEmit` all clean.
- Storybook → `Components/Feedback/Modal`: check `WidthLayoutCases`, `WithExtraAction`, `DestructiveWithAcknowledgement`, and `LegacyFooterProp` render correctly in light + dark.
- After bumping core-ui in a consumer repo, run `tsc` and fix any surfaced `role="alertdialog"` + `close` call sites; scan the editor for `@deprecated` `footer` / non-string `title` usages.

## References

- [CUI-12](https://scality.atlassian.net/browse/CUI-12) — analyse what from #1033 can be pulled directly into the modal definition; this PR is that adoption.
- #1033 — the source PR whose modal patterns this consolidates.

<details>

- **`Modal.component.tsx`** — new `actions: { confirm, cancel?, extra? }` prop (Modal renders each button, fixes variants: `confirm`→primary/danger, `cancel`→outline, `extra`→secondary; right-aligned, `confirm` rightmost); `footer` kept + `@deprecated`; `title` non-string branch `@deprecated`; `alertdialog` XOR `close`; body `min(480px, 90vw)` with `wide` opt-out and `overflow-x: auto`; `useId()` for the `aria-labelledby`/`aria-describedby` ids (was hard-coded, collided across stacked modals); ESC handler split into its own effect keyed on `[isOpen, close]`.
- **`src/lib/index.ts`** — re-export `ModalAction` / `ModalActions` types for consumers.
- **`AttachmentConfirmationModal.tsx`** — migrated from `footer` to `actions`, dropped the dead `modalFooter` helper (the sole in-repo consumer that used the deprecated path).
- **`stories/Modal/`** — stories moved into a per-component folder and rewritten to `actions`; added `WidthLayoutCases`, `WithExtraAction`, `DestructiveWithAcknowledgement`, and a `LegacyFooterProp` reference story. The `valalint` modal rules added in early commits were removed once the type system covered them.

</details>

[CUI-12]: https://scality.atlassian.net/browse/CUI-12?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/lib/components/modal/Modal.component.tsx
```diff
@@ -82,57 +210,87 @@ const Modal = ({
...
+    //Listen to esc key to close the modal
+    const handleEsc = (event: KeyboardEvent) => {
+      if (event.key === 'Escape') {
+        close?.();
+      }
+    };
+    document.addEventListener('keydown', handleEsc);
+    return () => {
+      document.removeEventListener('keydown', handleEsc);
+    };
```
JBWatenbergScality, line 229: nit: it's super minor but maybe we shouldn't register any event listener if the close function is not available

JBWatenbergScality (approved):

PR #1140, opened 2026-06-23
fix(navbar): keep user dropdown on one line and reveal navbar border

JeanMarcMilletScality:
## Problem

A long username in the navbar's right-action dropdown wrapped to two lines and grew **taller than the navbar**, breaking vertical alignment. The trigger's solid `navbarBackground` fill also covered the navbar's bottom border.

## Fix

In `NavbarMenuItem .sc-dropdown .trigger` (`Navbar.component.tsx`):
- Pin the trigger to `navbarHeight` and add `white-space: nowrap` so it never wraps.
- Truncate the label (`.sc-trigger-text`): `display: block; max-width: 12rem; overflow: hidden; text-overflow: ellipsis`.
- Remove the `background-color: navbarBackground` fill so the navbar's bottom border stays visible.

Scope matches the original bug — it applies to `Dropdown`-type right actions (the username case), whose trigger text carries `.sc-trigger-text`.

## Verification

- `NavbarWithLongUserName` Storybook story added for visual review.
- Full Jest suite: 36 suites / 454 tests pass.
- `tsc --noEmit` clean; ESLint clean.

### DIFF
--- a/src/lib/components/navbar/Navbar.component.tsx
+++ b/src/lib/components/navbar/Navbar.component.tsx
@@ -135,12 +135,19 @@ const NavbarMenuItem = styled.div`
   align-items: center;
   .sc-dropdown {
     .trigger {
-      background-color: ${getThemePropSelector('navbarBackground')};
       &:hover {
         background-color: ${getThemePropSelector('highlight')};
       }
-      height: auto;
+      height: ${navbarHeight};
+      white-space: nowrap;
       font-size: ${fontSize.base};
+      .sc-trigger-text {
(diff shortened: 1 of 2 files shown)


damiengillesscality (approved):
12 rem is near 200px I think it is still too much and that vw would probably be a better unit for this use case. Still approved as it is a clear improvement.

JeanMarcMilletScality:
Reopened as #1148 with the correct `bugfix/` branch prefix. Closing this one.

PR #1144, opened 2026-06-25
feat(responsive): add useContainerWidth hook

JeanMarcMilletScality:
## What

First of a progressive series merging the responsive-panels exploration into `development/1.0`. This PR adds **only the foundational hook** — `useContainerWidth` — plus its unit tests. Nothing else from the exploration branch ( column-drop, icon-only buttons, fluid forms, navbar/table changes) is included here.

## Details

`useContainerWidth` is a **container-width** (not viewport) responsiveness primitive, for screens that must stay usable when a host side-panel shrinks the available space without changing the viewport width.

- Callback-ref `ResizeObserver`; returns ref and width.
- Optional `hysteresis` band to avoid flicker when the container rests on a breakpoint.

## Compatibility

Purely additive / opt-in — no existing component behaviour changes.

## Verification

- `tsc --noEmit` clean
- `eslint` clean on changed files
- `jest` — 9 unit tests covering wide-first-before-measure, breakpoint flips in both directions, width rounding, custom breakpoint, the hysteresis band, and detaching the observed node

### DIFF
--- a/src/lib/components/responsive/useContainerWidth.ts
+++ b/src/lib/components/responsive/useContainerWidth.ts
@@ -1,62 +1,31 @@
 import { useCallback, useRef, useState } from 'react';
 
-/**
- * Container width (in px) below which a screen should switch to its compact
- * layout. The trigger is the *container* width, not the viewport, so screens
- * stay usable when a host side-panel shrinks the available space without
- * changing the viewport width.
- */
-export const NARROW_BREAKPOINT_PX = 640;
-
-/**
- * Tables carry several columns and a search/action toolbar, so they feel
- * cramped at a wider width than a single-column detail form does. List screens
- * pass this higher breakpoint to drop secondary columns / shrink the toolbar
- * earlier than detail panels switch to their compact layout.
- */
-export const TABLE_NARROW_BREAKPOINT_PX = 820;
-
-export type UseContainerWidthOptions = {
-  /**
-   * Hysteresis band (px). Once narrow, the container must grow back to
-   * `breakpoint + hysteresis` before it is considered wide again. Prevents
-   * flicker when the container is dragged to rest right on a breakpoint.
-   * Defaults to 0 (no band).
-   */
-  hysteresis?: number;
-};
-
 export type UseContainerWidthResult<T extends HTMLElement> = {
   /** Callback ref — attach to the element whose width should drive the layout. */
   ref: (node: T | null) => void;
-  /** Latest measured border-box width in px, or `null` before first measure. */
-  width: number | null;
-  /** True when `width` is below the hook's `breakpoint`. Wide-first until measured. */
-  isNarrow: boolean;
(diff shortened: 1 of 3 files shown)


src/lib/components/responsive/useContainerWidth.ts
```diff
@@ -0,0 +1,107 @@
...
+      // Border-box width (matches the element's CSS width and the sync read
+      // below); where borderBoxSize is unsupported, re-measure the border box
+      // from the target rather than fall back to contentRect's content box,
+      // which would disagree with the sync read on padded elements.
+      const observedWidth =
+        entry?.borderBoxSize?.[0]?.inlineSize ??
+        entry?.target.getBoundingClientRect().width;
+      if (typeof observedWidth === 'number') {
+        const rounded = Math.round(observedWidth);
+        setWidth((prev) => (prev === rounded ? prev : rounded));
```
damiengillesscality: strictly equivalent to the simpler setWidth(rounded) for number values

damiengillesscality (changes requested):
I don't vibe too much with this. The signature of `useContainerWidth` is a bit strange, `breakpoint` does not seem like a parameter I would expect for such function. Since we use this hook to monitor some container width, why would the default 640px breakpoint be relevant for most of our containers ? The choice of returning isNarrow and isNarrowerThan is odd too. Is isNarrowerThan is like admitting breakpoint parameter is irrelevant. And isNarrow is not that useful I would rather write a more explicit `(width ?? 0) < breakpoint` in my main code. On that, I will add a small picky note undefined is better than null for no width value both semantically and in the hypothesis a core-ui user would not use typescript since `undefined < number` is always false which is not true with `null`.
Regarding the use cases is it also necessary to export this or could it be contained in core-ui ? It could be better to not expend the not design system part of our api imho.
Have you considered the very good and stable `use-resize-observer` ?
And one last thing have you updated your knowledge on @container css queries ? They are well supported and can probably replace js in some of your use cases.

JeanMarcMilletScality:
Thanks for the review
Agree on the hook signature. So the hook only responsability will be to provide the width of the container (removing isNarrow and isNarrowerThan) and will not be exposed. 
I did not know about `use-resize-observer`, I'm not sure it is worth it to add this dep ?

JeanMarcMilletScality:
Nice catch for the `null` issue
For @container, yes I looked at it (Tabs is using it already), I can use it for the Button, the only drawback will be the Tooltip always on.
But for the Navbar and Table, JS is needed.

damiengillesscality:
> Thanks for the review Agree on the hook signature. So the hook only responsability will be to provide the width of the container (removing isNarrow and isNarrowerThan) and will not be exposed. I did not know about `use-resize-observer`, I'm not sure it is worth it to add this dep ?

Imho yes, but not a hard requierement we can recode it if you see value in it.

damiengillesscality (approved):

PR #1151, opened 2026-07-02
improvement(responsive): add dropAt column drop to Tablev2

JeanMarcMilletScality:
## TL;DR

`Tablev2` columns can now opt into responsive hiding: set `dropAt: <px>` on a column and it hides itself once the table gets narrower than that, with an optional `+N` popover that keeps the hidden values reachable. Both features are opt-in — tables that don't use them are unchanged.

## Context

Second step in the progressive responsive rollout, after the `useContainerWidth` measure-only hook landed in #1144. This PR is what consumes that hook: the Table observes its own `TableWrapper` width and hides columns whose `dropAt` breakpoint the container has fallen below — no consumer wiring, no context/provider, the Table self-measures.

The design rule being followed is **restyle → CSS; count/relocate DOM → JS**. Plain column hiding could have been CSS, but the reveal popover has to *count* the hidden columns and *relocate* their values, which CSS can't do — so the whole feature stays JS-driven on the table's own measured width to keep one source of truth. react-table keeps hidden cells on `row.allCells`, so the dropped values stay reachable for the popover without duplicating data.

How a resize turns into what you see:

```mermaid
flowchart LR
  A[TableWrapper width<br/>via useContainerWidth] --> B{width &lt; column.dropAt?}
  B -- yes --> C[hide column<br/>setHiddenColumns]
  C --> D{revealDroppedColumns<br/>&& any dropped?}
  D -- yes --> E[show trailing +N column<br/>popover lists Header: value]
  D -- no --> F[no reveal column]
  B -- no --> G[column stays visible]
```

## Screenshots

**`ResponsiveColumnDrop` — wide vs. narrow.** All columns show when wide; drop the container below 550px and both `Age` (dropAt 550) and `Last Name` (dropAt 700) hide, while `First Name` and `Health` (no `dropAt`) stay.

Tablev2 responsive column drop — wide vs. narrow

**`ResponsiveColumnDropWithReveal` — `+N` popover.** With `revealDroppedColumns`, a trailing per-row `+2` trigger opens a popover listing the dropped columns' `Header: value` pairs, so no data is lost on a narrow viewport.

Tablev2 reveal popover showing dropped column values

## 🔧 Usage

Opt in per column with `dropAt`, and (optionally) turn on the reveal popover at the Table level. Both are additive — omitting them keeps today's behaviour.

```tsx
// before — static columns
const columns = [
  { Header: 'First Name', accessor: 'firstName' },
  { Header: 'Last Name',  accessor: 'lastName' },
  { Header: 'Age',        accessor: 'age' },
];
<Table columns={columns} data={data} getRowId={getRowId}>
  <Table.SingleSelectableContent rowHeight="h40" />
</Table>

// after — responsive columns + reveal popover
const columns = [
  { Header: 'First Name', accessor: 'firstName' },              // no dropAt → always shown
  { Header: 'Last Name',  accessor: 'lastName', dropAt: 700 👈 },  // hidden below 700px
  { Header: 'Age',        accessor: 'age',      dropAt: 550 👈},  // hidden below 550px
];
<Table columns={columns} data={data} getRowId={getRowId} revealDroppedColumns>👈
  <Table.SingleSelectableContent rowHeight="h40" />
</Table>
```

## Review focus

- 🟡 **`Tablev2.component.tsx` › responsive `useEffect`** — the core reconciliation. It reconciles on either a changed drop-set *or* a runtime `revealDroppedColumns` toggle, and `setHiddenColumns` is written to preserve columns a consumer hid manually (only releases ids this effect itself dropped). Worth confirming the manual-hide preservation and the toggle path hold up.
- 🟡 **`useDroppedColumns.tsx` › `visibleColumns` / `visibleColumnsDeps`** — new react-table plugin injecting the synthetic `+N` column *only* when `revealDroppedColumns` is on (gated via a threaded `useTable` option), so feature-off tables keep an unchanged column pipeline. Check the gating + the popover reading dropped values off `row.allCells`.
- 🟡 **`TableCommon.tsx` › `columnsKey` / `itemData`** — virtualization sync. react-table keeps the `rows` array identity stable on visibility-only changes, so `columnsKey` (derived from visible `headerGroups`) gives `itemData` a fresh identity to force react-window past `areEqual`; otherwise the header updates while the body keeps stale cells.

## How to test

1. `npm run storybook`, open **Tablev2 › ResponsiveColumnDrop**.
2. Drag the frame's bottom-right corner to resize: **Age** drops below 550px, **Last Name** below 700px; **First Name** and **Health** (no `dropAt`) always stay.
3. Open **ResponsiveColumnDropWithReveal**, narrow below 700px until the trailing **`+N`** column appears, click a row's trigger, and confirm the popover lists the dropped columns' `Header: value` pairs for that row.
4. Sanity check a table with no `dropAt` columns is visually and behaviourally unchanged.

## References

- **#1144** — `useContainerWidth` measure-only hook; the width source this PR consumes (step 1 of the responsive rollout).

<details>

- **`useDroppedColumns.tsx`** (new) — react-table plugin: the synthetic `+N` column and its floating-ui popover cell; injected only when `revealDroppedColumns` is opted in.
- **`Tablev2.component.tsx`** — `dropAt` measuring effect, `revealDroppedColumns` prop, `droppedColumnIds` on context, and the plugin registration/gating.
- **`TableCommon.tsx`** — `columnsKey` threaded into `VirtualizedRows` to re-sync the virtualized body with the header on visibility changes.
- **`react-table-config.ts`** — types for `dropAt` (column) and `revealDroppedColumns` (table option).
- **`Tablev2.test.tsx`** — 6 new behavioural tests (jsdom stubs `ResizeObserver` + `getBoundingClientRect`).
- **`tablev2.stories.tsx`** — `ResponsiveColumnDrop` and `ResponsiveColumnDropWithReveal` stories (resizable frames).

</details>

### DIFF
--- a/src/lib/components/tablev2/useDroppedColumns.tsx
+++ b/src/lib/components/tablev2/useDroppedColumns.tsx
@@ -1,0 +1,161 @@
+import { useState } from 'react';
+import { Hooks, Row } from 'react-table';
+import styled from 'styled-components';
+import {
+  autoUpdate,
+  flip,
+  FloatingFocusManager,
+  FloatingPortal,
+  offset,
+  shift,
+  useClick,
+  useDismiss,
+  useFloating,
+  useInteractions,
+  useRole,
+} from '@floating-ui/react';
+import { spacing } from '../../spacing';
+import { fontSize, fontWeight, zIndex } from '../../style/theme';
+import { getThemePropSelector } from '../../utils';
+import { Icon } from '../icon/Icon.component';
+import { useTableContext } from './Tablev2.component';
+
+/**
+ * Id of the synthetic trailing column injected by {@link useDroppedColumns}. It
+ * starts hidden (listed in the table's `initialState.hiddenColumns`) and is
+ * revealed by the responsive effect only while at least one `dropAt` column is
+ * currently dropped and `revealDroppedColumns` is enabled.
+ */
+export const DROPPED_COLUMNS_COLUMN_ID = 'droppedColumns';
+
+const Container = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  height: 100%;
+  width: 100%;
+`;
+
+const TriggerButton = styled.button`
+  display: inline-flex;
+  align-items: center;
+  gap: ${spacing.r4};
+  height: 100%;
+  padding: 0 ${spacing.r8};
+  border: none;
+  background: transparent;
+  cursor: pointer;
+  color: ${getThemePropSelector('textPrimary')};
+  &:hover {
+    background-color: ${getThemePropSelector('highlight')};
+  }
+`;
+
+const Panel = styled.div`
+  min-width: 14rem;
+  max-width: 24rem;
+  padding: ${spacing.r4} 0;
+  background-color: ${getThemePropSelector('backgroundLevel1')};
+  border: 1px solid ${getThemePropSelector('border')};
+  z-index: ${zIndex.dropdown};
+`;
+
+const Field = styled.div`
+  display: flex;
+  flex-direction: column;
+  gap: ${spacing.r2};
+  padding: ${spacing.r8} ${spacing.r16};
+  color: ${getThemePropSelector('textPrimary')};
+`;
+
+const FieldLabel = styled.span`
+  font-size: ${fontSize.small};
+  color: ${getThemePropSelector('textSecondary')};
+`;
+
+const FieldValue = styled.div`
+  font-weight: ${fontWeight.base};
+`;
+
+const DroppedColumnsCell = ({ row }: { row: Row<object> }) => {
+  const { droppedColumnIds } = useTableContext();
+  const [isOpen, setIsOpen] = useState(false);
+  const { refs, floatingStyles, context } = useFloating({
+    open: isOpen,
+    onOpenChange: setIsOpen,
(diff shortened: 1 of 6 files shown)


src/lib/components/tablev2/useDroppedColumns.tsx
```diff
@@ -0,0 +1,165 @@
...
+import { Text } from '../text/Text.component';
+import { useTableContext } from './Tablev2.component';
+
+/**
+ * Id of the synthetic trailing column injected by {@link useDroppedColumns}. It
+ * starts hidden (listed in the table's `initialState.hiddenColumns`) and is
+ * revealed by the responsive effect only while at least one `dropAt` column is
+ * currently dropped and `revealDroppedColumns` is enabled.
+ */
+export const DROPPED_COLUMNS_COLUMN_ID = 'droppedColumns';
```
JBWatenbergScality, line 30: is there a pb if there is 2 table rendered on the same page ? id will conflict with each others no ?

JeanMarcMilletScality: The id is used by react-table to identify the column within the table, no collusion with other tables . the id only need to be unique per table

JBWatenbergScality (approved):

PR #1152, opened 2026-07-03
ARTESCA-17791: Navbar responsive — tab overflow "More" menu + condense actions

JeanMarcMilletScality:
**TL;DR** — The Navbar now measures its own width and adapts: tabs that no longer fit collapse into a "More" menu, and right-action labels condense (icon-only, or initials for the account menu) as space runs out — without a viewport media query.

### Context / Why
Responsive PR2 of the JS-measurement track, building on the measure-only `useContainerWidth` hook (PR1). Tab overflow and label condensing are the *one* responsiveness case CSS `@container` can't do — they require counting and relocating DOM by cumulative width — so this is done in JS. The goal is to keep the navbar usable when a host **side-panel shrinks the available space** (not the viewport).

Worth knowing before you dive in: the ticket flags this as the **highest-risk PR in the responsive track** — Navbar is a widely-used, already-shipped component, and this PR now routes **every** existing navbar through the new measure-and-overflow path, even consumers that set none of the new props. The new API surface itself is opt-in and non-breaking (only new *optional* props: `condenseActionsBreakpoint`, `Tab.alwaysVisible`, `DropdownAction.abbreviateWhenCondensed`), but the render-path change is not scoped to opt-in users.

### 📷 Screenshots
Wide vs. narrow container, captured from the PR's own `ResponsiveTabOverflow` story:

Wide container: all tabs inline, full name shown. Narrow container: tabs overflow into "More", account label condenses to initials.

- **Wide (1300px)**: all 5 tabs inline, full "Carlito Gonzalez" name shown.
- **Narrow (720px)**: "My instance" (custom render tab) and "Groups" (selected tab) stay pinned inline; the rest overflow into "More". The account dropdown condenses to "CG".

### 🔧 Usage
Non-breaking, opt-in — existing call sites compile unchanged. Real before/after from the PR's own diff (`stories/navbar.stories.tsx`), the only call site it touches:
```tsx
// before
{
  type: 'dropdown',
  text: 'Carlito',
  icon: <i className="fas fa-user" />,
  items: [{ label: 'Log out', onClick: () => {} }],
}

// after — opts into initials instead of hiding the label when condensed
{
  type: 'dropdown',
  text: 'Carlito Gonzalez',
  icon: <i className="fas fa-user" />,
  abbreviateWhenCondensed: true,
  items: [{ label: 'Log out', onClick: () => {} }],
}
```
`condenseActionsBreakpoint` (on `<Navbar>`) and `Tab.alwaysVisible` are also new, but no call site in this diff exercises them — not shown, per the never-invent rule.

### 🔍 Review focus
- 🔴 `Navbar.component.tsx › NavbarTabsRow` + `selectVisibleTabs()` — the priority+ fit is the heart of the PR and every navbar now goes through it. `selectVisibleTabs` is pure and well unit-tested (fit / pinning / order); the wiring around it in `NavbarTabsRow` (per-index width cache + measurement) is where a regression would actually ship. Look here first.
- 🟡 `Navbar.component.tsx › useCondensedActions()` — condense latch lives in a `useLayoutEffect` (not a render side-effect, so StrictMode/concurrent-safe) with a ±24px hysteresis band to stop flicker at the breakpoint. Confirm the wide-first→condensed→wide transitions behave.
- 🟡 Accessibility surface — overflow panel is a `<nav>` disclosure, **not** an ARIA `menu` (deliberate: no arrow-key contract for what are just links); `Dropdown.component.tsx` gains an `aria-label` prop applied *after* the action spread so a stray prop can't clobber the condensed name; menu items use `aria-current` instead of `role=tab`. Verify the condensed account menu still announces its full name.

> ⚠️ **Watch for** (each tied to a specific hunk):
> - **Stale width cache.** Widths are cached by tab **index** and only refreshed for tabs currently in the row (a tab in "More" keeps its old cached width). Correct while tabs are stable, but dynamic labels (i18n) or reordering would measure against a stale width. The code documents a `key`-bump workaround — but a consumer won't know to do that unless it's called out.
> - **Condensed-name precedence.** `getActionRenderer` spreads `...rest` from the (porous) action object, then applies `title`/`aria-label` after it. A stray `aria-label` on an action must not win over the condensed name — there's a test for exactly this boundary; confirm it still holds if the spread order changes.
> - **First-paint flash.** A 96px estimate is reserved for the "More" trigger until it self-measures. Confirm there's no one-frame flash of one-too-many tabs on initial mount.

### 🧪 How to test
1. Storybook → **Components/Navigation/Navbar → ResponsiveTabOverflow**.
2. Drag the container's right edge inward: tabs collapse from the right into the "More" menu; the selected tab and the instance-name field stay pinned inline.
3. Narrow further past the condense breakpoint: right-action labels drop to icon-only; the account dropdown shows initials ("CG") instead — hovering / assistive tech still gives the full name.
4. Keyboard-only: Tab to the "More" trigger, open it, Tab through the links (no arrow keys needed); the focus ring should match the inline tabs.
5. `npm test src/lib/components/navbar` → 13 passing.

### 🔗 References
- [ARTESCA-17791](https://scality.atlassian.net/browse/ARTESCA-17791) — [Responsive PR2] Navbar responsive: tab overflow + condense actions (Story, Ongoing; flagged highest-risk in the track).
- Depends on **PR1** — the measure-only `useContainerWidth` hook.
- Design: `docs/responsive-design-proposal.md` (§ "What stays JS", Track A step 2).

<details><summary>What changed & deliberately not changed</summary>

**Changed**
- `Navbar.component.tsx` (+479/-40): priority+ tab overflow (`selectVisibleTabs`, `NavbarTabsRow`, `NavbarTabsMenu`), `useCondensedActions` hysteresis helper, `getInitials`, condensed action rendering.
- `Dropdown.component.tsx` (+13): `aria-label` prop + icon-only fallback to `title`.
- `Navbar.component.test.tsx` (new, +289): unit tests for `selectVisibleTabs`/`getInitials` + behavior tests for overflow, pinning, condense, a11y names.
- `navbar.stories.tsx` (+27): `ResponsiveTabOverflow` story.

**Deliberately not changed** (author's notes, preserved — these answer likely review questions up front)
- Kept the existing Dropdown text-swap condense mechanism → **no** dependency on the Button `iconOnly` `@container` work (PR4).
- No `ResponsiveContainer`/provider — the navbar measures itself directly.
- The `useContainerWidth` hook stays internal (not exported from `index.ts`).
- Opt-in / non-breaking API: only new optional props.

</details>

[ARTESCA-17791]: https://scality.atlassian.net/browse/ARTESCA-17791?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/components/navbar/Navbar.component.tsx
+++ b/src/lib/components/navbar/Navbar.component.tsx
@@ -1,12 +1,123 @@
-import { cloneElement, Fragment } from 'react';
+import {
+  cloneElement,
+  useCallback,
+  useLayoutEffect,
+  useRef,
+  useState,
+} from 'react';
 import styled, { css } from 'styled-components';
+import { flip, offset, shift } from '@floating-ui/dom';
+import {
+  autoUpdate,
+  FloatingFocusManager,
+  FloatingPortal,
+  useClick,
+  useDismiss,
+  useFloating,
+  useInteractions,
+  useRole,
+} from '@floating-ui/react';
 import { Logo } from '../../icons/branding';
 import { spacing } from '../../spacing';
-import { fontSize, navbarHeight, navbarItemWidth } from '../../style/theme';
+import {
+  fontSize,
+  fontWeight,
+  navbarHeight,
+  navbarItemWidth,
+  zIndex,
+} from '../../style/theme';
 import { getContrastText, getThemePropSelector } from '../../utils';
 import { Dropdown, type Item } from '../dropdown/Dropdown.component';
 import { Icon } from '../icon/Icon.component';
 import { Button, FocusVisibleStyle, type Props as ButtonProps } from '../buttonv2/Buttonv2.component';
+import { useContainerWidth } from '../responsive/useContainerWidth';
+
+/**
+ * Default container width (px) below which the navbar condenses its
+ * right-action labels to icon-only. Content-dependent — override per app via
+ * the `condenseActionsBreakpoint` prop. Tabs overflow into a "More" menu
+ * automatically as space runs out, independent of this value.
+ */
+export const NAVBAR_COLLAPSE_BREAKPOINT_PX = 768;
+
+/**
+ * Hysteresis band (px) for the condense toggle. Once condensed, the navbar must
+ * grow back to `breakpoint + this` before the labels return, so a container
+ * dragged to rest right on the breakpoint doesn't flicker.
+ */
+const CONDENSE_HYSTERESIS_PX = 24;
+
+/**
+ * Wraps the measure-only width primitive with a hysteresis band and reports a
+ * single condense flag. Wide-first until the first measurement lands
+ * (`undefined < n` is `false`), then condensed once the container drops below
+ * `breakpoint`, staying condensed until it grows back past
+ * `breakpoint + CONDENSE_HYSTERESIS_PX`. This is the one place the navbar needs
+ * a threshold, so the hysteresis lives here rather than in the shared hook.
+ */
+function useCondensedActions(breakpoint: number): {
+  ref: (node: HTMLDivElement | null) => void;
+  condensed: boolean;
+} {
+  const { ref, width } = useContainerWidth<HTMLDivElement>();
+  const condensedRef = useRef(false);
+
+  let condensed = condensedRef.current;
+  if (width !== undefined) {
+    if (condensed) {
+      if (width >= breakpoint + CONDENSE_HYSTERESIS_PX) condensed = false;
+    } else if (width < breakpoint) {
+      condensed = true;
+    }
+    condensedRef.current = condensed;
+  }
+
+  return { ref, condensed };
+}
+
+/**
+ * Decide which tabs stay inline and which move into the "More" menu.
+ *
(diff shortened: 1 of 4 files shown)


src/lib/components/navbar/Navbar.component.tsx
```diff
@@ -16,6 +157,12 @@ type DropdownAction = {
   items: Array<Item>;
   text?: string;
   icon?: JSX.Element;
+  /**
+   * When the navbar condenses, show the label's initials next to the icon
+   * (e.g. "Carlito Gonzalez" → "CG") instead of hiding it. The full label
+   * stays the accessible name. Requires an `icon`. Defaults to false.
+   */
+  abbreviateWhenCondensed?: boolean;
```
JBWatenbergScality: As our design system is opinoinated in my mind such choice should be "hard coded" in core-ui not an opt-in choice

JBWatenbergScality (approved):

PR #1159, opened 2026-07-16
improvement(responsive): opt-in container prop on Box/Stack/Wrap + TwoPanelLayout

JeanMarcMilletScality:
## TL;DR
Adds an **opt-in `container` prop** that turns a layout element into a CSS `@container` query container named `responsive`, so descendant components can restyle based on that element's inline width — with no JS and no extra wrapper node.

## Context
This is the foundation the responsive Button (#1160) and Form (#1161) work builds on: those components write `@container responsive (…)` rules, which need an ancestor that actually *establishes* a container named `responsive`. This PR is the only place that establishes it. Independent, mergeable on its own.

## Approach
Two hosts, one fixed contract:
- **Box/Stack/Wrap** get `container?: boolean`. `Stack`/`Wrap` render a `Box`, so one prop covers all three.
- **TwoPanelLayout** gets `container?: 'left' | 'right' | 'both'` — panels are the highest-value host because they're what actually shrinks when the layout narrows.

The container **name is fixed to `responsive`** (not exposed as a prop) — it's the contract querying components rely on. Resolution is **nearest-wins**: a `@container responsive (…)` rule binds to the closest ancestor named `responsive`, so wrapping a subtree in `<Stack container>` declares a new responsive frame for its children. This is intended; exposing the name is a documented, non-breaking future escape hatch if nesting ever needs it.

```
<TwoPanelLayout container="right">   ← establishes `responsive`
  └─ RightPanel
       └─ <Button iconOnly={480}/>   ← queries nearest `responsive` = the panel
```

## 🔧 Usage
```tsx
// Before — a plain layout
<Stack>{children}</Stack>

// After — opt into being a @container named "responsive"
<Stack container>{children}</Stack>            // ←

// Panels: make a specific panel the query container
<TwoPanelLayout
  panelsRatio="50-50"
  container="right"                             // ←
  leftPanel={{ children: <LeftContent /> }}
  rightPanel={{ children: <RightContent /> }}
/>
```

## Review focus
- 🟡 `layout/v2/panels.tsx › LeftPanel/RightPanel` — the panel container **must stay opt-in**. `container-type: inline-size` makes an element stop sizing to its content on the inline axis, which would collapse a content-sized (`flex: 0 auto`) panel; ratio panels (`50-50`/`65-35`/`30-70`) size from flex and are safe.
- ⚪ `box/Box.ts › shouldForwardProp` — styled-components v6 forwards unknown props to the DOM, so `container` is explicitly filtered (`prop !== 'container'`). Guarded by a DOM-leak regression test.
- ⚪ Fixed name `responsive` + nearest-wins is a deliberate contract, not an oversight.

## How to test
- Wrap any component in `<Stack container>` (or set `container` on a panel) and confirm children still render unchanged.
- Storybook (deferred): drop a responsive Button/Form inside a `container` panel and resize the panel — the descendant should react to the panel's width. Not observable in jsdom.

## References
- **[CUI-31](https://scality.atlassian.net/browse/CUI-31)** (#1160) — Button `iconOnly`; the first consumer of this container.
- **[CUI-32](https://scality.atlassian.net/browse/CUI-32)** (#1161) — Form `flipAt` self-hosts a `responsive` container using the same contract.

<details>

`Box` gains a `$container`-style boolean emitting `container-type: inline-size; container-name: responsive`, filtered from the DOM via `shouldForwardProp`. `Stack` forwards `container` to `Box` through its rest-spread; `Wrap` inherits it via its `BoxComponentProps` type (no code change). `TwoPanelLayout`'s `LeftPanel`/`RightPanel` take a transient `$container` derived from the `'left' | 'right' | 'both'` prop.
</details>

[CUI-31]: https://scality.atlassian.net/browse/CUI-31?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/components/box/Box.test.tsx
+++ b/src/lib/components/box/Box.test.tsx
@@ -1,6 +1,7 @@
 import { render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { Box } from './Box';
+import { Stack } from '../../spacing';
 
 describe('Box', () => {
   it('keeps design-system style props out of the rendered HTML', () => {
@@ -50,5 +51,22 @@ describe('Box', () => {
   it('renders as the element given by the "as" prop', () => {
     render(<Box as="section" data-testid="b" />);
     expect(screen.getByTestId('b').tagName).toBe('SECTION');
+  });
+});
+
+describe('Box container', () => {
+  it('still shows its children when opted into being a responsive container', () => {
+    render(
+      <Stack container>
+        <span>panel content</span>
+      </Stack>,
+    );
+    expect(screen.getByText('panel content')).toBeVisible();
+  });
+
+  it('does not emit the container prop as a DOM attribute', () => {
+    render(<Box container data-testid="c" />);
+    const el = screen.getByTestId('c');
+    expect(el.hasAttribute('container')).toBe(false);
   });
 });
--- a/src/lib/components/box/Box.ts
+++ b/src/lib/components/box/Box.ts
@@ -1,4 +1,4 @@
-import styled from 'styled-components';
+import styled, { css } from 'styled-components';
 import shouldForwardProp from '@styled-system/should-forward-prop';
 import {
   layout,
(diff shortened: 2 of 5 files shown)


damiengillesscality (approved):
Interesting, I kinda like it gj

PR #1161, opened 2026-07-16
improvement(responsive): Input fluid + Form responsive (CUI-32)

JeanMarcMilletScality:
**TL;DR** — Adds an opt-in `responsive` prop to `Form` (and `fluid` to `Input`) so a form adapts in place to the width it's given — fields shrink to fit and the layout auto-flips to stacked when too narrow. Reworks `FormSection` layout to be **`forceLabelWidth`-driven** (self-contained rows when set, auto-sized `subgrid` when unset) and drops the old JS label-measurement.

## Context

Forms in narrow panels/drawers either overflow or need a separate mobile layout. [CUI-32](https://scality.atlassian.net/browse/CUI-32) asked for a form that adapts to its container. Opt-in and non-breaking: without `responsive`, existing forms render as before (see the behavior-change note below for the caveats).

## Approach

**Layout is `forceLabelWidth`-driven, not always `subgrid`.** How a `FormSection` aligns its labels — and how tolerant it is of nesting — depends on whether `forceLabelWidth` is set. This is the core of the design and the reason the old JS label-measurement is gone:

| | engine | label alignment | nesting |
|---|---|---|---|
| **`forceLabelWidth` set** (fixed px) | each `FormGroup` is a **self-contained** 2-col grid (`<px> <field>`), rows stacked in a flex column | shared fixed px | **robust** — a group inside a `div` / `Stack` / `Accordion` stays aligned; each row carries its own column definition |
| **`forceLabelWidth` unset** (auto) | section is a grid, each `FormGroup` a `subgrid` row | auto-sized to widest label, shared via `subgrid` | requires **direct child** (a `subgrid` constraint) — a nested group loses the row and stacks under its label |

The auto (unset) branch adds two grid rules so a section is well-behaved beyond plain `FormGroup`s:
- field track is `minmax(max-content, 1fr)` (not bare `max-content`), so the grid **fills its container** instead of hugging content — otherwise a `grid-column: 1 / -1` child only spans to the last content-sized line, not the edge.
- `& > * { grid-column: 1 / -1 }`, so non-`FormGroup` children dropped into a section (a `Banner`, `InfoMessage`, explanatory `Text`) span the **full section width** in both modes.

**`responsive` is orthogonal** — it doesn't change which strategy above is used, only the field track and whether a section can flip:

| | field track | flip |
|---|---|---|
| **non-responsive** | `max-content` (auto) / field size (fixed) — hugs content | none |
| **responsive** | `minmax(10rem, 1fr)` — fluid | `@container` below `calc(24ch + 10rem + 2rem)` |

**The flip is content-derived — there is no breakpoint prop.** A section stacks once it can no longer give its columns a readable minimum: field floor **10rem** (the `1/2` Input size), label floor **~24ch**. It keys off the Form's *own* width via `container-type: inline-size`, not the viewport.

**Behavior change on the stable (non-responsive) path — what to verify.** Moving non-responsive forms onto the grid engine has visible consequences. All are believed visually equivalent for typical forms, but they touch *every* existing form:
- **Fields share one column** (auto branch, sized to the widest field) instead of each hugging its own content. Identical for all-`Input` forms; width-less content now sits in the shared column.
- `forceLabelWidth` is now a **hard cap** (a label wider than it wraps), where before it was a floor the measurement could grow past.
- A **vertical** `FormGroup` renders a full-row-width stacked label (was constrained to the measured label column).
- An **auto-label section (`forceLabelWidth` unset) with a `FormGroup` nested inside a wrapper** stacks that group instead of aligning it (the `subgrid` direct-child constraint). Setting `forceLabelWidth` on the section fixes it. See *Known limitations*.

**Only `Input` is fluid today.** What happens to a non-`Input` field in a `responsive` Form depends on its own width:

| field content | in a `responsive` Form |
|---|---|
| `Input` (fluid) | shrinks to fit, floored at 10rem |
| checkbox / toggle / radios / read-only `Text` (no width) | safe — never overflows; occupies the field column |
| `Select` / `SearchInput` / `TextArea` (fixed or browser-default width) | keeps its width and **overflows** a narrow column — tracked in **CUI-36** |

## Screenshots

**`responsive` in action** — as the frame narrows, the whole section flips from label-beside-field rows to a stacked single column. No breakpoint prop; the flip is content-derived (`ResponsiveForm` story).

<table>
<tr><th>Wide — label beside field</th><th>Narrow — flipped to stacked</th></tr>
<tr>
<td valign="top">Responsive form at a wide width: each label sits to the left of its field</td>
<td valign="top">Responsive form at a narrow width: each field has dropped below its label</td>
</tr>
</table>

**Nesting &amp; non-field content** (`NestingAndNonFieldContent` story) — with `forceLabelWidth` (top section) a `FormGroup` nested inside a `div` stays aligned with the direct-child rows; without it (bottom section) the nested group stacks under its label. A `Banner` dropped straight into a section spans the full width either way.

Two form sections: the fixed-label section keeps a div-nested field aligned in the label column, the auto-label section stacks the nested field; a Banner spans the full section width in both
## 🔧 Usage

Opt in once on the `Form` (every contained `Input` becomes fluid and the section auto-flips):

```tsx
<Form responsive>                                            // ←
  <FormSection>
    <FormGroup id="name" label="Name" content={<Input id="name" size="1/2" />} />
  </FormSection>
</Form>
```

…or make a single `Input` fluid on its own, no `Form` needed:

```tsx
<Input id="q" fluid aria-label="Search" />                   // ←
```

Nesting `FormGroup`s inside a wrapper (`div` / `Stack` / `Accordion`)? Set `forceLabelWidth` on the section so rows stay aligned:

```tsx
<FormSection forceLabelWidth={180}>                          // ←
  <Stack direction="vertical">
    <FormGroup id="a" label="Name" content={<Input id="a" />} />
    <FormGroup id="b" label="Description" content={<Input id="b" />} />
  </Stack>
</FormSection>
```

> Flex ancestors between the Form and the field must allow shrinking (`min-width: 0`) for `responsive` to take effect — documented on the prop.

## Review focus

- 🟡 `form/Form.component.tsx › SectionGrid / FieldSubgridRow / FormGroup` — the layout engine now branches on `forceLabelWidth` (self-contained rows vs. `subgrid`) **and** on `responsive`. Confirm existing non-responsive forms are visually unchanged (the behavior notes above).
- 🟡 `form/Form.component.tsx › FormSection labelTrack` — `forceLabelWidth` became a hard cap; scan for consumers that relied on the label column growing past it.
- ⚪ `form/Form.component.tsx › SectionGrid auto branch` — `minmax(max-content, 1fr)` field track + `& > * { grid-column: 1 / -1 }` so the grid fills its container and non-`FormGroup` children span full width.
- ⚪ `inputv2/inputv2.tsx › InputBorder $fluid` — the `max-width: 100%; min-width: 0` opt-in, driven by `fluid` or the Form's `responsive` via `useFieldContext`.

## How to test

- **Storybook** → **Templates/Form** → `ResponsiveForm` (drag the frame's right edge to watch the row/stacked flip) and `NestingAndNonFieldContent` (a fixed-label section keeps a nested group aligned; an auto-label section stacks it; a `Banner` spans full width in both). Sanity-check `PageForm` / `FormWithAccordion` look unchanged.
- **Jest** — `npx jest src/lib/components/form src/lib/components/inputv2` (fields stay visible/typeable and the Form still submits when `responsive`/`fluid` are set).

## Known limitations

- **`responsive` makes only `Input` fluid.** `Select`, `SearchInput`, `TextArea` (and a horizontal `RadioGroup`) keep their width and overflow a shrinking column — tracked in **CUI-36**. `TextArea` with no `width` prop falls back to the browser default (~20rem), so it overflows too.
- **Auto-label sections with nested groups.** A `FormGroup` nested inside a wrapper under a `FormSection` that has **no** `forceLabelWidth` will stack instead of aligning (the `subgrid` direct-child constraint). Fix: add `forceLabelWidth` to the section. This surfaces in two consumer sections when they next bump `@scality/core-ui`, both a one-line fix: data-browser `BucketNotificationFormPage.tsx` ("Filters" section) and metalk8s `CreateVolume.tsx`. Not a regression until those repos bump.

## References

- **[CUI-32](https://scality.atlassian.net/browse/CUI-32)** — responsive/fluid Form (this PR), incl. the "flex ancestors need `min-width: 0`" rule.
- **[CUI-36](https://scality.atlassian.net/browse/CUI-36)** — follow-up: extend the fluid treatment to `Select` / `SearchInput` / `TextArea` / horizontal `RadioGroup`.

[CUI-32]: https://scality.atlassian.net/browse/CUI-32?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/components/form/Form.component.tsx
+++ b/src/lib/components/form/Form.component.tsx
@@ -104,6 +104,46 @@ const ScrollArea = styled(BasicPageLayout)`
   overflow-y: auto;
 `;
 
+const FieldRow = styled.div<{
+  $direction: 'vertical' | 'horizontal';
+  $responsive?: boolean;
+  $flipAt?: number;
+}>`
+  display: flex;
+  flex-direction: ${({ $direction }) =>
+    $direction === 'horizontal' ? 'row' : 'column'};
+  align-items: baseline;
+  gap: ${({ $direction }) =>
+    $direction === 'horizontal' ? spacing.r32 : spacing.r4};
+  ${({ $responsive }) =>
+    $responsive &&
+    css`
+      min-width: 0;
+    `}
+  ${({ $flipAt }) =>
+    $flipAt &&
+    css`
+      @container responsive (max-width: ${$flipAt}px) {
+        flex-direction: column;
+        align-items: stretch;
+        gap: ${spacing.r4};
+      }
+    `}
+`;
+
+const FieldLabelBox = styled.div<{ $width: string; $flipAt?: number }>`
+  width: ${({ $width }) => $width};
+  flex: none;
+  ${({ $flipAt }) =>
+    $flipAt &&
+    css`
+      @container responsive (max-width: ${$flipAt}px) {
+        width: auto;
+        flex: initial;
+      }
+    `}
+`;
+
 const LabelContext = createContext<{
   maxLabelWidth: number;
   setMaxLabelWidth: (setter: (value: number) => number) => void;
@@ -177,17 +217,10 @@ const FormGroup = ({
 
   return (
     <FieldContext.Provider value={value}>
-      <Box
-        display="flex"
-        flexDirection={direction === 'horizontal' ? 'row' : 'column'}
-        alignItems="baseline"
-        gap={direction === 'horizontal' ? spacing['r32'] : spacing['r4']}
-      >
-        <div
-          style={{
-            width: maxLabelWidth === 0 ? 'max-content' : `${maxLabelWidth}px`,
-            flex: 'none',
-          }}
+      <FieldRow $direction={direction} $responsive={responsive} $flipAt={flipAt}>
+        <FieldLabelBox
+          $width={maxLabelWidth === 0 ? 'max-content' : `${maxLabelWidth}px`}
+          $flipAt={flipAt}
         >
           <label
             htmlFor={id}
@@ -213,10 +246,11 @@ const FormGroup = ({
               </Box>
             )}
           </label>
-        </div>
+        </FieldLabelBox>
         <Stack
           direction={helpErrorPosition === 'right' ? 'horizontal' : 'vertical'}
           gap={helpErrorPosition === 'right' ? 'r8' : 'r4'}
+          style={responsive ? { minWidth: 0 } : undefined}
         >
           {content}
           {error ? (
@@ -235,7 +269,7 @@ const FormGroup = ({
             </HelperText>
           )}
         </Stack>
-      </Box>
+      </FieldRow>
     </FieldContext.Provider>
   );
 };
--- a/src/lib/components/form/Form.component.tsx
+++ b/src/lib/components/form/Form.component.tsx
@@ -34,19 +34,16 @@ type FormProps = Omit<
   rightActions?: ReactNode;
   banner?: ReactNode;
   /**
-   * Makes contained fields fluid: their `size`-derived width becomes a
-   * preferred width that shrinks to fit the column (capped at `max-width: 100%`)
-   * instead of overflowing. Note: only `Input` currently honors this — `Select`,
-   * `SearchInput`, `TextArea` and other size-driven field content keep their
-   * fixed width for now (tracked in CUI-36). Flex ancestors between the Form and
-   * the field must allow shrinking (`min-width: 0`) for this to take effect.
+   * Makes contained fields fluid: their `size`-derived width becomes a preferred
+   * width that shrinks to fit the column (capped at `max-width: 100%`) instead of
+   * overflowing. It also lays each `FormSection` out as a two-column grid that
+   * auto-flips to a stacked single column on narrow widths — there is no
+   * breakpoint prop; the flip point is derived from the layout (see below).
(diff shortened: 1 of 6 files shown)


damiengillesscality:
Why did you choose to use custom logic to "flipAt" and not `flex-wrap` ?
Flex wrap I think would work fairly consistently we always use the same width for our label in our UIs (I think around 200px in general) and for the fields we can either let them wrap when they would overflow or give them a min-width so that at some point they all wrap at the same time.

damiengillesscality:
Following our call there are large label cases to test and the following alternative to explore

.label {
  width: max(0.5 container-width, 280px) # not the exact syntax but the idea to scale label and allow wrap of stuff simultaneously
}
flex-wrap: wrap;
align-content: flex-start; # prevent gap when wrapping

damiengillesscality (approved):

PR #1170, opened 2026-07-24
fix(accordion): grid-row open animation — no clipping on growth, popovers escape the box (ARTESCA-17819)

JeanMarcMilletScality:
**TL;DR** — Rewrites the `Accordion` open/collapse animation to a pure-CSS grid-row transition (`0fr → 1fr`) instead of a JS-measured pixel `height`. Content that grows or shrinks after opening is never clipped, and once open the box releases `overflow` so a `Select` menu, dropdown or tooltip can escape the border.

## Context

The Accordion animated open by transitioning an **explicit measured** pixel `height`, measured once at render time (`element.scrollHeight` in a ref callback). Two bugs fell out of that:

| Bug | Cause |
|---|---|
| Content that grows/shrinks after open is clipped | the measured height goes stale on any reflow that fires no React render — a container resize, a `@container` responsive flip |
| A `Select` menu / tooltip opened near the bottom is cut off at the border | `overflow: hidden` (needed to clip the height animation) also clips anything a child legitimately renders outside the box — [ARTESCA-17819] |

This PR absorbs and closes **#1171**, which fixed only the second bug.

## Approach

**Animate a grid row, not a pixel height.** A single-row grid transitioning `grid-template-rows` from `0fr` to `1fr`; a `1fr` track always resolves to the content's *current* height, so growth/shrink is followed instantly with no measurement.

Before:
```tsx
<AccordionContent
  ref={(el) => {                                                    // ←
    el?.style.setProperty('height', isOpen ? el.scrollHeight + 'px' : '0px');
  }}
  $isOpen={isOpen}
>
  <Wrapper>{children}</Wrapper>
</AccordionContent>
// AccordionContent styled: overflow: hidden; transition: height 0.3s;   // ←
```

After:
```tsx
<AccordionContent $isOpen={isOpen} onTransitionEnd={handleTransitionEnd}>
  <ContentClip $isExpanded={isOpen && isExpanded}>                  // ←
    <Wrapper>{children}</Wrapper>
  </ContentClip>
</AccordionContent>
// AccordionContent styled: display: grid; grid-template-rows: 0fr ↔ 1fr;  // ←
// ContentClip styled:      overflow: hidden → visible; min-height: 0;
```

No `ResizeObserver`, no `scrollHeight`, no refs — the whole measurement path is deleted. (Grid-row transitions need Chrome 107 / Firefox 66 / Safari 16, all 2022+.)

**Release overflow once settled (ARTESCA-17819).** The inner `ContentClip` keeps `overflow: hidden` while the row animates, then flips to `visible` once the open transition finishes — gated by an `isExpanded` state set on the element's **own** `grid-template-rows` `transitionEnd` (`e.target === e.currentTarget` so a descendant's bubbled transition can't flip it early), initialized to `open`, and reset on close so the collapse animation clips again.

## How to test

- **Storybook → Accordion → `WithSelectAtBottom`**: open the `Select` near the bottom; its menu overflows past the accordion border and stays fully visible/scrollable.
- Open an accordion whose content changes height *after* opening (e.g. a `responsive` Form that flips to a taller stacked layout as the frame narrows) — the box follows the content; nothing is clipped.
- **Jest** — `npx jest src/lib/components/accordion`

## References

- [ARTESCA-17819] — a `Select` menu opened inside an Accordion is clipped by the box border.
- Supersedes **#1171** (overflow-when-open); its fix is folded in here.

[ARTESCA-17819]: https://scality.atlassian.net/browse/ARTESCA-17819

### DIFF
--- a/src/lib/components/accordion/Accordion.component.tsx
+++ b/src/lib/components/accordion/Accordion.component.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 
 import { spacing, Stack } from '../../spacing';
 import { Box } from '../box/Box';
@@ -65,10 +65,14 @@ const AccordionContent = styled.div<{
     visibility 0.3s;
 `;
 
-// Clips the content while the row collapses; min-height: 0 lets the grid track
-// shrink below the content's min-content height so it can reach 0.
-const ContentClip = styled.div`
-  overflow: hidden;
+// Clips the content while the row collapses (min-height: 0 lets the grid track
+// shrink to 0). Once open and fully expanded, overflow is released so a child
+// that renders outside the box — a Select menu, dropdown or tooltip — is not
+// clipped by the accordion border (ARTESCA-17819).
+const ContentClip = styled.div<{
+  $isExpanded: boolean;
+}>`
+  overflow: ${(props) => (props.$isExpanded ? 'visible' : 'hidden')};
   min-height: 0;
 `;
 
@@ -86,6 +90,9 @@ export const Accordion = ({
 }: AccordionProps) => {
   const [isOpen, setIsOpen] = useState(open);
   const [previousOpenProp, setPreviousOpenProp] = useState(open);
+  // Gates overflow: true only once the open animation has finished, so the
+  // box still clips while it animates but lets content escape once settled.
+  const [isExpanded, setIsExpanded] = useState(open);
 
   // Sync to the `open` prop when it changes, while still letting the header
   // toggle drive state in between (adjusting state during render, per React docs).
@@ -94,8 +101,28 @@ export const Accordion = ({
     setIsOpen(open);
   }
 
+  useEffect(() => {
+    // Re-clip as soon as the accordion starts closing so nothing spills during
+    // the collapse animation, and so the next open starts clipped again.
+    if (!isOpen) {
+      setIsExpanded(false);
+    }
+  }, [isOpen]);
+
   const handleToggleContent = () => {
     setIsOpen((prev) => !prev);
+  };
+
+  const handleTransitionEnd = (e: React.TransitionEvent<HTMLDivElement>) => {
+    // Only react to this element's own row transition finishing, not a
+    // transition bubbling up from a descendant.
+    if (
+      e.target === e.currentTarget &&
+      e.propertyName === 'grid-template-rows' &&
+      isOpen
+    ) {
+      setIsExpanded(true);
+    }
   };
 
   return (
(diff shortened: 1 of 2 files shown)


src/lib/components/accordion/Accordion.component.tsx
```diff
@@ -72,19 +89,42 @@ export const Accordion = ({
...
+    // Only react to this element's own row transition finishing, not a
+    // transition bubbling up from a descendant.
+    if (
+      e.target === e.currentTarget &&
+      e.propertyName === 'grid-template-rows' &&
+      isOpen
+    ) {
+      setIsExpanded(true);
+    }
+  };
```
damiengillesscality: Here your Claude assumes react 19 memoization compiler. For better performance memoize theses.

I personaly created a skill for my Claude to not forget that but you may enforce it however you want:

```
---
name: react-memoization
description: >-
  Default-memoize convention for React/TSX code. When writing or editing React components —
  adding event handlers, passing props, building dependency arrays, or deriving values —
  wrap callbacks in useCallback and non-primitive/derived values in useMemo by default
  rather than leaving them inline. Use whenever authoring or modifying React component code.
---

# React: memoize by default

Assume the project does **NOT** use the React Compiler unless you've verified it does
(check `package.json` / babel / build config for `babel-plugin-react-compiler`). When it's
absent, memoization is manual and expected — do not skip it on the assumption a
compiler/runtime handles it or that the cost is negligible. If the React Compiler *is*
enabled, this skill does not apply: prefer plain inline code and let the compiler memoize.

## Always memoize
- Every callback / event handler passed as a prop, used in JSX, or listed in another hook's
  dependency array → wrap in `useCallback`.
- Every non-primitive value (object literal, array literal) passed as a prop or used in a
  dependency array → wrap in `useMemo`.
- Derived values from non-trivial computation (sort / map / filter / reduce, parsing,
  formatting) → wrap in `useMemo`.

## Don't memoize
- Primitive literals (string / number / boolean) passed as props — pass them raw.
- Values read only once that never cross a prop or dependency-array boundary.
- Already-stable references (`navigate`, `useState` setters, react-query `mutate`) don't
  need wrapping, but still belong in the dependency array.

## Keep deps honest
Dependency arrays must be exhaustive and correct. Match the project's linter
(e.g. Biome `useExhaustiveDependencies`, or ESLint `react-hooks/exhaustive-deps`).

## Example
    // ✗ new function + new object every render
    <Chip onRemove={() => remove(item.id)} style={{ marginLeft: 4 }} />

    // ✓
    const handleRemove = useCallback(() => remove(item.id), [remove, item.id]);
    const chipStyle = useMemo(() => ({ marginLeft: 4 }), []);
    <Chip onRemove={handleRemove} style={chipStyle} />
```

damiengillesscality (approved):

PR #1173, opened 2026-07-24
build(deps): move styled-components to peerDependencies (CUI-37)

JeanMarcMilletScality:
## What

Move `styled-components` from `dependencies` to `peerDependencies` (matching `react`/`react-dom`), and add it to `devDependencies` so core-ui's own build/test/Storybook still resolve it.

## Why

As a regular `dependency`, npm was free to install a **nested** `node_modules/@scality/core-ui/node_modules/styled-components` in workspace consumers. That second instance yields two `ThemeProvider` contexts, so `useTheme()` returns `undefined` at test/build time — this broke identity-ui's Jest run (140 failures from `Checkbox`/theme consumers).

Making it a peer guarantees a single hoisted instance provided by the consumer, exactly as `react` already works.

### No Module Federation impact
The MF singleton version is negotiated from each consumer's rspack `shared` config (`requiredVersion: deps['styled-components']`, read from the consumer's own package.json), **independent** of core-ui's dependency classification. Runtime behavior is unchanged; this only affects install/build/test-time npm resolution.

### Why the devDependencies entry is required
Peer deps of the root package are not auto-installed for the package's own dev build. `react` survives peer-only here only because Storybook/testing-library pull it in transitively — nothing pulls `styled-components`, so `tsc` fails without the explicit devDep. Verified: peer-only → build fails across ~40 files; with devDep → clean.

## Verification

- **core-ui**: `npm run build` clean, `npm test` → 43 suites / 428 tests pass.
- **Pilots repointed at this branch** (fresh `--legacy-peer-deps` install, cache-clean) — all confirm a single deduped `styled-components@6.4.4`, no nested copy:
  - shell-ui: 92/92 pass
  - metalk8s ui: 220 pass / 2 skip / 3 todo
  - identity-ui: 248 pass (2 pre-existing `GroupsTab` form-logic failures, unrelated — this commit changes only package metadata, not compiled code)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39,7 +39,6 @@
         "react-window": "^1.8.6",
         "recharts": "^3.0.2",
         "remark-gfm": "^4.0.1",
-        "styled-components": "^6.4.3",
         "styled-system": "^5.1.5",
         "uuid": "^14.0.0"
       },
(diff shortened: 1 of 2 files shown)


damiengillesscality (changes requested):
I'm not sure about this. It's at least semantically false, styled-components is a core dep not a dev dep as it is used by the final build. I don't know about this jest issue in identity-ui maybe we can make a meeting to discuss it.

PR #1157, opened 2026-07-16
build(deps): bump dependabot dependencies

JeanMarcMilletScality:
## TL;DR

Bumps six dependencies (react-router, react-router-dom, @babel/core, esbuild, markdown-it, linkify-it, ws) in one PR — the changes dependabot proposed across separate PRs, consolidated so they can actually land.

## Context

Dependabot opened these as individual PRs, but none can merge: the `claude-review` check fails on every dependabot PR because repository secrets aren't exposed to the dependabot Actions context. This PR ships the same bumps from a normal branch instead. (The CI gate itself is tracked for a separate fix.)

## Changes

| Dependency | From → To | Kind | Why it matters |
|---|---|---|---|
| `react-router` | 7.13.0 → 7.15.1 | direct (ships) | fetcher/SSR/onError bug fixes |
| `react-router-dom` | 7.13.0 → 7.15.1 | direct (ships) | kept in lockstep with react-router — see Approach |
| `@babel/core` | 7.28.6 → 7.29.7 | dev | source-map bug fixes |
| `esbuild` | 0.25.12 → 0.28.1 | transitive | security: dev-server path traversal, Deno integrity |
| `markdown-it` | 14.1.1 → 14.3.0 | transitive | security/perf + CommonMark fixes |
| `linkify-it` | 5.0.0 → 5.0.2 | transitive | security: fuzzy-link ReDoS |
| `ws` | 8.20.1 → 8.21.1 | transitive | security: remote memory-exhaustion DoS |

`esbuild` and `ws` match dependabot's proposed versions exactly; the other transitive/dev deps resolved one patch/minor newer via `npm update` (all within existing ranges, no breaking changes — markdown-it 14.3.0 is a single CommonMark hard-line-break fix).

## Approach

**Why react-router-dom moves with react-router.** Dependabot's PR bumped only `react-router`. But `react-router-dom@7.13.0` hard-pins `react-router@7.13.0`, so bumping react-router alone leaves a second, nested copy of react-router in the tree — two separate Router contexts, which breaks `useNavigate`/`useLocation` for consumers whose provider comes from the other copy.

Before — react-router bumped alone (dependabot #1131):
```
node_modules/react-router                                 7.15.1
node_modules/react-router-dom                             7.13.0
node_modules/react-router-dom/node_modules/react-router   7.13.0   // ← duplicate: 2nd Router context
```

After — pair bumped together (this PR):
```
node_modules/react-router       7.15.1
[...]

PR #1176, opened 2026-07-28
build(deps): bump dependabot dependencies

JeanMarcMilletScality:
## TL;DR

Consolidates the pending dependabot dependency bumps into one PR, plus the advisories that were published while those PRs sat open. Takes `npm audit` from **9 root advisories on `development/1.0` down to 0** — every package this branch touches is now free of high and critical advisories.

Also bumps `dependency-review-action` v3 → v5 (the v3 Node 20 runtime is deprecated) and enables `show-patched-versions`.

## Context

Dependabot opened these as individual PRs, but none can merge: the `claude-review` check fails on every dependabot PR because repository secrets aren't exposed to the dependabot Actions context. This PR ships the same bumps from a normal branch instead. (The CI gate itself is tracked for a separate fix.)

## Changes

| Dependency | From → To | Kind | Why |
|---|---|---|---|
| `react-router` | 7.13.0 → **7.18.2** | direct (ships) | 5 high advisories — see Approach |
| `react-router-dom` | 7.13.0 → **7.18.2** | direct (ships) | kept in lockstep with `react-router` — see Approach |
| `@babel/core` | 7.28.6 → 7.29.7 | dev | source-map bug fixes |
| `esbuild` | 0.25.12 → 0.28.1 | transitive | security: dev-server path traversal, Deno integrity |
| `markdown-it` | 14.1.1 → 14.3.0 | transitive | security/perf + CommonMark fixes |
| `linkify-it` | 5.0.0 → 5.0.2 | transitive | security: fuzzy-link ReDoS |
| `ws` | 8.20.1 → 8.21.1 | transitive | security: remote memory-exhaustion DoS |
| `postcss` | 8.5.15 → 8.5.24 | transitive | security: path traversal in source-map auto-loading |
| `fast-uri` | 3.1.2 → **3.1.5** | transitive | security: host confusion via backslash authority delimiter (GHSA-7p8r-x3mc-p8w7) |
| `js-yaml` | 4.1.1 → **4.3.1**, 3.14.2 → **3.15.1** | transitive | security: quadratic DoS via merge-key chains, then via `!!omap` resolution (GHSA-5p4m-2wfm-xmqj) |
| `nanoid` | 3.3.15 → **3.3.18** | transitive | security: custom generators loop indefinitely when size is zero (GHSA-2v37-7h3g-55p8) |
| `brace-expansion` | nested copies → **1.1.18 / 2.1.4 / 5.0.9** | transitive | security: exponential-expansion DoS, plus a bypass of that mitigation (GHSA-rgw5-rvv9-x895) |
| `dependency-review-action` | v3 → **v5** | CI | Node 20 runtime deprecated; adds `show-patched-versions` |

## Approach

**Why `react-router-dom` moves with `react-router`.** Dependabot's PR bumped only `react-router`. But `react-router-dom@7.13.0` hard-pins `react-router@7.13.0`, so bumping react-router alone leaves a second, nested copy of react-router in the tree — two separate Router contexts, which breaks `useNavigate`/`useLocation` for consumers whose provider comes from the other copy.

Before — react-router bumped alone (dependabot #1131):
```
node_modules/react-router                                 7.18.2
node_modules/react-router-dom                             7.13.0
node_modules/react-router-dom/node_modules/react-router   7.13.0   // ← duplicate: 2nd Router context
```

After — pair bumped together (this PR):
```
$ npm ls react-router
├─┬ react-router-dom@7.18.2
│ └── react-router@7.18.2 deduped
└── react-router@7.18.2
```

**Why 7.18.2 and not 8.3.0.** Dependabot #1174 proposes `react-router@8.3.0`. That isn't viable here:

- `react-router@8.3.0` declares peers `react >=19.2.7`, but core-ui's `peerDependencies` still allow `react ^18.0.0`. Taking it would silently drop React 18 support for every consumer.
- `react-router-dom` **has no 8.x release at all** — it stops at 7.18.2. So bumping `react-router` to 8 recreates exactly the duplicate-Router-context problem described above.

`7.18.2` is a minor bump that clears **all five** open react-router advisories, and keeps peers at `react >=18`. Four were fixed in 7.18.0 (open redirect via backslash in `<Link>`/`useNavigate`, RSCErrorHandler missing protocol validation, arbitrary constructor injection in `deserializeErrors()`, unauthenticated DoS via inefficient route matching). The fifth — **GHSA-qwww-vcr4-c8h2**, RSC mode CSRF bypass — was initially published as fixed only in 8.3.0, but GitHub has since backported it to **7.18.2**, so no React 19 bump is needed.

core-ui imports only `MemoryRouter`, `BrowserRouter`, `useLocation`, `useNavigate` — all unchanged across 7.x.

**Why `brace-expansion` is not just dev tooling.** An earlier revision of this description claimed the flagged copy was reachable only from dev tooling. That is wrong, and worth stating plainly because `dependency-review-action` already defaults to `fail-on-scopes: runtime` — a genuinely dev-only package would never have tripped the gate. The lockfile shows a **runtime**-scoped path (`dev: false`):

```
react-query@3 → broadcast-channel → rimraf@3 → glob@7 → brace-expansion@1.x
```

So these bumps are a real fix, not noise. All three lines are now at their patched releases (1.1.18 / 2.1.4 / 5.0.9).

**Why the action bump rides along.** For `pull_request` events GitHub runs the workflow definition from the PR head, so bumping the action here takes effect on this PR's own check — which is how the v5 result below was produced. `show-patched-versions` is the input that matters: the v3 failure summary named only the vulnerable package, which is what made GHSA-qwww-vcr4-c8h2 and GHSA-mh99-v99m-4gvg look unfixable when in-range patches already existed. Both v4.0.0 and v5.0.0 are runtime-only majors with no input changes, so `fail-on-severity: high` carries over untouched. `permissions: contents: read` is now declared explicitly (the value GitHub's own README specifies) instead of inheriting the repository default.

## Review focus

- 🟡 `package.json` › `react-router` + `react-router-dom` — both pinned to exactly `7.18.2`; confirm they stay aligned (this is the shipped runtime dep).
- 🟡 `.github/workflows/dependency-review.yaml` — v5 requires Actions runner ≥ v2.327.1, which GitHub-hosted `ubuntu-latest` exceeds. If this job ever moves to a self-hosted runner, that floor has to be met.
- ⚪ `package-lock.json` — changes are scoped to the deps above plus their `@babel/*` / `@esbuild/*` sub-packages; no unrelated churn.

## How to test

```bash
npm ci          # 0 vulnerabilities
npm test        # 428 tests / 43 suites pass
npm run build   # tsc, exit 0
npm ls react-router   # exactly one copy, 7.18.2
```

## Remaining advisories

**None.** All 56 packages this branch changes relative to `development/1.0` were swept against the GitHub advisory database; none carries a high or critical advisory, and `npm audit` reports 0. The `dependency-review` check passes on v5 with `fail-on-severity: high`.

Note that `development/1.0` itself still reports open advisories from packages this PR does not touch — that is a separate cleanup, not a regression here.

## References

Supersedes these dependabot PRs (can be closed once this merges):

- #1131 — bump react-router 7.13.0 → 7.15.1
- #1134 — bump @babel/core 7.28.6 → 7.29.6 (dev)
- #1136 — bump markdown-it 14.1.1 → 14.2.0
- #1145 — bump esbuild 0.25.12 → 0.28.1
- #1149 — bump linkify-it 5.0.0 → 5.0.1
- #1154 — bump ws 8.20.1 → 8.21.1
- #1166 — bump linkify-it 5.0.0 → 5.0.2
- #1167 — bump fast-uri 3.1.2 → 3.1.4
- #1175 — bump postcss 8.5.15 → 8.5.23

Should be closed as **not viable** rather than superseded:

- #1174 — bump react-router 7.13.0 → 8.3.0 (React-19-only peer; no matching `react-router-dom` 8.x)

Replaces #1157, which was opened from a branch with an invalid `build/` prefix.

---


damiengillesscality (approved):
Please handle the bump in the other applications and share with the team the info

PR #1190, opened 2026-08-27
Feat/new UI components

clement-scality:
**Component**: `charts` — Barchart, LineTimeSerieChart, and two new Storybook recipes

**TL;DR** — Adds an opt-in logarithmic Y axis to `Barchart` and `LineTimeSerieChart`, plus two story-only recipes proposing a `Heatmap` and a `NestedProportionBar` built from primitives that already exist.

### Context / Why

A metric that idles at 2 ms and spikes to 4 s cannot be read on a linear axis: the axis is sized by the spike, so the baseline where the metric spends nearly all its time is pressed flat against zero.

The two recipes are a **design proposal, not a component**. They live entirely in `stories/` so the shape can be reviewed, and argued with, before anything is added to `src/`.

### 🧩 Approach — the logarithmic axis

`yAxisScale="log"`, defaulting to `linear`. It is a display option and nothing more: no value is rescaled, and tooltips, legend and unit scaling all keep the numbers the caller passed in. Only the axis and where the marks land on it change.

Three things a log axis has to do differently, all in `charts/common` so both charts share them:

| | Why |
|---|---|
| The scale cannot start at zero | Bounded by the decades enclosing the data, read off the smallest positive value and the largest — so it follows the data rather than being anchored to 1. Values between 0.001 and 0.1 with spikes to 50 give a 0.001..100 axis. |
| Its ticks are the decades | Evenly spaced linear ticks on a log axis crowd into the top of the plot and read as a broken chart. Past six decades, whole decades are skipped — never subdivided. |
| A measured zero needs somewhere to go | It has no logarithm and no place on the scale, so the axis reserves one slot below its first decade, labels it `0`, and draws zeros there. |

That last one is the part worth arguing about. Without the reserved slot a zero has to be dropped, and a dropped zero leaves exactly the gap a missing sample leaves — making "the metric read zero" indistinguishable from "we have no data", which are not the same fact. Clamping to the first decade is worse, not better: it draws the zero as the smallest measured value, a number the reader will believe. The slot costs a decade's worth of height, so it is only reserved when the data actually holds a zero.

Two pairings are refused rather than drawn wrong. **Stacked bars fall back to linear**, since stacking places each segment at a cumulative sum and a segment's height would stop matching its value. **A symmetrical line chart refuses the scale at the type level** — half its axis is negative by construction.

### 📷 Screenshots

**Barchart — `LogarithmicScale`.** Linear on top: only the 40 000 bar is legible. Log below: every decade gets the same height, and `category5` (value `0`) is the stub at the reserved `0`.

Barchart logarithmic scale story, linear and log side by side

**LineTimeSerieChart — `LogarithmicScaleExample`.** Same series twice. On the linear axis the baseline is a flat line on zero; on the log axis its variation and both spike tiers are readable.

LineTimeSerieChart logarithmic scale story, linear and log side by side

**Heatmap — default story.** One recipe of the proposal; the rest is worth clicking through rather than screenshotting.

Heatmap composition recipe, services against time slots with a status legend

**NestedProportionBar — default story.**

image
### 🔍 Review focus

- 🟡 **Moderate** — `charts/common/chartUtils.ts › getLogAxis`, `placeNonPositiveValues` — the domain, the decade ticks and the reserved zero slot. The zero handling is the design decision in this PR; if you disagree with reserving a slot, this is the place to say so.
- 🟡 **Moderate** — `barchart/Barchart.utils.ts › getCurrentPoint` and `LineTimeSerieChartTooltip` — a zero is plotted at the reserved slot, so both tooltips map it back before formatting, including a caller's own Barchart `tooltip` renderer. No real value can collide with the slot: it sits below the smallest positive value in the data.
- ⚪ **Minor** — `stories/Heatmap`, `stories/NestedProportionBar` — story-only, nothing in `src/`. Review the shape, not the code.

### 🧪 How to test

```bash
npm run storybook
```

**The logarithmic axis** — *Charts → Barchartv2* and *Charts → LineTimeSerieChart*:

1. Open `Logarithmic scale` on either chart and compare the two stacked charts. Same data, only the axis differs.
2. Open `Logarithmic scale playground` and use the Controls panel — `yAxisScale` flips the axis live.
3. On the Barchart playground, tick **`stacked`**. The axis snaps back to linear: that is the documented refusal, not a bug.
4. On the LineTimeSerieChart playground, tick **`zeros`**. The zeros land on the reserved `0` at the bottom; untick and the axis loses that slot.
5. `Logarithmic scale below one` shows the axis following data between 0.001 and 0.1 with spikes to 50 — nothing is anchored to 1.

**The two proposed recipes** — *Charts → Heatmap (composition)* and *Charts → NestedProportionBar (composition)*. Both open on a `Playground` story whose **data is a control**, so this is meant to be poked at rather than read:

6. Heatmap → `Playground` → Controls → **`rows`**. It is a JSON editor: add a row, rename one, change any cell to `OK` / `WARNING` / `CRITICAL` / `NONE`. The x-axis is derived from the longest row, so adding cells adds columns.
7. Click a legend entry. Unselected statuses fade rather than vanish, so the shape of the data survives the filter.
8. NestedProportionBar → `Playground` → Controls → **`root`**. Change a value, add a child, set a `color` to a theme token, a `chartColors` key or a raw CSS colour.
9. Still on `root`: make the children sum **past** their parent (say 80 and 60 of 100). They scale to fit the row while every label keeps stating what you typed — an inconsistent payload should read as inconsistent, not as a broken component.
10. `Edge cases` covers the rest: an unattributed remainder, labels wider than their box, a total of zero.

Theme switching is worth a pass on all of the above — the reserved `0`, the status colours and the box outlines all come from tokens.

```bash
npm test -- src/lib/components/charts
```

### 🔗 References

**Breaking Changes**: none. `yAxisScale` is optional and defaults to `linear`, so every existing caller renders exactly as before. Nothing is added to `src/` for the two recipes.

`charts/common/chartUtils.ts` carries the whole log-axis vocabulary — `getLogAxis`, `getMinPositiveValue`, `hasZeroValue`, `placeNonPositiveValues`, `formatLogTickValue`, `readLogPlottedValue` — next to `getRoundReferenceValue`/`getTicks`, which do the same job for the linear axis. Both charts wire it the same way, which is why the two diffs read almost identically.

`stories/Heatmap/` proposes two recipes: `StatusHeatmap` for discrete statuses, built from `Box` + `Tooltip` + the existing `ChartLegend`, and `NumericHeatmap` for continuous values under an opacity ramp. An earlier third recipe stacking one `GlobalHealthBar` per row was dropped — it worked, but it needs one chart instance per row and hides every x-axis but the last.

`stories/NestedProportionBar/` also carries `resolveChartColor`, which is a candidate for `charts/common` in its own right: `GlobalHealthBar` hardcodes `theme.statusX`, `Sparkline` wants a hex and `LegendShape` does `chartColors[c] || c` — three conventions for one question, where a single theme-token → palette → raw-CSS resolution serves all three.

</details>

### DIFF
--- a/src/lib/components/charts/barchart/Barchart.test.tsx
+++ b/src/lib/components/charts/barchart/Barchart.test.tsx
@@ -153,6 +153,117 @@ describe('Barchart', () => {
     });
   });
 
+  describe('Logarithmic Y axis', () => {
+    const spanningBars = [
+      {
+        label: 'Success',
+        data: [
+          ['category1', 2],
+          ['category2', 300],
+          ['category3', 0],
+        ],
+      },
+    ] as const;
+
+    /**
+     * The Y-axis tick labels, whitespace stripped: formatISONumber groups
+     * thousands with a non-breaking space, which is not what a test wants to
+     * pin down.
+     */
+    const yAxisTicks = (container: HTMLElement) =>
+      Array.from(
+        container.querySelectorAll('.recharts-yAxis-tick-labels text'),
+      ).map((tick) => (tick.textContent ?? '').replace(/\s/g, ''));
+
+    const renderBarchart = (props: {
+      yAxisScale?: 'linear' | 'log';
+      stacked?: boolean;
+    }) => {
+      const { Wrapper } = getWrapper();
+      return render(
+        <Wrapper>
+          <ChartLegendWrapper colorSet={testColorSet}>
+            <Barchart
+              title="Test Title"
+              type={{ type: 'category' }}
+              bars={spanningBars}
+              {...props}
+            />
+          </ChartLegendWrapper>
+        </Wrapper>,
+      );
+    };
+
+    it('labels the axis with the decades enclosing the data, plus a zero', () => {
+      const { container } = renderBarchart({ yAxisScale: 'log' });
+
+      // 2..300 gives a 1..1000 scale — a log axis cannot start at zero — and the
+      // dataset holds a zero, so the axis reserves a slot below it for one.
+      expect(yAxisTicks(container)).toEqual(['0', '1', '10', '100', '1000']);
+    });
+
+    it('reserves no zero slot when the data has none', () => {
+      const { Wrapper } = getWrapper();
+      const { container } = render(
+        <Wrapper>
+          <ChartLegendWrapper colorSet={testColorSet}>
+            <Barchart
+              title="Test Title"
+              type={{ type: 'category' }}
+              bars={[
+                {
+                  label: 'Success',
+                  data: [
+                    ['category1', 2],
+                    ['category2', 300],
+                  ],
+                },
+              ]}
+              yAxisScale="log"
+            />
+          </ChartLegendWrapper>
+        </Wrapper>,
+      );
+
+      // An empty band would spend a decade's worth of plot height on nothing.
+      expect(yAxisTicks(container)).toEqual(['1', '10', '100', '1000']);
+    });
+
+    it('keeps the linear ticks when no scale is asked for', () => {
+      const { container } = renderBarchart({});
+
+      // A linear axis starts at zero and tops out at the data, not at a decade.
+      expect(yAxisTicks(container)[0]).toBe('0');
+      expect(yAxisTicks(container)).not.toContain('1000');
+    });
+
+    it('draws a measured zero at the reserved slot rather than hiding it', () => {
+      const { container } = renderBarchart({ yAxisScale: 'log' });
+
+      // category3 is 0. It gets a rectangle like the other two — minPointSize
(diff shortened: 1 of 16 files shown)


eloisca:
About the "NestedProportionBar — default story" part:

Rectangle of the same level should have the same height, so should be equal to max height. 

Ex: In the given exemple, the height of "Available" should be equal to the height of "Used".

clement-scality:
Split in two: the logarithmic Y axis is now #1218, the heatmap — reworked from the story recipe into a component — is #1219. Both are drafts off a fresh `development/1.0`.

This branch is restored to the state it was in before the split (`263466e8`), NestedProportionBar recipe included, so nothing here is lost.

PR #1192, opened 2026-08-27
Table: six row and header fixes — in-cell clicks, column alignment, cut-off labels

JeanMarcMilletScality:
**TL;DR** — Six defects in `Table` that only show up once a table is used in earnest: clicking a button inside a row also selected the row (and the re-render closed whatever the button had just opened), header labels sat over the wrong columns, a value ignored its column's alignment, a cut-off header label had no way to show its full text, and a selectable table looked inert until the first click. All six are fixed.

## Context

All six surfaced while putting a table to real use, and they overlap so heavily in the same two files that as separate PRs they would conflict with each other rather than review independently.

## Approach

Most of them come back to one mechanism — how a header cell and its body cells agree on a column's width — which is what makes them one PR rather than six.

| Defect | Cause → fix |
| --- | --- |
| A click on an in-cell control also selected the row, and the selection re-render closed whatever that control had just opened | row `onClick`/`onKeyDown` fired for any target inside the row → one `shouldIgnoreRowEvent` guard, bounded to the row at both ends |
| Header tracks drifted from body tracks, and drifted again once one column ran short | `TableRowMultiSelectable` never declared the `gap` `HeadRow` has, and body cells lacked the header's `min-width: 0` reset → declare both |
| A plain string value ignored its column's alignment — 94px off centre on a centred column | `ConstrainedTextContainer` hard-coded `text-align: left`, plus a one-sided `Box mr={4}` → inherit the alignment and drop the margin; the row's own `gap` already separates columns |
| An ellipsized header label had nothing to recover it | body cells recover via `ConstrainedText`, headers had no equivalent → `TruncatableHeaderLabel`, offering a `Tooltip` only once the label is genuinely cut off |
| A selectable table looked inert until its first click | the hover affordance was gated on `$selectedId` — "something is selected" — not on whether the table can be selected → gate it on `$selectable` |
| The header reserved the caret's room on its behalf, across three constants that had to agree by hand | the caret was a zero-width proxy with the room in `HeaderContent`'s padding → the caret carries its own width, and one constant describes it |

The in-flow caret is only safe because of the `min-width: 0` fix: real caret width lifts a sortable header's min-content above its body cell's, and that only stops mattering once `min-width: auto` is out of play — the two changes are coupled. Header-minus-body `left` and `width` are exactly 0 on every column across ten widths from 1400px down to 360px, single- and multi-selectable.

## Screenshots

**Table → Simple Content Table** at full width — `Age` and its values share one right edge, `Health` and its caret are centred over the statuses, and the trailing blank header carries one right-aligned button per row:

cui1192-table-alignment-wide

The same table at 620px — labels ellipsize, the four tracks still agree between header and body, and the action column keeps its button intact rather than shrinking it:

cui1192-table-alignment-620px

## Review focus

- 🔴 `SingleSelectableContent.tsx` / `MultiSelectableContent.tsx` › row handlers — a consumer relying on "a click anywhere in the row selects it" will notice, and one carrying its own `stopPropagation` on an interactive cell can now drop it; `TableCommon.tsx › INTERACTIVE_SELECTOR` is the whole guard, so check nothing a cell renders is missing from it.
- 🟡 `Tablestyle.tsx › SortCaretWrapper` + `HeaderContent` — changes header geometry for every sortable table, and is safe only because the `min-width: 0` fix landed in the same PR, so review the two together.
- 🟡 `constrainedtext/Constrainedtext.component.tsx` — the only change outside `tablev2`, on a publicly exported component: `text-align` goes from a hard-coded `left` to `inherit`, so an external consumer relying on it staying left-aligned under a centring ancestor will see it centre.

## How to test

1. `npm run storybook`, open **Table → Simple Content Table**. Each header label sits over its own column and stays there as you narrow the browser: **Age** flush-right over the numbers, **Health** centred over the statuses, the name columns flush-left.
2. Hover a sortable header — the sort incentive appears and **nothing moves**. Hover the blank action header: no caret, and not a tab stop.
3. **Responsive Column Drop With Reveal** — narrow until the `+N` trigger appears, then click it on an *unselected* row. The panel opens and stays open; it used to select the row, re-render and close.
4. **Table With View Action** — click the button (the action fires, the row does not select), click the row's plain text (it selects), Tab to the button and press Enter (the button activates only). The button now sits inside its own column instead of overhanging by 7px.
5. Narrow until a header label ellipsizes and hover it — the full label appears; widen until it fits and hover again — no tooltip. Then hover a row in any single-selectable story **before clicking anything**: the affordance and pointer cursor are already there.

## Follow-up

- `TooltipContainer` is an `inline-block` with no `min-width: 0`, which is why `ConstrainedText` and now `HeaderLabelFrame` both carry the same unwrapping workaround. Fixing it at source deletes both, but it touches every `Tooltip` in the library.
- `organisms/attachments/AttachmentTable.tsx` carries a `marginLeft: 'auto'` that is inert at a grow-sum above 1 but becomes live below it, and a `<Box flex={0.5} />` `Header` that is a no-op. Both were cleared as causes of the misalignment fixed here, so they are left alone rather than widening this PR.
- react-table v7 spreads `key` inside `getRowProps()`/`getCellProps()`, which React reports as an error on every row and header. Pre-existing and library-wide, but it is console noise for anyone testing the above.

<details>

The alignment demonstration moved onto the **shared** default columns in `stories/tablev2.stories.tsx` rather than a separate story, so every story built on them changes appearance and gains a button per row. That action column's `8.5rem` is measured against the button's current label rather than derived, so it needs revisiting if the label changes.

</details>

### DIFF
--- a/src/lib/components/constrainedtext/Constrainedtext.component.tsx
+++ b/src/lib/components/constrainedtext/Constrainedtext.component.tsx
@@ -22,7 +22,12 @@ type Props = {
 const ConstrainedTextContainer = styled.div`
   overflow: hidden;
   text-overflow: ellipsis;
-  text-align: ${(props) => (props.$centered ? 'center' : 'left')};
+  /* inherit, not left: this is a rule on the element, so a hard-coded value
+     silently outranks whatever alignment the container set -- a table cell
+     declaring textAlign: center kept left-aligned text with nothing to explain
+     why. The centered prop stays as an explicit override, for a container that
+     sets no alignment of its own. */
+  text-align: ${(props) => (props.$centered ? 'center' : 'inherit')};
 
   ${(props) =>
     props.$lineClamp > 1
--- a/src/lib/components/tablev2/MultiSelectableContent.test.tsx
+++ b/src/lib/components/tablev2/MultiSelectableContent.test.tsx
@@ -67,7 +67,7 @@ describe('MultiSelectableContent', () => {
     expect(lastCallRows[0].original).toEqual(data[3]); // Ninette (firstName-sorted index 0)
   });
 
-  it('does not include a previously single-clicked row in subsequent multi-selection (ARTESCA-8467)', async () => {
+  it('does not include a previously single-clicked row in subsequent multi-selection', async () => {
     const onSingleRowSelected = jest.fn();
     const onMultiSelectionChanged = jest.fn();
     renderMultiSelectTable({ onSingleRowSelected, onMultiSelectionChanged });
--- a/src/lib/components/tablev2/MultiSelectableContent.test.tsx
+++ b/src/lib/components/tablev2/MultiSelectableContent.test.tsx
@@ -121,3 +121,79 @@ describe('MultiSelectableContent', () => {
     expect(within(screen.getAllByRole('row')[3]).getByRole('checkbox')).toBeChecked();
   });
 });
+
+describe('MultiSelectableContent row click vs in-cell controls', () => {
+  const columnsWithButton: TableProps['columns'] = [
+    { Header: 'First Name', accessor: 'firstName' },
+    {
+      Header: 'Action',
+      accessor: 'lastName',
+      disableSortBy: true,
+      Cell: ({ row }) => (
+        <button onClick={() => row.original.onAction()}>
+          Detach {row.original.firstName}
+        </button>
+      ),
+    },
+  ];
+
+  const renderWithAction = (
+    onAction: jest.Mock,
+    props: {
+      onMultiSelectionChanged?: jest.Mock;
+      onSingleRowSelected?: jest.Mock;
+    } = {},
+  ) =>
+    render(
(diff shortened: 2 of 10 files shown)


JeanMarcMilletScality:
@claude The `Missing SortCaretHeaderAlignment story` finding is stale — it was read against the previous PR description. That story was **deliberately removed** at the reviewer's request; the alignment behaviour now lives on the default `columns` fixture in `SimpleContentTable` (`Age` right-aligned, `Health` centred, plus a header-less action column), and `tablev2.guideline.mdx` gained a `Column alignment` section. The description and test plan were rewritten to match at 08:12 UTC, after this check had already started at 08:10 — the live body no longer references any panel A–D. Please re-review against the current description.

damiengillesscality (approved):
Approved but should probably have been splited in a format-refacto and header-change PRs, a lot of things are moved or reindented it makes identifying actual changes somewhat hard to spot

PR #1193, opened 2026-08-28
Form: let the label column share the squeeze with the field

JeanMarcMilletScality:
## TL;DR

In a `responsive` Form, a label column with an explicit width now gives ground at the same time as the field beside it, instead of holding its full width until the field hits its floor and the row stacks.

## Context

`CUI-32`/`CUI-36` already let the label column shrink. What it does not do is shrink *in time*: a section with `forceLabelWidth` kept its full cap while the field walked all the way down to `10rem`, so a form ran out of room while a third of it was still label.

## Approach

A grid track satisfies its **max** before a flexible track flexes. `minmax(min-content, 224px)` therefore takes its 224px first, and `minmax(10rem, 1fr)` absorbs the entire squeeze. Clamping the cap to a share of the container makes the two shrink together:

Before:
```ts
const labelTrack = responsive
  ? `minmax(min-content, ${labelWidthCap})`   // ←
  : labelWidthCap;
```

After:
```ts
const labelTrack = responsive
  ? `minmax(min-content, ${
      labelWidth ? `min(${labelWidth}, ${LABEL_MAX_CQI}cqi)` : labelWidthCap   // ←
    })`
  : labelWidthCap;
```

`cqi` resolves against the `responsive` container the Form already establishes, so this needs no measurement and no new prop. Above ~2.5× the cap the fraction is the larger of the two and the column is sized exactly as before.

`40cqi` is a **starting point for review on the preview, not a derived constant** — it is "slightly under half, with the column gap and padding accounted for".

Measured on the `ResponsiveLabelSqueeze` story — container width → label track / field track (widest input):

| `forceLabelWidth` | Container | Before | After |
| --- | --- | --- | --- |
| 224 | 520 | 224 / 233 (206) | 208 / 249 (222) |
| 224 | 480 | 224 / 193 (167) | 192 / 225 (198) |
| 224 | 440 | 221 / 156 (132) | 176 / 201 (174) |
| 265 | 620 | 265 / 309 (282) | 248 / 309 (282) |
| 265 | 520 | 265 / 192 (167) | 208 / 249 (222) |
| 265 | 460 | 241 / 156 (132) | 184 / 213 (186) |
| 265 | 440 | 241 / 156 (132) | *stacked* |

Zero overflow at every width in both columns; after the change the field never reaches its floor.

**The stacking threshold was also wrong for any wide cap.** It was a module constant built from `LABEL_MIN_CH`, so a section capped at 265px needed 437px to fit but did not stack until ~367px — a 70px band where the field sat pinned at its floor with no recovery. It is now derived from the section's own label width. An auto-sized column keeps the `24ch` floor, so that case is unchanged.

## Usage

`forceLabelWidth` now takes a CSS length as well as a pixel number, so a cap can follow the root font size instead of being computed once. Both forms are in the stories:

Before:
```tsx
<FormSection title={{ name: 'Identity', icon: 'Account' }} forceLabelWidth={224}>
```

After — either is valid, the string form is new:
```tsx
<FormSection title={{ name: 'Identity', icon: 'Account' }} forceLabelWidth="15rem">   // ←
```

## Review focus

- 🟡 `form/Form.component.tsx` › `FormSection` / `labelTrack` — the `min(cap, 40cqi)` clamp. It changes resting appearance on every `responsive` form with an explicit label cap, between roughly 1024 and 1200px. This is the number to argue about.
- 🟡 `form/Form.component.tsx` › `stackBelow` — the threshold moved from a module constant to a per-section value threaded through `SectionGrid` and `FieldSubgridRow` as `$stackBelow`. Same plumbing as the existing `$labelTrack`, no new context, but it is the change that alters *when* a section stacks.
- ⚪ `form/Form.component.tsx` › `forceLabelWidth` — widened to `number | string`. Additive; a bare number still means px.

## How to test

1. `npm run storybook`
2. Templates → Form → **ResponsiveLabelSqueeze**. Drag the dashed frame's right edge inward. The label column should start shrinking well before the input reaches its minimum, and the row should stack only once both are genuinely out of room.
3. Set the `forceLabelWidth` control to `265` and repeat — it should stack before anything overflows, which it previously did not.
4. **ResponsiveLabelSqueezeRemCap** — the same shape with a `15rem` cap; the column should track the root font size.
5. Every other Form story should be unchanged at full width.

## Follow-up

The `40cqi` share needs a designer's call on a live preview before this leaves draft.

## References

- `CUI-32` / `CUI-36` — made the label column shrink at all; this PR is about when.

### DIFF
--- a/src/lib/components/form/Form.component.tsx
+++ b/src/lib/components/form/Form.component.tsx
@@ -430,12 +430,16 @@ type FormSectionProps = {
   title?: { name: string; icon?: IconName; helpTooltip?: string };
   /**
    * Caps the label column at this width: labels wider than it wrap rather than
-   * widening the column. A number is pixels; a string is any CSS length, so
-   * `'15rem'` follows the document's root font size where a px value computed once
-   * cannot. In a `responsive` Form the column keeps this width while there is
-   * room but shrinks below it (down to the longest word) as the field track is
-   * squeezed, so rows stay aligned; otherwise it is pinned to this exact width.
-   * When unset, the column auto-sizes to the widest label.
+   * widening the column. A number is pixels; a string is an **absolute** length
+   * (`px`, `rem`, `em`, `ch`), so `'15rem'` follows the document's root font size
+   * where a px value computed once cannot. A percentage or a keyword such as
+   * `max-content` is not usable: the width is also fed to the container query
+   * that decides when the section stacks, and neither is valid there, so the rule
+   * is dropped whole and the section silently stops stacking. In a `responsive`
+   * Form the column keeps this width while there is room but shrinks below it
+   * (down to the longest word) as the field track is squeezed, so rows stay
+   * aligned; otherwise it is pinned to this exact width. When unset, the column
+   * auto-sizes to the widest label.
    */
   forceLabelWidth?: number | string;
   rightActions?: ReactNode;
--- a/src/lib/components/form/Form.component.tsx
+++ b/src/lib/components/form/Form.component.tsx
@@ -57,16 +57,29 @@ type TabFormProps = { layout: { kind: 'tab' } } & FormProps;
 
 // In a `responsive` Form a FormSection lays its FormGroups out as a two-column
 // CSS grid (label track + field track) and each FormGroup is a `subgrid` row, so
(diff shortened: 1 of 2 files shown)


MonPote (approved):
lGTM

PR #1194, opened 2026-08-28
Form: reserve the scroll gutter and dim the FormGroup label

JeanMarcMilletScality:
## TL;DR

Two resting-appearance changes to `Form`: the scroll area reserves a gutter for its scrollbar, and a `FormGroup`'s label renders dimmer than the value beside it.

## Context

Both come from a design review of forms at narrow widths: the scrollbar sat directly against the fields, and a key/value row read as two columns of equally weighted text with the pairing left to be inferred from position.

## Approach

Bundled into one PR because they are one designer look at the same component, not because they share a mechanism.

**Gutter.** A `page` form insets its content 16px on the right; a `tab` form's inset comes from an ancestor *outside* the scroller, so the bar lands against the fields. Independently, either layout shifted every field left by the scrollbar's width at the moment the form grew past its box.

```
const ScrollArea = styled(BasicPageLayout)`
  flex-grow: 1;
  align-self: stretch;
  overflow-y: auto;
  scrollbar-gutter: stable;   // ←
`;
```

Chosen over adding `padding-right` to the `tab` branch: padding is the smaller diff and would match `page` exactly, but it re-introduces the reflow the moment a form crosses into scrolling. Note this reserves the **scrollbar's** width, not 16px — if a visible gap is wanted as well it is the gutter *plus* padding, which is a decision for the preview.

**Label tone.** One line, `<Text>` → `<Text color="textSecondary">`. The value keeps `textPrimary`, so the contrast comes from dimming the key rather than emphasising the value.

## Review focus

- 🟡 `form/Form.component.tsx` › `FormGroup` label — `textSecondary` restyles every form label in every consuming application. One line, entirely a design call.
- 🟡 `form/Form.component.tsx` › `ScrollArea` — `scrollbar-gutter: stable` applies to both layouts, so every scrolling form gains a few px of right gutter.
- ⚪ `stories/form.stories.tsx` › `ScrollGutterAndLabelTone` — a deliberately short frame so both changes are visible in one view.

## How to test

1. `npm run storybook`
2. Templates → Form → **ScrollGutterAndLabelTone**.
3. Drag the frame's bottom edge until the form scrolls and back again. The fields should not shift horizontally as the scrollbar appears, and the bar should not touch them.
4. Check the label tone against the values, and specifically the disabled row — a disabled label is dimmed a second time by its `0.5` opacity, and may want to keep the primary tone rather than compound the two.
5. Confirm the custom scrollbar treatment (`ScrollbarWrapper` / `.scroll-fade`) still composes with the gutter.

## Follow-up

Needs designer sign-off on a live preview before it leaves draft — both halves change resting appearance fleet-wide.

## References

- Both items come from the same design review; the gutter is "adjust scrollbar layout" and the tone is the key/value standardisation item.

### DIFF
--- a/src/lib/components/form/Form.component.tsx
+++ b/src/lib/components/form/Form.component.tsx
@@ -139,6 +139,12 @@ const ScrollArea = styled(BasicPageLayout)`
   flex-grow: 1;
   align-self: stretch;
   overflow-y: auto;
+  /* Reserve the scrollbar's width whether or not it is showing. Without it a form
+     that grows past its box moves every field left by the scrollbar's width the
+     moment it becomes scrollable, and the bar sits directly against the fields
+     because the tab layout's 16px inset comes from an ancestor outside this
+     scroller. */
+  scrollbar-gutter: stable;
 `;
 
 // A FormSection lays its FormGroups out so labels align across the section from
@@ -334,7 +340,9 @@ const FormGroup = ({
       id={`${LABEL_PREFIX}${id}`}
       style={{ opacity: disabled ? 0.5 : 1 }}
     >
-      <Text>
+      {/* The key of a key/value pair: dimmer than the value beside it, so the two
+          are told apart by tone rather than by position alone. */}
+      <Text color="textSecondary">
         {label}
         {requireMode !== 'all' && required && ' *'}
         {requireMode === 'all' && !required && ' (optional)'}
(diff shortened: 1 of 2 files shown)


JeanMarcMilletScality:
Split into two PRs so the repair does not wait on the design decision:

- **#1196** — `scrollbar-gutter: stable` on the form's scroll area. Repair lane, dev review only.
- **#1197** — `FormGroup` label in `textSecondary`. Design lane, needs designer sign-off.

The inline review on the disabled-label contrast is carried into #1197, with measured ratios. Short version: the compounded dimming does drop it (4.29 → 2.97 on `darkRebrand`), so a disabled label now keeps the primary tone — but the disabled label already fails AA today at 4.29, so that guard avoids a regression rather than fixing contrast.

Closing in favour of the two above.

PR #1195, opened 2026-08-28
CopyButton: make iconOnly actually collapse an outline button

JeanMarcMilletScality:
**TL;DR** — `iconOnly` had no visible effect on an outline `CopyButton`: the label was hidden but the button kept its full width, leaving a lone icon in a wide empty box. It now collapses to the icon, and `iconOnly={240}` stays full width until its container drops below 240px.

## Context

`iconOnly` is the affordance a narrow container reaches for, and on an outline `CopyButton` it did nothing visible: the button stayed a wide outline box with one glyph in it. Found while making a password-field composite responsive — it was the one control in the row that could not narrow.

## Approach

The floor exists for a reason: an outline copy button's label swaps from `Copy X` to `Copied X!` on click, so without a reserved width the button resizes under the cursor. It was applied as an **inline style**, and that is the crux — an inline style outranks every stylesheet rule, so the `@container` query behind `iconOnly` could never reach it: `Button` dutifully hid the label while the box kept its full labelled width. Switching the floor off in JS would fix the boolean form and leave the numeric one half broken, so the floor moved into the stylesheet under the same query, reading its breakpoint off `iconOnly` itself rather than off a second copy of the number:

```ts
const FlooredButton = styled(Button)<{ $floor?: string }>`
  min-width: ${({ $floor }) => $floor ?? 'auto'};
  ${({ iconOnly }) =>                                                      // ←
    typeof iconOnly === 'number' &&
    `@container responsive (max-width: ${iconOnly}px) { min-width: auto; }`}
`;
```

| `iconOnly` | Floor | Result |
| --- | --- | --- |
| unset | always | never resizes on click |
| `true` | none emitted | no label to reserve room for |
| `240` | above 240px only | holds its width while labelled, collapses below |
| `0` | always | matches `Button`, which treats 0 as a breakpoint nothing is narrower than |

`0` is the row worth a second look: it is falsy, so a JS guard would read it as "always collapse" while `Button` reads it as "never collapse". Measured in a browser, since jsdom evaluates no `@container`: computed `min-width` is `175px` above the breakpoint and `0px` below it, and neither button carries an inline `min-width`.

## Screenshots

**CopyButton → Outlined Copy Button Icon Only** with the frame at its default 20rem — `iconOnly` is icon-width, `iconOnly={240}` is still labelled:

cui1195-icononly-frame-20rem

The same story with the frame dragged to 216px — the numeric form has collapsed too:

cui1195-icononly-frame-216px

## Usage

Sourced from the story in this diff. Before, `iconOnly` was accepted on an outline button but had no visible effect:

```tsx
<CopyButton variant="outline" label="Certificate" textToCopy="Certificate" />
```

After — either form collapses the button; the numeric form needs a `container` ancestor to resolve against, and keeps its no-resize floor above the breakpoint:

```tsx
<Box container>
  <CopyButton variant="outline" label="Certificate" textToCopy="Certificate" iconOnly={240} />
</Box>
```

## Review focus

- 🟡 `buttonv2/CopyButton.component.tsx` › `FlooredButton` — the floor moves from an inline style to a class, so a consumer's `style={{ minWidth }}` now wins over it where before it was silently ignored; intended, and the one thing here worth disagreeing with. The success-state `cursor`/`opacity` deliberately keeps its precedence over `style`, so it cannot be unstyled while clicks are still being swallowed.
- ⚪ `buttonv2/CopyButton.component.tsx` › `outlineWidthFloor` — the arithmetic is unchanged, only where it is applied; every outline `CopyButton` with no `iconOnly` renders the same floor as before.
- ⚪ `buttonv2/CopyButton.component.test.tsx` — the width-floor test is removed rather than adapted: jsdom cannot evaluate the query and `jest-styled-components` is not a dependency, so the assertion would have claimed more than it checked. The accessible-name test stays and the story is the gate.

## How to test

1. `npm run storybook`, then Components → CopyButton → **Outlined Copy Button Icon Only**.
2. The first button should be icon-width, not a wide outline box; hovering it shows `Copy Certificate`.
3. The second is full width at the frame's default size. Click it — the label becomes `Copied Certificate!` and the button must **not** change width. This is the one check no automated test covers: an isolated browser cannot be granted `clipboard-write`, so the success state never engages there.
4. Drag that frame below 240px: the button collapses to its icon, and returns to full width above 240px again. Click it while collapsed — the accessible name becomes `Copied Certificate!` and the icon turns into a check.
5. The other `Outlined Copy Button*` stories should be unchanged — same width, no resize on click.

### DIFF
--- a/src/lib/components/buttonv2/CopyButton.component.tsx
+++ b/src/lib/components/buttonv2/CopyButton.component.tsx
@@ -6,25 +6,19 @@ import { Button, type Props } from './Buttonv2.component';
 // The outline variant swaps its label on success ("Copy X" -> "Copied X!"), so the
 // button reserves room for the longer string and keeps its width when clicked.
 //
-// The reservation is a character-count heuristic, carried over unchanged from when
-// this was an inline style: `7rem` is the fixed part -- icon, padding, border and
-// the word "Copied!" -- and `length / 2` adds 0.5rem per label character as a
-// stand-in for average character width. That average is the weak point. Measured at
-// the 14px root the floor grows 7px per character against ~6px of real text, so it
-// over-reserves for a narrow label and is too small for an all-uppercase one, where
-// the content outgrows the floor and the button resizes on click anyway.
+// Character-count heuristic, inherited unchanged: `7rem` for the fixed part (icon,
+// padding, border, "Copied!") plus 0.5rem per label character as an average width.
+// The average under-reserves for an all-uppercase label, which still resizes.
 const outlineWidthFloor = (label?: string) =>
   `${(label ? label.length / 2 : 0) + 7}rem`;
 
 /**
  * A `Button` carrying the outline variant's width floor.
  *
- * The floor is a stylesheet rule rather than an inline `style` because `iconOnly`
- * hides the label through a `@container` query and an inline style outranks every
- * stylesheet rule — so an inline floor held a collapsed button at its full labelled
- * width with nothing in it. The breakpoint is read from `iconOnly` itself rather
- * than passed in beside it, so the floor and the label cannot answer to different
- * widths.
+ * A stylesheet rule, not an inline `style`: `iconOnly` hides the label via a
+ * `@container` query, which an inline style outranks -- so an inline floor held a
+ * collapsed button at its full labelled width. The breakpoint is read from
+ * `iconOnly` itself so the floor and the label cannot disagree.
  */
 const FlooredButton = styled(Button)<{ $floor?: string }>`
   min-width: ${({ $floor }) => $floor ?? 'auto'};
--- a/src/lib/components/buttonv2/CopyButton.component.tsx
+++ b/src/lib/components/buttonv2/CopyButton.component.tsx
@@ -3,9 +3,23 @@ import styled from 'styled-components';
 import { Icon } from '../icon/Icon.component';
 import { Button, type Props } from './Buttonv2.component';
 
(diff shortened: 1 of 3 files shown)


src/lib/components/buttonv2/CopyButton.component.tsx
```diff
@@ -1,7 +1,32 @@
...
 import { Button, type Props } from './Buttonv2.component';
 
+// The outline variant swaps its label on success ("Copy X" -> "Copied X!"), so the
+// button reserves room for the longer string and keeps its width when clicked.
+//
+// Character-count heuristic, inherited unchanged: `7rem` for the fixed part (icon,
+// padding, border, "Copied!") plus 0.5rem per label character as an average width.
+// The average under-reserves for an all-uppercase label, which still resizes.
+const outlineWidthFloor = (label?: string) =>
+  `${(label ? label.length / 2 : 0) + 7}rem`;
```
damiengillesscality, line 13: Does `.5rem` work better than `1ch` or is it for retro-compat ?

damiengillesscality (approved):

src/lib/components/buttonv2/CopyButton.component.tsx
JeanMarcMilletScality: On this, I did not touch the value of 7rem as it is working well. It was there before, with no explanation on why 7. Claude infered this explanation

PR #1218, opened 2026-09-10
improvement(charts): an optional logarithmic Y axis on Barchart and LineTimeSerieChart

clement-scality:
**Component**: `charts` — Barchart, LineTimeSerieChart

**TL;DR** — Adds an opt-in logarithmic Y axis to `Barchart` and `LineTimeSerieChart`.

Split out of #1190, which carried this plus a heatmap. The heatmap is in its own PR now; this one is the axis and nothing else.

### Context / Why

A metric that idles at 2 ms and spikes to 4 s cannot be read on a linear axis: the axis is sized by the spike, so the baseline where the metric spends nearly all its time is pressed flat against zero.

### 🧩 Approach

`yAxisScale="log"`, defaulting to `linear`. It is a display option and nothing more: no value is rescaled, and tooltips, legend and unit scaling all keep the numbers the caller passed in. Only the axis and where the marks land on it change.

Three things a log axis has to do differently, all in `charts/common` so both charts share them:

| | Why |
|---|---|
| The scale cannot start at zero | Bounded by the decades enclosing the data, read off the smallest positive value and the largest — so it follows the data rather than being anchored to 1. Values between 0.001 and 0.1 with spikes to 50 give a 0.001..100 axis. |
| Its ticks are the decades | Evenly spaced linear ticks on a log axis crowd into the top of the plot and read as a broken chart. Past six decades, whole decades are skipped — never subdivided. |
| A measured zero needs somewhere to go | It has no logarithm and no place on the scale, so the axis reserves one slot below its first decade, labels it `0`, and draws zeros there. |

That last one is the part worth arguing about. Without the reserved slot a zero has to be dropped, and a dropped zero leaves exactly the gap a missing sample leaves — making "the metric read zero" indistinguishable from "we have no data", which are not the same fact. Clamping to the first decade is worse, not better: it draws the zero as the smallest measured value, a number the reader will believe. The slot costs a decade's worth of height, so it is only reserved when the data actually holds a zero.

Two pairings are refused rather than drawn wrong. **Stacked bars fall back to linear**, since stacking places each segment at a cumulative sum and a segment's height would stop matching its value. **A symmetrical line chart refuses the scale at the type level** — half its axis is negative by construction.

### 📷 Screenshots

**Barchart — `LogarithmicScale`.** Linear on top: only the 40 000 bar is legible. Log below: every decade gets the same height, and `category5` (value `0`) is the stub at the reserved `0`.

Barchart logarithmic scale story, linear and log side by side

**LineTimeSerieChart — `LogarithmicScaleExample`.** Same series twice. On the linear axis the baseline is a flat line on zero; on the log axis its variation and both spike tiers are readable.

LineTimeSerieChart logarithmic scale story, linear and log side by side

### 🔍 Review focus

- 🟡 **Moderate** — `charts/common/chartUtils.ts › getLogAxis`, `placeNonPositiveValues` — the domain, the decade ticks and the reserved zero slot. The zero handling is the design decision in this PR; if you disagree with reserving a slot, this is the place to say so.
- 🟡 **Moderate** — `utils.ts › formatISONumber` — the raised decimal ceiling, and the only change outside `charts/`. `formatISONumber` is exported from the package root, so every caller sees it; it can only ever add decimals to a value below `1`, never remove any.
- 🟡 **Moderate** — `barchart/Barchart.utils.ts › getCurrentPoint` and `LineTimeSerieChartTooltip` — a zero is plotted at the reserved slot, so both tooltips map it back before formatting, including a caller's own Barchart `tooltip` renderer. No real value can collide with the slot: it sits below the smallest positive value in the data.

### 🧪 How to test

```bash
npm run storybook
```

*Charts → Barchartv2* and *Charts → LineTimeSerieChart*:

1. Open `Logarithmic scale` on either chart and compare the two stacked charts. Same data, only the axis differs.
2. Open `Logarithmic scale playground` and use the Controls panel — `yAxisScale` flips the axis live.
3. On the Barchart playground, tick **`stacked`**. The axis snaps back to linear: that is the documented refusal, not a bug.
4. On the LineTimeSerieChart playground, tick **`zeros`**. The zeros land on the reserved `0` at the bottom; untick and the axis loses that slot.
5. `Logarithmic scale below one` shows the axis following data between 0.001 and 0.1 with spikes to 50 — nothing is anchored to 1.
6. Still on that story, hover the baseline: the tooltip reads `0.002 /s`, not the `0.00 /s` it used to round to.

Theme switching is worth a pass: the reserved `0` and the tick colours come from tokens.

```bash
npm test -- src/lib/components/charts src/lib/utils.test.ts
```

### 🔗 References

**Breaking Changes**: none on the charts — `yAxisScale` is optional and defaults to `linear`, so every existing caller renders exactly as before. The one output change is `formatISONumber` on values below `0.01`, which is the bug fix itself: a caller that was getting `0.00` for `0.002` now gets `0.002`.

**Also fixes a tooltip rounding bug this PR surfaced.** On `LogarithmicScaleBelowOne` the tooltip printed `0.00 /s` for a measured `0.002`: `formatISONumber` capped the fraction at two decimals, while the axis derives its decimals from each tick's own magnitude. A rounded `0.00` is indistinguishable from the `0` the reserved slot legitimately reports — which is the exact confusion this PR exists to remove, so it is fixed here rather than deferred.

Below `1`, `formatISONumber` now raises its decimal ceiling to whatever keeps **two significant digits** — `0.002`, `0.0012`, `0.0025` — the same rule `formatLogTickValue` already applies to a tick, so a tooltip and the tick beside it agree. It only ever raises the ceiling, never lowers it, so every value from `0.01` up formats exactly as before. Two things it does *not* change: `0` still renders as `0`, and the `0.001` floor below which a value switches to scientific notation (`0.0009` → `9e-4`) is pre-existing and untouched — the notation switch is keyed on the value's magnitude, not on a digit count.

`charts/common/chartUtils.ts` carries the whole log-axis vocabulary — `getLogAxis`, `getMinPositiveValue`, `hasZeroValue`, `placeNonPositiveValues`, `formatLogTickValue`, `readLogPlottedValue` — next to `getRoundReferenceValue`/`getTicks`, which do the same job for the linear axis. Both charts wire it the same way, which is why the two diffs read almost identically.

`utils.ts` is the only file outside `charts/` — `formatISONumber`'s decimal ceiling, with `utils.test.ts` gaining a `formatISONumber` describe alongside the existing `getContrastText` one.

</details>

### DIFF
--- a/src/lib/components/charts/barchart/Barchart.tsx
+++ b/src/lib/components/charts/barchart/Barchart.tsx
@@ -13,7 +13,15 @@ import { Stack } from '../../../spacing';
 import { chartColors, ChartColors, fontSize } from '../../../style/theme';
 import { useChartLegend } from '../legend/ChartLegendWrapper';
 import { BarchartTooltip } from './BarchartTooltip';
-import { formatTickValue, getTicks, splitTickLines } from '../common/chartUtils';
+import {
+  formatLogTickValue,
+  formatTickValue,
+  getLogAxis,
+  getTicks,
+  hasZeroValue,
+  placeNonPositiveValues,
+  splitTickLines,
+} from '../common/chartUtils';
 import { useChartData } from './Barchart.utils';
 import {
   ChartHeader,
@@ -55,6 +63,9 @@ const BARCHART_PRESETS: Record<'default' | 'modern', ResolvedBarchartDisplayOpti
   default: { noBackground: false, showHorizontalGridLines: true, noYAxisLine: false, noTickLine: false, noHeader: false },
   modern:  { noBackground: true,  showHorizontalGridLines: true,  noYAxisLine: true,  noTickLine: true,  noHeader: true  },
 };
+
+/** A Recharts row. `null` is a value the axis cannot place — see `dropNonPositiveValues`. */
+type BarchartRow = { [key: string]: string | number | null };
 
 export type Point = {
   key: string | number;
@@ -116,6 +127,32 @@ export type BarchartProps<T extends BarchartBars> = {
    * Only the properties you specify are overridden; the rest come from the preset.
    */
   displayOptions?: BarchartDisplayOptions;
+  /**
+   * Y-axis scale.
+   *
+   * `'log'` is for a dataset whose values span orders of magnitude: on a linear
+   * axis the small bars collapse onto the baseline and only the largest one is
+   * readable, and a log axis gives each decade the same height instead. The
+   * axis is bounded by the decades enclosing the data, and its ticks are the
+   * decades themselves.
+   *
+   * It changes the display and nothing else: no value is rescaled, and the
+   * tooltip, the legend and the unit scaling all keep the numbers the caller
+   * passed in.
+   *
+   * Two things it cannot do:
+   * - A bar whose value is zero or negative is dropped, because a log axis
+   *   has nowhere to put it. It renders as a gap, which is the same thing an
+   *   absent bar renders as — so on a log axis "measured zero" and "no data"
+   *   become indistinguishable. They are not the same fact. Stay linear where
+   *   that difference matters.
+   * - `stacked` falls back to a linear axis. Stacking places each segment at
+   *   a cumulative sum, so on a log axis a segment's height stops matching
+   *   its value and the chart misreads.
+   *
+   * @default 'linear'
+   */
+  yAxisScale?: 'linear' | 'log';
 };
 
 /* ---------------------------------- MAIN COMPONENT ---------------------------------- */
@@ -141,6 +178,7 @@ export const Barchart = <T extends BarchartBars>(props: BarchartProps<T>) => {
     rightTitle,
     isLoading,
     isError,
+    yAxisScale = 'linear',
     displayPreset = 'default',
     displayOptions,
   } = props;
@@ -175,6 +213,8 @@ export const Barchart = <T extends BarchartBars>(props: BarchartProps<T>) => {
     rechartsData,
     topDomain,
     valueBase,
+    minPositiveValue,
+    normalizedMaxValue,
   } = useChartData(
     bars || [],
     type,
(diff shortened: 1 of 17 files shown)


src/lib/utils.ts
```diff
@@ -120,11 +121,24 @@ export const formatISONumber = (
...
+   * that reserves a slot for measured zeros, the two are indistinguishable.
+   * The ceiling is raised to whatever keeps two significant digits, which is
+   * the rule `formatLogTickValue` already applies to a log axis tick. It only
+   * ever raises the ceiling, so values from 0.01 up are formatted as before.
+   */
+  const maximumFractionDigits =
+    absValue < 1
+      ? Math.max(decimals, Math.ceil(-Math.log10(absValue)) + 1)
+      : decimals;
+
```
damiengillesscality, line 136: I think this will break stuff, I get the idea and the 0.00 issue but I can see some of our code doing something bad math and sudenly printing the result of `.1+.2-.3` which would show ~16 zero and destroy the view. I think in this case we should do a rounding not change the number of decimal which will break the display.
Also the rounding behaviour of Intl.NumberFormat is very inconsistant not sure how the "bug" behave in many browser but rounding beforehand can mitigate that.

src/lib/components/charts/linetimeseries/useChartData.ts
damiengillesscality, line 1: not sure of the value if we already give everything to get the minPositiveValue, perarps we should not compute it for every graph could be expensive in some cases

src/lib/components/charts/linetimeseries/LineTimeSerieChart.types.ts
```diff
@@ -26,6 +52,11 @@ export type NonSymmetricalChartSerie = {
  */
 export type SymmetricalChartSerie = {
   yAxisType: 'symmetrical';
+  /**
+   * Not available on a symmetrical chart: its axis spans negative values, which
```
damiengillesscality, line 56: This is a funny AI response sure the log of a neg value does not exist good point for your math exam claude but in this case we would just use symlog

src/lib/components/charts/linetimeseries/LineTimeSerieChart.types.ts
```diff
@@ -131,6 +162,11 @@ export type LineTimeSerieChartTooltipProps = {
     duration?: number,
   ) => React.ReactNode;
   isSymmetrical?: boolean;
+  /**
+   * A log axis's reserved zero band. A value drawn there is reported as the 0 it
+   * actually is, not as the position it occupies.
+   */
+  logZeroValue?: number | null;
```
damiengillesscality, line 169: I don't get what does the number you pass does exactly

src/lib/components/charts/linetimeseries/LineTimeSerieChart.tsx
```diff
@@ -185,8 +224,15 @@ export function LineTimeSerieChart({
               fill: theme.textSecondary,
               fontSize: fontSize.smaller,
             }}
-            tickFormatter={tickFormatter}
-            ticks={getTicks(topValue, yAxisType === 'symmetrical')}
+            tickFormatter={
+              logAxis
+                ? (value: number) =>
+                    formatLogTickValue(value, logAxis.zeroValue)
```
damiengillesscality, line 230: This `(value: number) => formatLogTickValue(value, logAxis.zeroValue)` should be memoized

src/lib/components/charts/barchart/Barchart.tsx
```diff
@@ -272,9 +346,18 @@ export const Barchart = <T extends BarchartBars>(props: BarchartProps<T>) => {
...
+            scale={logAxis ? 'log' : 'auto'}
+            domain={logAxis ? logAxis.domain : [0, topDomain]}
+            // A log axis is bounded by whole decades, so a value below the
+            // floor is clipped rather than dragging the axis down to it.
+            allowDataOverflow={logAxis !== null}
+            ticks={logAxis ? logAxis.ticks : getTicks(roundReferenceValue, false)}
+            tickFormatter={
+              logAxis
+                ? (value: number) =>
+                    formatLogTickValue(value, logAxis.zeroValue)
```
damiengillesscality, line 358: please memoize this

damiengillesscality (changes requested):
changes needed as per the messages above

src/lib/components/charts/linetimeseries/LineTimeSerieChart.types.ts
clement-scality: It's the Y coordinate zeros were drawn at, since a log axis has no position for 0 — I've rewritten the docstring to spell out where it comes from and when it's null.

damiengillesscality (approved):
LGTM

PR #1219, opened 2026-09-10
improvement(charts): a Heatmap composed from existing primitives

clement-scality:
**Component**: `charts` — a new `Heatmap`

**TL;DR** — Adds a `Heatmap` component: one row per entity, one column per time slot, each cell coloured by its value.

Split out of #1190, which carried this plus a logarithmic Y axis. The axis is in #1218 now; this one is the heatmap and nothing else.

### Context / Why

Nothing in `core-ui` reads one metric across many entities over time — the shape a service-status board or a per-node health strip wants. It started life in #1190 as a story-only recipe, a design proposal to be argued with before anything landed in `src/`. That review happened, the shape held, so it is a component.

It is assembled from `Box`, `Tooltip`, `Text` and `ChartLegend` rather than from a charting library: the grid is CSS, so it costs no Recharts instance and scales to a few hundred cells without one.

### 🧩 Approach

The component owns everything it draws — title, row labels, grid, x-axis, legend — so an app hands it data and a colour scale, not a composition:

```tsx
<Heatmap
  title="Monitoring Services Status"
  legendTitle="Service Status"
  scale={{ colorSet: { OK: theme.statusHealthy, /* … */ } }}
  rows={rows}
  columns={timeSlots}
/>
```

Two scales, and they type the data with them:

| `scale` | Cells | Legend |
|---|---|---|
| `discrete` (default) | value names | mounts its own `ChartLegendWrapper` from `colorSet`, so the legend both colours the cells and filters the grid on click. Omit `colorSet` to read a wrapper the caller put above instead — that is how several charts share one legend. |
| `continuous` | numbers | ramps one colour by opacity, beside a `HeatmapGradientScale` handed the same `max` the cells were ramped against, so the printed domain cannot drift from what the grid shows. |

The grid is driven by `columns`, not by each row's cells. A row shorter than the axis has to leave the rest of its line empty, and CSS auto-placement does the opposite — it pulls the next row's label out of the gutter and skews everything below it. A `null` cell holds its slot.

What stays with the caller is domain, not composition: which colour `OK` / `WARNING` / `CRITICAL` / `NONE` are worth, and in what order the legend lists them. The component has no business knowing those statuses.

### 📷 Screenshots

**`ScreenshotEquivalent`** — 5 services, 4 columns, the last one with no data yet. The rest of the stories are worth clicking through rather than screenshotting.

Heatmap, services against time slots with a status legend

### 🔍 Review focus

- 🟡 **Moderate** — `charts/heatmap/Heatmap.tsx` — the props surface, and two calls worth a second opinion. **One implementation per scale** rather than one branch inside it, because the discrete scale reads the legend context and a hook cannot be called conditionally; the dispatcher pays for it with a cast, since TypeScript narrows on a top-level discriminant and not on `scale.type` one level down. And **`showLegend` defaults to `true` even when the wrapper is inherited**, so two heatmaps under one shared wrapper draw two identical legends until the caller says otherwise — a predictable default preferred over an implicit rule.
- ⚪ **Minor** — `charts/heatmap/Heatmap.utils.ts` — the opacity ramp: clamped to its domain, and rounded to two decimals so a dense grid does not mint a styled class per cell.
- ⚪ **Minor** — accessibility: every cell is a tab stop, which is what makes its tooltip reachable from the keyboard (`Tooltip` opens on focus). On the `Dense grid` story that is ~400 tab stops. Say so if you would rather trade the keyboard path for a shorter tab order.

### 🧪 How to test

```bash
npm run storybook
```

*Charts → Heatmap*. It opens on a `Playground` whose **data is a control**, so it is meant to be poked at rather than read:

1. `Playground` → Controls → **`rows`**. It is a JSON editor: add a row, rename one, change any cell to `OK` / `WARNING` / `CRITICAL` / `NONE`. The x-axis is derived from the longest row, so adding cells adds columns.
2. Still on `rows`: **delete the last cells of one row**. Its line ends early and every other row stays put — that is the auto-placement fix.
3. Click a legend entry. Unselected statuses fade rather than vanish, so the shape of the data survives the filter.
4. `Dense grid` → push **`cellGap`** to 0 for a continuous timeline, then raise **`columns`** to 96 and use **`labelEvery`** to keep the axis legible. `noDataColumns` is the "collection has not caught up" tail.
5. `Numeric values` is the continuous scale: **`minOpacity`** is the floor that keeps the low end visible, and the gradient beside the grid states the domain the cells were ramped against.
6. Tab into the grid. A focused cell outlines itself and opens its tooltip — the same one hover gives.

Theme switching is worth a pass: the status colours, the cell outlines and the gradient all come from tokens.

```bash
npm test -- src/lib/components/charts/heatmap
```

### 🔗 References

**Breaking Changes**: none. `Heatmap`, `HeatmapGradientScale` and `getHeatmapMaxValue` are new exports on `next.ts`; nothing existing changes.

`charts/heatmap/` is the component: `Heatmap.tsx` (the frame, the grid and one implementation per scale), `HeatmapGradientScale.tsx` (the continuous legend, placed by the component or by the caller, like `ChartLegend`), `Heatmap.utils.ts` (the ramp and the domain) and their tests. `stories/Heatmap/` keeps the data and the theme colours, and nothing else.

Two shapes were tried and dropped along the way in #1190. One stacked a `GlobalHealthBar` per row — it worked, but it needs one chart instance per row and hides every x-axis but the last. The other was a `NestedProportionBar`, a different chart altogether, which is not part of this PR.

</details>


src/lib/components/charts/heatmap/Heatmap.tsx
```diff
@@ -0,0 +1,426 @@
...
+   * Hide the legend, for a heatmap whose scale is stated elsewhere: several
+   * grids under one shared legend, a ramp beside its own
+   * `HeatmapGradientScale`.
+   */
+  showLegend?: boolean;
+  /** Show one x-axis tick every N columns, to keep a dense axis legible. */
+  labelEvery?: number;
+  cellHeight?: string;
+  cellGap?: string;
+  /** The row label gutter. Labels truncate rather than widen it. */
```
JeanMarcMilletScality: No truncate mechanism found on the label. Label should use the ConstrainedText component

Image

Also labelWidth is string but a not valid CSS width value broke the layout completely. Improve typing or change how the grid column is compute to avoid this

src/lib/components/charts/heatmap/Heatmap.tsx
```diff
@@ -0,0 +1,426 @@
...
+      columns.length,
+      1,
+    )}, minmax(0, 1fr))`}
+    gap={cellGap}
+    alignItems="center"
+    flex="1"
+  >
+    {rows.map((row, rowIndex) => (
+      <React.Fragment key={`${row.label}-${rowIndex}`}>
+        <Box textAlign="right" pr={spacing.f8}>
```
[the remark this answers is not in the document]
JeanMarcMilletScality: Use constrainedText

Cuervino:
Hello Clément,

I pulled the branch and ran the storybook locally, so this is feedback from using it, not from reading the code.

First, the thing I want us to be explicit about: in a heatmap, colour is the only channel carrying information. By definition it is not accessible on its own, and here the tooltip is what saves it. I am fine with that for this component, it is the very nature of a heatmap, but two consequences follow. The limitation has to be written down in the guidelines rather than discovered, and I want to keep the number of components that rely on colour alone as small as possible. So this one should not become a habit.

What I like: the row labels are right-aligned, which makes them much easier to scan against the grid. The selected state is a good call too, it uses the design system's selection token and it is painted as an outline, so nothing moves when you hover. Accessibility is clearly not an afterthought either, every cell is reachable and announces itself, and the tooltip opens on focus, not only on hover.

Now the points I would like you to look at.

1. A new component needs its guideline page, and for this one it matters more than usual, because of the colour limitation above. Take the Button guideline as the model and keep it short to start with, the designers can grow it later. What I would want in a first pass: when to use a heatmap and when a table or a line chart is the better answer, the accessibility limitation stated plainly, which colour sets are allowed (status versus categorical), how the time axis and the tooltip express time, and the empty and no-data states.

2. The time axis does not follow our date format. It shows "08-25" and "09-02", which is month first and numeric, so it reads as 11 September for us and as 9 November for anyone used to day first. Our format guideline gives "DD MMM" (20 Jul) for a chart time axis. The tooltip itself is fine as it is.

3. The period in the title is optional at best. Our charts carry the unit, not the range, and even where you see a period in the BarChart stories, the time range is usually controlled globally, outside the chart. So "last 24 hours" and "last 14 days" written into a title will not hold once a real range selector drives the data.

4. Midnight. Showing it as 00:00 is fine by me, but no story crosses midnight today, so the choice is invisible. Please add one that does, so the decision to change nothing is on the record.

5. The hover should say how long a cell lasts. Today it gives a single instant, and I have to compare it with the next tick to work out whether I am looking at 15 minutes, an hour or a day. My preference is the full slot on the first line, "25 Aug 03:30 to 04:00". If that is too long, keep the start time and add the duration next to it, "25 Aug 03:30, 30 min".

6. The cursor turns into a hand on every cell, but nothing is clickable. It promises an action that does not exist, in every use case.

7. The stories shout. "OK", "WARNING", "CRITICAL", "NONE" in full uppercase reads as generated content, and stories are what people copy. Normal capital initials please.

8. Colours. When the values are not statuses, please do not reuse the status palette: in the backup story the cyan that means "Healthy" elsewhere becomes "Full", and that story mixes job types (Full, Incremental, Snapshot) with job outcomes (Skipped, Failed) in one scale. A neutral categorical set is what is missing, and we already have the series colours in the theme, which is what the same story uses at the top. One story goes further and hardcodes four colours that exist in no theme (#3B9EDB, #E8A33D, #8E6FD8, #5A6270), so they will not follow any theme at all.

9. Have you checked both light and dark? I ran it in A-Light and the component itself holds up. The hardcoded colours above are the part that will not.

10. What is the use case for the continuous scale? I am not convinced we need it. Distinct colours read better than a ramp, and if we do need intensity I would rather discretise it into a few steps. What settles it for me is the low end: a nearly transparent cell and a "no data" cell look almost the same, so "almost nothing" and "nothing" become hard to tell apart. What is your take on that one?

11. For percentages, the top of the scale should be pinned to 100 by default, overridable. Today it stops at the highest value in the data, so two charts side by side are not comparable. LineTimeSerieChart already does this with its percentage axis.

Cuervino (changes requested):
Requesting changes so this does not land before we settle the design points, which I left in full in the comment above: https://github.com/scality/core-ui/pull/1219#issuecomment-5637263389

Happy to discuss any of them, several are questions rather than asks.

Cuervino:
Even if my previous content looked AI-generated, it's really my personal review, but maybe the style is not understandable enough because generated by Claude. Please tell me if it's the case, I need to adapt and to have the correct tone, and it should be explainable enough.

src/lib/components/charts/heatmap/Heatmap.tsx
```diff
@@ -0,0 +1,478 @@
...
+  rows,
+  columns,
+  appearanceOf,
+  labelEvery = 1,
+  cellHeight = spacing.f20,
+  cellGap = spacing.f4,
+  labelWidth = '7rem',
+  formatColumnTick = (column) => (
+    <FormattedDateTime format="time" value={column} />
+  ),
```
Cuervino: The default tick is hardcoded to the time of day and ignores the step of the axis, so a daily axis prints "00:00" over and over unless the caller passes `formatColumnTick`. The component already knows the slot duration from `getColumnEnds`, so it could derive the default: time of day below a day, the abbreviated date from a day up.

It is invisible in the stories because the only daily one passes the override, so the first consumer is the one who will find it.

src/lib/next.ts
```diff
@@ -30,6 +30,9 @@ export {
   LineTimeSerieChart,
   GlobalHealthBar,
   Sparkline,
+  Heatmap,
+  HeatmapGradientScale,
+  getHeatmapMaxValue,
```
Cuervino: Let's drop the continuous scale: no use case today, and a published prop we have never exercised still has to be documented, tested and supported. If we need it later, we can easily introduce it.

That takes out `HeatmapGradientScale`, `getHeatmapMaxValue`, `ContinuousHeatmapProps` and `HeatmapContinuousScale`, plus the Numeric Values story. Two open points go with them: pinning the top of the ramp to 100, and the gradient issue the bot flagged when `min > 0`.

On steps instead of a ramp, no need to settle fixed versus configurable now. The day a product needs _intensity_, the use case will tell us.

stories/Heatmap/heatmap.stories.tsx
```diff
@@ -0,0 +1,631 @@
...
+    const columnCount = Math.max(
+      1,
+      ...args.rows.map((row) => row.cells.length),
+    );
+
+    return (
+      <Box maxWidth="60rem">
+        <Heatmap
+          title="Monitoring Services Status"
+          legendTitle="Service Status"
```
Cuervino: Sometimes you use title case and sometimes sentence case. I would prefer sentence case, so "Monitoring services status" here and "Service status" for the legend. The other charts in this file are already in sentence case, and examples are what people copy.

stories/Heatmap/heatmap.guideline.mdx
```diff
@@ -0,0 +1,119 @@
...
+
+**Collected nothing.** When the entity existed and the value is unknown — collection has not caught
+up, the exporter was down — that is a value like any other. Give it a name in the `colorSet` and a
+theme neutral, so it appears in the legend, can be filtered, and reads as "we do not know" rather
+than as a hole in the grid.
+
+<Canvas of={HeatmapStories.DenseGrid} sourceState="none" />
+
+An entirely empty grid is not the component's business. A heatmap with no rows renders its axis and
+nothing else; the surrounding page is what should say why.
```
Cuervino: I reworked this page. Two things were missing: what a heatmap hides through aggregation, and the name itself, since in UX research a heatmap is a different object (clicks or gaze painted over a screenshot), so it should read as "heatmap chart".

I also dropped the date format rule, because it belongs in the component rather than in every caller, and the sentence about an empty grid, because I could not verify it: with no rows the component still renders the title, the axis and the full legend.

Full version below, take what you want from it.

```suggestion
# Heatmap chart

A heatmap chart is a matrix. One row per entity, one column per slice of a
second dimension, and a colour at each crossing standing for the value there.

That second dimension is time in most cases, and status history is the case this
component was built for: one column per time slot, the colour saying how each
entity was doing then. Time is not a requirement. Any second dimension with few
enough values works.

<Canvas of={HeatmapStories.ScreenshotEquivalent} sourceState="none" />

A note on the name: in UX research, a heatmap is a different object, clicks,
scrolls or gaze painted over a screenshot of a page. Both are called heatmaps
and they have nothing to do with each other. Here it should be understood as
"heatmap chart".

## When to use it

Use it when the pattern matters more than the number, and when there are enough
rows and enough columns **for a pattern to exist**.

Use something else when:

- The exact values matter. A table states them, a heatmap only implies them
  through colour.
- There is one entity, or a handful. A line chart shows the trend and the
  magnitude, which a row of coloured squares throws away.
- There are two or three columns. That is a status list, not a grid.

## Every cell is an aggregate

A cell covers a slice and says one thing about it, so something was dropped on
the way. The rule that dropped it decides what the reader sees, and the same
data under two rules produces two different grids.

- **Worst value wins.** One minute of outage colours the whole hour. Nothing is
  missed, but a brief incident and a sustained one look alike.
- **Average.** Brief incidents disappear into the slots around them.
- **Last value.** The grid shows the state at the end of each slot and says
  nothing about what happened inside it.

Worst value wins is the usual choice for status history, because missing an
outage is worse than overstating one. It is still a choice, and one the reader
cannot recover from the picture. Say which rule applies, in the legend title or
beside the chart.

The aggregation happens before the component. `Heatmap` paints the values it is
given and never computes them, so this is a decision for the page, not a prop.

Slot width is part of the same decision. Narrower slots aggregate less and show
more, up to the point where the grid stops being readable. Anything that lets a
user change the window has to re-aggregate, and the answer on screen changes
with it.

## Accessibility: colour is the only channel

Colour carries the value on its own. No shape, no label and no position tells
one cell from another, so the grid stays closed to a reader who does not
separate the hues. The tooltip is what makes it usable: every cell takes focus,
carries an `aria-label`, and opens its tooltip on focus as well as on hover, so
the grid can be read from the keyboard and by a screen reader.

That trade is acceptable here, because a heatmap chart without colour is not a
heatmap chart. It is not acceptable as a habit:

- Pick colours that stay apart from one another, including for the most common
  colour vision deficiencies. Two neighbouring hues in the same set are the
  first thing to check.
- Keep the value set small. Four or five values is the ceiling at which distinct
  hues stay distinct.
- Do not treat this page as a precedent for the next component that leans on
  colour alone.

## Colour sets

Which palette applies follows from what the values mean, and the two cases do
not mix.

**Status values**, meaning states of health, take the theme status tokens
(`statusHealthy`, `statusWarning`, `statusCritical`) and a theme neutral for the
absence of data. They follow the theme, so a status grid holds in light and
dark with nothing else to do.

**Categorical values**, meaning anything else, take the series colours
(`lineColor1` to `lineColor8`, or `lineTimeSeriesColorRange`).

<Canvas of={HeatmapStories.CustomColorSet} sourceState="none" />

Two rules hold in both cases. Never borrow the status palette for a category: a
value painted `statusHealthy` claims to be healthy, and a reader who learnt that
colour elsewhere in the product will read it the same way here. Avoid hardcoded
hex values, they belong to no theme and follow none.

Order the legend the way the values mean something, through `scale.sortOrder`.
Alphabetical is the default and is rarely what the reader expects. Where the
stored value is not what should be shown, `scale.labelMap` renames it in the
legend and `formatValue` does the same for the tooltip and the `aria-label`.

## Time

The axis carries the start of each slot, thinned out by `labelEvery` so a dense
axis stays legible.

The axis does not mark the change of day. A slot that opens a new day carries
the same kind of label as the others, and the date is only in the tooltip.

<Canvas of={HeatmapStories.AcrossMidnight} sourceState="none" />

## Empty and no data

Two cases have to be distinguished: a slot where the entity had nothing to
report, and a slot where a value exists but was not collected.

**No cell.** A `null` cell, or a row shorter than the axis, leaves its slot
empty. It applies when the entity did not exist during that slot. The grid
follows `columns` rather than the length of each row, so a short row keeps its
place and the rows below it are not shifted.

**No value collected.** When the entity existed but its value is unknown, it is
treated as a value of its own. Declare it in the `colorSet` with a theme
neutral, so that it appears in the legend and can be filtered like the others.

<Canvas of={HeatmapStories.DenseGrid} sourceState="none" />
```

Cuervino (approved):
Hello Clément,

I ran the storybook on `7267bcf1`, most of what I raised is in. Thanks, and good call on the response codes.

I am approving so this is not stuck on me, on the understanding that the continuous scale goes before merge. See my comment on `next.ts`.

One question (it is what made me rework the guideline): every cell is an aggregate, and the rule behind it changes the picture: worst value wins, average and last value give three different grids from the same data. 
👉 Which rule should we recommend ? Do you think the component should help state it to the reader?

The rest is on the lines.

src/lib/components/charts/heatmap/Heatmap.tsx
```diff
@@ -0,0 +1,420 @@
...
+  renderTooltip,
+}: HeatmapGridProps<T>) => {
+  // read off the axis once, not once per cell
+  const columnEnds = getColumnEnds(columns);
+
+  return (
+    <Box
+      display="grid"
+      /* `repeat(0, …)` is invalid CSS, which would drop the whole template */
+      gridTemplateColumns={`${labelWidth} repeat(${Math.max(
```
JeanMarcMilletScality: The labelWidth type is string but an invalid CSS declarations will broke the whole grid.
The type should be tighten or move this value out of the grid templates

src/lib/components/charts/heatmap/Heatmap.tsx
```diff
@@ -0,0 +1,420 @@
...
+  legendTitle?: ReactNode;
+  /**
+   * Hide the legend, for a heatmap whose scale is stated elsewhere: several
+   * grids standing under one shared legend.
+   */
+  showLegend?: boolean;
+  /** Show one x-axis tick every N columns, to keep a dense axis legible. */
+  labelEvery?: number;
+  cellHeight?: string;
+  cellGap?: string;
```
JeanMarcMilletScality: `cellGap` ans `cellHeight` need a docs on what is expected and the default value. Also the type could be tighten.

src/lib/components/charts/heatmap/Heatmap.tsx
```diff
@@ -0,0 +1,420 @@
...
+export type HeatmapDiscreteScale = {
+  type?: 'discrete';
+  /**
+   * The color of each value. Omit it to read a `ChartLegendWrapper` the caller
+   * put above instead — which is how several charts come to share one legend.
+   */
+  colorSet?: ChartLegendWrapperProps['colorSet'];
+  sortOrder?: ChartLegendWrapperProps['sortOrder'];
+  /** Display labels for the legend items, when a value is not its own label. */
+  labelMap?: ChartLegendWrapperProps['labelMap'];
```
JeanMarcMilletScality, line 67: `sortOrder` and `labelMap` are ignored if there is no `colorSet`. Need at least a comment setting this clearly

src/lib/components/charts/heatmap/Heatmap.tsx
```diff
@@ -0,0 +1,420 @@
...
+                    : defaultTooltip(cell, formatValue)
+                }
+              >
+                <Cell
+                  $color={color}
+                  $opacity={opacity}
+                  $height={cellHeight}
+                  tabIndex={0}
+                  role="img"
+                  aria-label={`${row.label} ${formatValue(value)}`}
```
JeanMarcMilletScality: Missing the time value for the label.
Also each cell is a tab stop, the container should have be `role="grid"` and the cell role="gridcell"

src/lib/components/charts/heatmap/Heatmap.tsx
```diff
@@ -0,0 +1,420 @@
...
+  const columnEnds = getColumnEnds(columns);
+
+  return (
+    <Box
+      display="grid"
+      /* `repeat(0, …)` is invalid CSS, which would drop the whole template */
+      gridTemplateColumns={`${labelWidth} repeat(${Math.max(
+        columns.length,
+        1,
+      )}, minmax(0, 1fr))`}
```
JeanMarcMilletScality: No min on cell, meaning that if the column number increase or if available space is small, the cell will not stop shrinking. Also the labels for time can overlap.
@Cuervino Should the cell have a min size ? And what happens if there is no enough space ?

clement-scality: A new Horizontal Scroll story is available to try this, giving each cell a minimum width so a long axis scrolls sideways instead of shrinking.

Cuervino: Yes, it's a very good question, @JeanMarcMilletScality. 
No minimum size that makes the grid scroll, as a status history is read as one picture, so the window has to fit.

As is, this component might be tricky, especially if it doesn't calculate anything regarding how much space it has, but only depends on the backend. 
And maybe you (Jean-Marc) need to provide a bit of responsiveness rules here since it's a topic that you start to be an expert at. 
It might be possible to set the legend on the bottom and not on the right if there is not enough space, and also maybe there is a way to combine and to make a new aggregation if there is not enough space to reduce the number of horizontal slots.

src/lib/components/charts/heatmap/Heatmap.tsx
```diff
@@ -0,0 +1,567 @@
...
+                <Tooltip
+                  key={key}
+                  placement="top"
+                  overlay={
+                    renderTooltip
+                      ? renderTooltip(cell)
+                      : defaultTooltip(cell, formatValue)
+                  }
+                >
+                  <Cell
```
JeanMarcMilletScality: A wrapper holding the role `gridcell`should be put on top of the tooltip
  ```suggestion
                  <Box
                    role="gridcell"
                    key={key}
                    /* grid, not block: TooltipContainer is inline-block and would
                       otherwise shrink to fit a cell that has no width of its own */
                    display="grid"
                  >
                    <Tooltip
                      placement="top"
                      overlay={
                        renderTooltip
                          ? renderTooltip(cell)
                          : defaultTooltip(cell, formatValue)
                      }
                    >
                      <Cell
                        $color={color}
                        $opacity={opacity}
                        $height={cellHeight}
                        tabIndex={0}
                        /* without the slot, a screen reader counts columns to place it */
                        aria-label={`${row.label}, ${formatSlot(
                          column,
                          columnEnds[columnIndex],
                        )}, ${formatValue(value)}`}
                      />
                    </Tooltip>
                  </Box>
  ```

stories/Heatmap/heatmap.guideline.mdx
```diff
@@ -0,0 +1,153 @@
...
+(`lineColor1` to `lineColor8`, or `lineTimeSeriesColorRange`).
+
+<Canvas of={HeatmapStories.CustomColorSet} sourceState="none" />
+
+Two rules hold in both cases. Never borrow the status palette for a category: a
+value painted `statusHealthy` claims to be healthy, and a reader who learnt that
+colour elsewhere in the product will read it the same way here. Avoid hardcoded
+hex values, they belong to no theme and follow none.
+
+Order the legend the way the values mean something, through `scale.sortOrder`.
```
JeanMarcMilletScality, line 124: Maybe we should mention here the `sortOrder: 'status' `

stories/Heatmap/heatmap.guideline.mdx
```diff
@@ -122,9 +122,23 @@ colour elsewhere in the product will read it the same way here. Avoid hardcoded
...
 Order the legend the way the values mean something, through `scale.sortOrder`.
-Alphabetical is the default and is rarely what the reader expects. Where the
-stored value is not what should be shown, `scale.labelMap` renames it in the
-legend and `formatValue` does the same for the tooltip and the `aria-label`.
+It takes three forms:
+
+- **`'alphabetical'`**, the default. It is rarely the order a reader is looking
+  for: severity, or the order a pipeline runs in, almost never sorts by name.
+- **`'status'`**, a shorthand for `Success`, `Warning`, `Failed`, in that order.
+  It keeps only those three, and a value named anything else still paints its
```
JeanMarcMilletScality: Warning is correct but I'm not sure this is the correct place to put it or at least not in this form: Example:   - **`'status'`**, a shorthand for `Success`, `Warning`, `Failed`, in that order.
    Any other value is left out of the legend, including the absence value.
 Maybe add a more complete warning in JSDoc of sortOrder ?

JeanMarcMilletScality (approved):

PR #1223, opened 2026-09-18
improvement(layout): a guideline for the page patterns (CUI-52)

Cuervino:
Templates/Layout had nine stories and no guideline, so the page composition rules existed nowhere. The `AppContainer` slots were demonstrated in code, but nothing stated what goes in each slot, for which type of page, or when the overall summary applies. Two stories even labelled it "optional", which left the decision to whoever built the screen.

## What the rules are based on

An inventory of the twelve ARTESCA listing pages: Truststore, Locations, Data Services, Identity Providers, Accounts, Virtual machines, Users, Groups, Nodes, Volumes and All Buckets. Every rule below classifies all twelve without exception.

## What is added

`stories/layout.guideline.mdx`:

- the three slots, and what the context bar is for
- the two resource listing page forms, full page and list-detail, each with what it is, when to use it, and its composition rules
- browsable collections, which govern the breadcrumb, the presence of the overall summary, the level stated in the header, and the control that moves up one level
- dashboards, which have no canonical composition, with the condition under which one should be promoted to a pattern
- forms, marked as not yet defined

`ResourceListPage` in `stories/layout.stories.tsx`, embedded in the guideline. It composes the overall summary and a real `Table` with its search and create button. The existing stories demonstrate the plumbing with placeholder panels, which is not something anyone can start from. This one is meant to be copied.

## Known follow-up

The guideline documents two page patterns. The list-detail form has no story of its own yet.

Generated with with a deep interaction with Claude Code.


stories/layout.guideline.mdx
```diff
@@ -0,0 +1,141 @@
...
+
+**Composition rules**
+
+- Selection drives the right panel. The selected row stays highlighted in the list.
+
+- Actions on the selected item sit at the top of the detail panel, not in the list.
+
+- The overall summary is present only when the collection is browsable, see below. When it is not, the detail panel carries its own header naming the selected item.
+
+Where the split between the two panels sits is an implementation detail, not a rule. Three ratios are available.
```
JBWatenbergScality: How to guide the choice between them ? If we stay vague I think we will always ask about this

stories/layout.guideline.mdx
```diff
@@ -0,0 +1,141 @@
...
+
+<Canvas of={LayoutStories.ResourceListPage} />
+
+#### List-detail page
+
+**What it is**<br />
+A list on the left, the detail of the selected item on the right.
+
+**When to use it**<br />
+Items are compared and edited without leaving the list. If the details are complex and do not fit a key-value presentation, use an edit control (via a "Edit" button, for instance) on the value instead of a right-hand detail panel, and present the resource on a full page (and its edition will be a form in full page).
```
JeanMarcMilletScality: This one is not clear and ends up with 'use Full Page'

stories/layout.guideline.mdx
Cuervino: Agreed, staying vague just moves the question to every new page. Here is what replaces that sentence:

The split follows the information density on each side, measured on two things:

1/ how many columns the list needs to identify and compare an item,
2/ and whether the detail part is a single key-value list, or also carries tabs, nested tables, charts, configuration, forms.

- **30 / 70** when the list only identifies items (a name, possibly with a status) and the detail is where the work happens: tabs, several sections, nested tables, configuration.

- **65 / 35** when the list is the subject, scanned, sorted and compared over three or more columns, and the detail can be compacted in a key-value list. Key-value content is kept narrow on purpose (even if the values can be edited via forms or something else).

- **50 / 50** when both sides carry a moderate amount. Most often needed when you have two tables, one for the list, one for an attribute of an item in the list.

When both sides are dense, list-detail is not the correct pattern. Use a full page for the collection and a dedicated page for the item.

stories/layout.guideline.mdx
```diff
@@ -0,0 +1,141 @@
...
+    </div>
+    <div style={{ marginTop: '0.5rem' }}>30 / 70</div>
+  </div>
+  <div>
+    <div style={{ width: '328px', height: '185px', overflow: 'hidden' }}>
+      <div style={{ width: '1600px', height: '900px', transform: 'scale(0.205)', transformOrigin: 'top left' }}>
+        <Story of={LayoutStories.Layout2TwoPanelsSeventyThirty} />
+      </div>
+    </div>
+    <div style={{ marginTop: '0.5rem' }}>70 / 30</div>
```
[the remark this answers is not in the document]
Cuervino: Fixed, the label now reads 65 / 35.

stories/layout.guideline.mdx
Cuervino: You are right, the paragraph answered its own question and then argued the opposite. Split in b242e91b:

**When to use it**
Items are compared and edited without leaving the list.

**When not to use it**
When both sides are dense: the list needs several columns to be scanned, and the detail does not fit a key-value list. Use a full page for the collection and a dedicated page for the item. Editing then happens on the value itself (via an "Edit" button, for instance) or as a full page form.

The same idea was also stated further down, under the ratios. It is now only here, so the pattern is settled before the split is chosen. Leaving this open for you to confirm the wording works.

Cuervino: I updated the text.

JeanMarcMilletScality (approved):

